PR #15, opened 2013-09-28
Add some missing extensions

mp3butcher:
Add some missing extensions ..Don't know if's the right place


robertosfield:
I am just going through pending pull requests, this one is waaay out of date.  I've done a quick scan of the changes and it looks like the OSG master has all the ones I looked in.  If it's missing something please update to master and apply your changes and send them to me.

PR #16, opened 2013-11-06
Some fixes

maksqwe:


openscenegraph:
I'm closing this pull request for the following reason.

The change to FBX plugin is no longer relevant as the code is completely different now.

The change to MeshOptimizers.cpp is likely inappopriate.  I recall a discussion about this change or perhaps a simular with the code author on osg-users/osg-submissions and the author of the code felt the orignal code was actually correct.

If it is still felt there is an issue wit the cost in OSG master then I'd encouraging the submitter to open a discussion or the osg-users mailing list/forum.

PR #24, opened 2014-10-02
removed redundent extra header include.

AdriCS:
Removed an extra header include for <QGLWidget>.


openscenegraph:
Merged, commit bf343240e4263207150a2bf8bab206dbb3783cc2

PR #33, opened 2015-05-22
Observer pattern for state changes (request for code review)

szellmann:
Dear Robert,

in a discussion a while ago (http://forum.openscenegraph.org/viewtopic.php?t=14721) I asked if general changes to the scene graph can be monitored in a rather generic fashion.

I now started to enhance the osg::Observer class so that it may not only be used to monitor deletion, but also state changes.

The feature is far for complete, I implemented it for Matrix / PositionAttituteTransform and for modifications to Group's children list. At this early stage, I'd now like to ask you if you would have a look to decide if the approach is suitable at all. Being able to monitor the scene graph is a rather important feature to us. But then, it is only useful to us if it eventually goes into the official repos. If you don't find it suitable, what approach would you suggest?

I added an "osgobserver" example which demonstrates how one should use the feature (the example needs C++11 for I feel that we're not going to keep it, anyway. Please tell me if that's a problem).

If you find this useful, I'd be glad to put some more effort into it to make the feature more complete.

For the time being I designed this for backwards compatibility, but I believe that interface changes (e.g. implementing osg::Observer as a template) could simplify the way the feature is used. There may however be caveats to this.

I decided to issue a "pull request" for simplicity's sake, but of course I do not expect that you merge at this early stage.

Best regards,
Stefan


szellmann:
Oh, just seen that the old submission rules still apply. Sorry for that.

openscenegraph:
This request was reviewed but not found to be appropriate for merging with the core OSG as the changes are too intrusive, entail a performance overhead for all users for niche functionality.

PR #38, opened 2015-10-11
use uintptr_t type to save pointer value

crvv:


openscenegraph:
I'm closing this pull request as the request is not an appropriate change.

The relevant code in master uses in64_t in place of (long int) this correctly adapts a 64 bit type, which this pull request will change type depending upon the build target which is not appropriate.

crvv:
1. `long int` is `int32_t` on Windows.
2. The variable `image` is a pointer, and C++ has the specialized type `uintptr_t` for pointers.

openscenegraph:
uintrptr_t isn't supported on older compilers, it's C++11 and later so it just isn't an option.  

The OSG/GL define GLintptr and GLsizeiptr perhaps one of these could be used as a stand in,

PR #57, opened 2016-02-04
Add missing implementations for asDrawableCullCallback and asDrawableEventCallback

scrawl:
Hi Robert,

I just updated to the latest commit and noticed that drawable cull callbacks were no longer working. As it turns out there was a missing implementation of the asDrawableCullCallback() and asDrawableEventCallback(), after implementing these methods everything is working as expected.

Thanks,
Jannik


scrawl:
Merged in ca337bb7344216f4cbbcdceb86063a99c8b49395

PR #60, opened 2016-02-24
Fixed shadowing warnings.

tingdahl:
Hi,

I (finally) got so annoyed with all shadowing warnings on mac, so I chose to kill em all. There are no functional changes in these changes. Variables are renamed, if unused, they are removed. Other variables have received smaller scope.

Everything compiles and links well, without shadow warnings. If this change is accepted, I suggest that -Wshadow is added as default, or perhaps int the Aggressive build flags.


openscenegraph:
Don't know why github has closed this and other pull requests, there seems to have been some error somewhere along the line.  Files changed also looks broken so it's not possible to find out what this pull request changed or it's source.  Don't know if this is temporary glitch with github or not, or repository specific one.

susnux:
@openscenegraph You should be able to see the changes here: https://github.com/openscenegraph/OpenSceneGraph/commit/f63f9fd44cb899f9a9f142b9d6d8205f0f5225f1

Did you some push -f (history rewrite)? Because that can lead to problems like this.

openscenegraph:
The problems with the pull requests getting invalidated was doing to me fixing problems in the authorship of commits, in fixing these the history was rewritten,

For this particular pull request I ended up adding -shadow to the gcc compile options and then going through the OSG master fixing all the warning that came up, so this submission should now be moot, simply moving to git master head should fix all the warnings.

PR #66, opened 2016-03-22
convert AUTHORS.txt to UTF-8

fcami:
convert AUTHORS.txt to UTF-8
Some of the process was manual but I believe I got things right.


openscenegraph:
AUTHORS.txt is auto-generated so changes it will be lost when it's updated.  The correct place to look at is osgversion/Contributors.cpp.

PR #68, opened 2016-04-21
Add a cut visualize for osgvolume

vivijind:
add CutFuncProperty and AnyCutFuncProperty, so we can use it to visualize the cutted volume in x or y or z direction. And we can also visualize it from a cutting plane.


openscenegraph:
For cutting planes on the XYZ axis one simply changes the osgVolume::Locator dimensions there is no need to add additional functionality/overhead to do this.

If you still feel your submission is worthwhile could you please rebase you fork of the OSG and then generated a new submission.  I will assume this topic is closed till a new submission is generated.

PR #71, opened 2016-05-03
add safety checking when dereferencing ref_ptr

rdiankov:
currently my applications segfaults whenever i try to do something with uninitialized ref_ptrs, this is pretty bad since i can't even cleanup my applications and resources when this happens.

so i made a small patch to ref_ptr that throws an exception whenever the pointer is dereferenced. this is similar to the boost and c++11 shared pointers and i think a lot of users will appreciate this.
it is surrounded in #ifdef blocks, so old behavior can be restored using a cmake option.


openscenegraph:
For some reason github has closed this and other pull requests without any intervention on my part.  No clue yet to why this has happened.  The web github interface isn't allowing me to reopen the request or track down the changed files.  Could you please submit a request.

PR #74, opened 2016-05-21
Added missing stdexcept header, fixes #61

### LINKED ISSUE #61: stdexcept header missing
Hello

During installation, I got an error saying std::runtime is not a member of std (at /src/osgPlugins/daedaewriter.cpp line267). Simply including stdexcept solved it. Please include it in the source as well.

Thanks,
Rohan

susnux:
As the title says this will fix the build on some more pedantic systems.
See also #61 which is fixed by this one.


openscenegraph:
Which systems did you encounter problems with and what were the errors you saw?

openscenegraph:
Merged.

PR #77, opened 2016-06-08
Mods to make GL3 profile work on MacOSX

mathieu:
Following some (old) thread on GL3 and Mac, I managed to get it working with the osgsimplegl3 sample (I added an Optimizer to tessellate since Polygons and Quads don't get handled anymore by OpenGL )

The path followed is to attach VBO to VAO on Mac GL3 profile.

I added _useVertexBufferObjects=true when fixed pipeline isn't available to geometries.


mathieu:
not mature please ignore

PR #94, opened 2016-07-01
Updates gles plugin

marchelbling:
This PR updates the GLES plugin with latest Sketchfab changes:

* improves geometry splitting
* processes/cleans animation (for subsequent osgjs serialization)
* adds (yet another) normal smoother


openscenegraph:
I have merged the changes but am concerned that various fixes I've made over the past couple of weeks to resolve Covertiy Scan reported defects will be lost. I'll be running another scan later today.  It would be useful to get any new/old defects resolved and pushed back to Sketchfabs version of the gles plugin.

marchelbling:
@openscenegraph thanks for merging so quickly, Robert.
I'm in the process of syncing sketchfab repo with the main trunk so those PRs *should* have your last changes included.
I hope to then sync our repo with osg main repo once every week or 2 weeks as it is too painful otherwise and will look at coverity to improve the coverity score.

PR #97, opened 2016-07-02
Updates osgjs

marchelbling:
* cleans serialization
* adds animation support
* updates text serialization
* updates LOD serialization
* updates binary buffers control

### DIFF
--- a/src/osgPlugins/osgjs/Adaptor
+++ b/src/osgPlugins/osgjs/Adaptor
@@ -1,0 +1,117 @@
+/*  -*-c++-*- 
+ *  Copyright (C) 2010 Cedric Pinson <cedric.pinson@plopbyte.net>
+ *
+ * This library is open source and may be redistributed and/or modified under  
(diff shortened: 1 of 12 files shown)


openscenegraph:
Marc, the Vec3us animation codes don't compile as there no slerp method in Vec4us.  It doesn't make sense to do general vector math on unsigned shorts so I'm surprised these have been added to osgAnimation and used in the osgjs plugin.  I can't merge something that is broken so I'm closing this pull request.

marchelbling:
Robert, I removed the slerp from my PR as I thought it compiled anyway (I'm surprised the code compiled on my machine).
The reason we have this `Vecxus::slerp` method is because we perform animation channels compression and those methods are needed as the osgAnimation is a bit rigid.
So basically in our code base we have

```cpp
void slerp ( value_type  t, const Vec3us& from, const Vec3us& to) {
}
```

I did not see a clean solution to avoid this ugliness at the time I needed it so I'll need to see how to fit this more cleanly.

PR #102, opened 2016-07-05
Update gles+osgjs

marchelbling:
This PR

* cleans some gles coverity defects (remaining should only be false positive that need to be sorted out cleanly)
* updates osgjs plugin to support serialization; the history of changes is squashed; details can still be found on cedricpinson fork if needed

As compressed animation channels are no longer part of the main repo, I added a compilation flag for the osgjs plugin. The commit is isolated and the flag is not activated by default.
I am yet to find a better solution to make this plugin entirely free from our specific code.

Note: this PR will not change the gles compilation issues. We only compile on OSX/ubuntu and did not encounter any issue with the plugin.


openscenegraph:
Hi Marc,

I've gone ahead and merged, things compile OK for me but got a few warnings about an unused parameter and a showed variable.  I've checked in fixes to this:

https://github.com/openscenegraph/OpenSceneGraph/commit/ca4355974f68501f5b2c074d35c361ce04450768

Could you make sure this gets merged back into your fork of the OSG. Thanks.

PR #104, opened 2016-07-05
Fixes LWS scene load

marchelbling:
`readRefFile` calls `readRefObjectFile` which finally calls `readObject`. However, the LWS plugin currently implements only `readNode` and not `readObject`.


openscenegraph:
Just reviewed the changes and it'll introduce a dangling pointer as the temporary smart pointer return form readRefNodeFile() will be deleted within the same line of code.

The correct fix will be to implement the readObjectFile().   I've gone ahead an implement this, see commit:

https://github.com/openscenegraph/OpenSceneGraph/commit/8066a1da044a187d526d7285b07bba1684914922

PR #105, opened 2016-07-12
Moves gles visitors implementation in cpp files

marchelbling:
@robertosfield as discussed last week, this moves *most* gles visitor implementations into clean cpp files. This should hopefully make it easier to compile using MSVC.
I hope this do not introduce any new coevrity defect but let me know if I missed some


openscenegraph:
Thanks Marc.  I'm still deep into the Vertex Attribute Object work so will wrap this up before I tackle this submission.  The VAO should be wrapped up in the next day or so.

openscenegraph:
I have had two confirmations from VS2015 users that this resolves the build failure with large object sizes, thanks again for the refactor :-)

PR #106, opened 2016-07-19
Added OSG_CURL_SSL_VERIFYPEER option to the curl plugin

rafagaitan:
Added OSG_CURL_SSL_VERIFYPEER option to the curl plugin to allow connecting to secure servers without the certificate

### DIFF
--- a/src/osgPlugins/curl/ReaderWriterCURL.cpp
+++ b/src/osgPlugins/curl/ReaderWriterCURL.cpp
@@ -142,6 +142,7 @@ EasyCurl::EasyCurl()
     _previousHttpAuthentication = 0;
     _connectTimeout = 0; // no timeout by default.
     _timeout = 0;
+    _sslVerifyPeer = 1L;
 
     _curl = curl_easy_init();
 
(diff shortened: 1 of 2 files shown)


openscenegraph:
I have just had a look at the changes and searched for which versions of libcurl support the CURLOPT_SSL_VERIFYPEER option.  I have libcurl-7.47.0 installed on my Kubuntu 16.04 system and the curl headers headers don't contain this option so I can't see how this patch will compile on my pretty standard and modern system.  

If we are to merge changes relating to CURLOPT_SSL_VERIFYPEER then at minimum there needs to be a version check to make sure that build isn't broken.

rafagaitan:
I will try the check the version and resubmit the pull request asap.

rafagaitan:
I've checked the 7.47 tag in the github repository version and it actually contains the CURLOPT_SSL_VERIFYPEER (https://github.com/curl/curl/blob/curl-7_47_0/include/curl/curl.h#L1064), but it seems they use a macro to create the  option so probably a grep with the whole option name didn't find anything. 
Did you find any problems building it in your system?

openscenegraph:
Hi Rafa, the macro is indeed why I missed the CURLOPT_SSL_VERIFYPEER in the header.  I have just done a search from the libcurl github history for curl.h and got back to 2001 and the entry is still in there so it looks safe to merge the change without a check for version :-)

PR #109, opened 2016-08-01
Fix RenderBuffer not being attached

scrawl:
Commit 55b3becb7c0ceecf26b0dae2c6db80f9f2a8a6d4 broke the attachment of render buffers. This PR fixes the issue.


openscenegraph:
I just started looking into this bug today and was starting to wonder what recent changes had caused the regression.  It's great that you spotted to cause of the bug and come up with a sensible resolution. Thanks,

PR #120, opened 2016-08-29
added missing _EXPORT tags to all META_(Object, NodeVisitor, StateAttribute)

LaurensVoerman:


openscenegraph:
Why did you judge these classes to need an export? These classes have all been compiling fine under Windows as far as I'm aware.

LaurensVoerman:
Hi Robert,

You are correct, these classes do not NEED an export tag, the current
version is correct and is fully usable.
However I did think it might be a good idea to add the tags anyway - and
have now changed my mind about that.

The reason it currently works is that the code is completely in the header
file, so the compiler can generate inline versions for the final
executable. Adding the export tag will force the compiler to add the code
to the library and place it in the export table. This will give the
compiler the option to not inline the code while generating the final
executable, as the exported version from the library is available.
I expected this to result in a smaller total size, but that's not the case.
I just compiled both version (Release build - VS 2015 update 3)
library's (13259.5 kB) add 153.5 kB with the update tags,
applications (479.5 kB) add 0.5 kB with the update tags,
plugins (45826 kB) add 71 kB with the update tags.
I am especially disappointed by the size increase in the plugins, I
expected those to stay the same or decrease in size.




I have two reasons to think it would be useful to add those export tag:
> I think it's not very clear why it works without the export tag, and that
it will stop working if an implementation is moved into a .cpp file. This
will only show up as a link error (missing symbol) while compiling the
application actually using this function on windows, as the osg project
will continue to build just fine.
> ABI checking: these classes are part of the ABi, and changes will cause
crashes if the client app does not match the osg dll's. By exporting them
it would be possible to detect the change in the exported symbols.

Regards, Laurens.

robertosfield:
I the cases where you've added the *_EXPORT the class is a base class that has to be sub classed to be used.  These subclasses will need to have their own *_EXPORT usage.  These a small base classes that principally just define interfaces and aren't the type of class that will usually need an implementation in the .cpp.

As yet I haven't seen a compelling reason to merge the change as it doesn't offer any additionally functionality, portability and even adds to the binaries size.  For all these reason I'm going to close this pull request.  If later we find that we do need to add *_EXPORT because the lack of it is causing problem then we can merge these on a case base case basis.

LaurensVoerman:
Hi Robert,
I agree. I'll close this request so I can safely prune some branches in my git repro.
Regards, Laurens.

PR #125, opened 2016-09-10
Lazy Disabling disabled non-aliased vertex attrib array

fwiesel:
When OSG_GL_VERTEX_ARRAY_FUNCS_AVAILABLE is not true, all vertex operations are reflected to the aliased vertex attrib pointers.

So any addtional disabling of the aliased functions is not only not necessary, it also leeds to vertex attributes being disabled, unless they are used via their aliased functions

### DIFF
--- a/src/osg/State.cpp
+++ b/src/osg/State.cpp
@@ -1265,6 +1265,7 @@ void State::disableVertexAttribPointersAboveAndIncluding( unsigned int index )
 void State::lazyDisablingOfVertexAttributes()
 {
     // OSG_NOTICE<<"lazyDisablingOfVertexAttributes()"<<std::endl;
+    #ifdef OSG_GL_VERTEX_ARRAY_FUNCS_AVAILABLE
     if (!_useVertexAttributeAliasing)
     {
         _vertexArray._lazy_disable = true;
@@ -1279,7 +1280,7 @@ void State::lazyDisablingOfVertexAttributes()
             itr->_lazy_disable = true;
         }
     }
-
+    #endif
     for(EnabledVertexAttribArrayList::iterator itr = _vertexAttribArrayList.begin();
         itr != _vertexAttribArrayList.end();
         ++itr)
@@ -1291,6 +1292,7 @@ void State::lazyDisablingOfVertexAttributes()
 void State::applyDisablingOfVertexAttributes()
 {
     //OSG_NOTICE<<"start of applyDisablingOfVertexAttributes()"<<std::endl;
+    #ifdef OSG_GL_VERTEX_ARRAY_FUNCS_AVAILABLE
     if (!_useVertexAttributeAliasing)
     {
         if (_vertexArray._lazy_disable) disableVertexPointer();
@@ -1303,6 +1305,7 @@ void State::applyDisablingOfVertexAttributes()
             if (_texCoordArrayList[i]._lazy_disable) disableTexCoordPointer(i);
         }
     }
+    #endif
     for(unsigned int i=0; i<_vertexAttribArrayList.size(); ++i)
     {
         if (_vertexAttribArrayList[i]._lazy_disable) disableVertexAttribPointer(i);


fwiesel:
Hi,

the calling sequence of the (presumed) bug is the following:

In `Geometry::drawVertexArraysImplementation`:

[lazyDisablingOfVertexAttributes](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Geometry.cpp#L785) sets all non-ffp attributes to `_lazy_enable=true` [here](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/State.cpp#L1270-L1280)

Later the Geometry sets my various vertex attributes with `setVertexAttribPointer`, and at the end calls [applyDisablingOfVertexAttributes](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Geometry.cpp#L834), which disables [all aliased attributes](applyDisablingOfVertexAttributes), which means also my vertex attribute, if it happens to coincide with an aliased index.


Could it be, that the same problem occurs, when I disable the `VertexAttributeAliasing` on a fixed function pipeline? The logic seem there to be the same, except that it depends there on the _current_ state of `_useVertexAttributeAliasing`.

src/osg/State.cpp
```diff
@@ -1265,6 +1265,7 @@ void State::disableVertexAttribPointersAboveAndIncluding( unsigned int index )
 void State::lazyDisablingOfVertexAttributes()
 {
     // OSG_NOTICE<<"lazyDisablingOfVertexAttributes()"<<std::endl;
+    #ifdef OSG_GL_VERTEX_ARRAY_FUNCS_AVAILABLE
```
filnet: should be left aligned

robertosfield:
Hi,

The suggested changes don't look like they'll probably resolve the issue.  The best thing to do would be to raise this issue on osg-users mailing list/forum providing a small example that illustrates the problem so others can reproduce the problem and then between us find a viable solution.

fwiesel:
Okay, I will do that instead.

fwiesel:
Sorry, for re-opening it. It looks, like it would take more more than a couple of minutes to extract that problem.

I can work on creating an isolated case, but I hope, I can avoid that work and convince you with simply showing you the api trace.

I can run it without the patch and with, and in without you will see the API calls disabling all the vertex attributes, which are in the alias list. And that without any prior api call enabling them.

Let me turn the tables of the issue:
What is the purpose of tracking the state of the vertex attributes (_lazy_disable/_enabled) twice when aliased? E.g once in _vertexArray and the second time in _vertexAttribArrayList[0]

I assume, the initial idea of the code was to disable the fixed function attributes, when using aliasing, but I cannot see a point in disabling vertex attributes, which have not even been enabled.

But I fail to see, how the current code achieves that.

robertosfield:
I believe you have highlighted a potential bug with the duality of old vertex array and vertex attrib path with vertex alias, this will need to be resolve but I pull request is not the best place for a general discussion on this topic - as I've said, the right place for discussion on this is the osg-users mailing list/forum.  I'm not going get drawn into a discussion that on a pull request when it's the wrong place for such discussions.

openscenegraph:
I'm closing this pull requests as it introduces it;s own bugs without properly addressing the issue raised. This issue needs to be resolved, the best way to resolve it will be to create a test case that illustrates the bug. share this with the OSG community via the osg-users mailing list/forum and then have the community tackle the problem.

openscenegraph:
I have just had a chance to look at the issue again, in particular looking at the how things are implemented in the vertex_array_object branch so make sure that it's sounds, and I believe it should fix the issues you have seen in previous incarnation found in the master branch of include/osg/State.

I will soon be merging vertex_array_object with master so it should hopefully resolve the issues you have. I would recommend testing out the vertex_array_object branch.

fwiesel:
Thanks, the small test program does indeed work without any issues with the vertex_array_object branch.

I'll have a go with it on my actual app.

PR #129, opened 2016-09-13
VS2015 fix a few cases of warning C4589: Constructor of abstract clas…

LaurensVoerman:
Constructor of abstract class 'X' ignores initializer for virtual base class
apparently I missed these warnings in previous attempts to get a clean compile.

### DIFF
--- a/src/osgAnimation/AnimationManagerBase.cpp
+++ b/src/osgAnimation/AnimationManagerBase.cpp
@@ -63,8 +63,6 @@ void AnimationManagerBase::operator()(osg::Node* node, osg::NodeVisitor* nv)
 
 
 AnimationManagerBase::AnimationManagerBase(const AnimationManagerBase& b, const osg::CopyOp& copyop) :
-    osg::Object(b, copyop),
-    osg::Callback(b, copyop),
     osg::NodeCallback(b,copyop) // TODO check this
 {
     const AnimationList& animationList = b.getAnimationList();
--- a/src/osgGA/CameraManipulator.cpp
+++ b/src/osgGA/CameraManipulator.cpp
@@ -21,8 +21,6 @@ CameraManipulator::CameraManipulator()
 
 
 CameraManipulator::CameraManipulator(const CameraManipulator& mm, const CopyOp& copyOp):
-     osg::Object(mm,copyOp),
-     osg::Callback(mm,copyOp),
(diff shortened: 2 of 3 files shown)


openscenegraph:
Have you tried these changes out under Linux?  I fear this might be a case of fixing warnings under VS but then creating warnings elsewhere as I've seen this happen previously.

LaurensVoerman:
Hi Robert,
I have not tested these on linux yet, the travis CI
<https://travis-ci.org/openscenegraph/OpenSceneGraph/jobs/159613374> build
did not create a waring (don't know which compiler version that is).
I just ran the modifications trough
gcc version 4.7.2 (Debian 4.7.2-5)
gcc version 6.1.0 (GCC)
and got no warnings.
Regards, Laurens.

openscenegraph:
Hi Laurens,

I've just created a branch on my system and pulled you're changes to them, follows are the warnings that I get:

/home/robert/OpenSceneGraph/src/osgGA/CameraManipulator.cpp:23:1: warning: base class ‘class osg::Object’ should be explicitly initialized in the copy constructor [-Wextra]
 CameraManipulator::CameraManipulator(const CameraManipulator& mm, const CopyOp& copyOp):
 ^
/home/robert/OpenSceneGraph/src/osgGA/CameraManipulator.cpp:23:1: warning: base class ‘class osg::Callback’ should be explicitly initialized in the copy constructor [-Wextra]
and:

/home/robert/OpenSceneGraph/src/osgGA/StandardManipulator.cpp: In copy constructor ‘osgGA::StandardManipulator::StandardManipulator(const osgGA::StandardManipulator&, const osg::CopyOp&)’:
/home/robert/OpenSceneGraph/src/osgGA/StandardManipulator.cpp:50:1: warning: base class ‘class osg::Object’ should be explicitly initialized in the copy constructor [-Wextra]
 StandardManipulator::StandardManipulator( const StandardManipulator& uim, const CopyOp& copyOp ):
 ^
My compiler is:

> c++ --version
c++ (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609

This leaves us with the quandary that fixing the warning under VS will generate a warning elsewhere...  Not a good tradeoff.  Fixing the VS warning may require us to disable it.  I'm open to suggestions.

PR #133, opened 2016-09-16
Visual Studio 2015: disable C4589: Constructor of abstract class X ignores initializer for virtual base class Y

LaurensVoerman:
Hi Robert,
I suggest we disable the warning for visual studio 2015. It's valid C++ and I would prefer to keep the full initializer list (as g++ -Wextra suggests) as template for the derived classes.

Regards, Laurens.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1056,7 +1056,13 @@ ELSEIF(MSVC)
         #C4100 'identifier' : unreferenced formal parameter
         #C4127 Error Message conditional expression is constant
         #C4706 assignment within conditional expression
+    IF (MSVC_VERSION GREATER 1800)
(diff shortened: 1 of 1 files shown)


robertosfield:
Hi Laurens,  Is there any reasons why we would just disable this warning for all VS versions?

LaurensVoerman:
Hi Robert,
I think we can disable the warning for all VS versions, I was just a bit
careful because I can't find the warning in the documentation (here)
<https://msdn.microsoft.com/en-us/library/mt694070.aspx>
Regards, Laurens.

LaurensVoerman:
Hi Robert,
Submitted the proposed simple version, closing this PR.
Regards, Laurens.

PR #134, opened 2016-09-16
Visual Studio 2015: disable C4589: Constructor of abstract class X ignores initializer for virtual base class Y

LaurensVoerman:
Hi Robert,

I failed to see the g++ -Wextra warnings because my cmake (version 2.8.9) is to old for the current script. Here is an enhanced version with support for a few more cmake version, and added warning flags for g++ 4.7 and 4.8 (-Wmaybe-uninitialized -Wextra).
I left -Wshadow out because my 4.7 compiler (4.7.2 ) generates a huge number of shadow warnings.
For g++ 6 and above I added -Wno-misleading-indentation ; I checked all warning g++ 6.1 generated and currently they don't indicate actual bugs.

Regards, Laurens.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1044,30 +1044,17 @@ IF(CMAKE_COMPILER_IS_GNUCXX)
     # to be the workaround.
     SET(OSG_AGGRESSIVE_WARNING_FLAGS -Wall -Wparentheses -Wno-long-long -Wno-import -pedantic -Wreturn-type -Wmissing-braces -Wunknown-pragmas -Wunused)
 
-    IF(${CMAKE_MAJOR_VERSION} GREATER 2 OR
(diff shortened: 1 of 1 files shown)


robertosfield:
Hi Laurens,

The number of changes feels far more extensive than the change in functionality warrants.  If users want to add extra warnings then they can via ccmake  and just editing the approprirate variable.

FYI, my plan is to make CMake 2.8 the minimum for git master and hence OSG-3.6.

openscenegraph:
Could we just simplify the code so that it only has the checks against CMAKE_CXX_COMPILER_VERSION without any of the fallbback for older versions of CMake?  If we have older versions of CMake and we use this variable I'd expect the comparisons to fail and the additions to the warning flags to be bypassed.  

This should lead to  slightly conservative behaviour w.r.t. warning flags on systems that have out of date CMake but up to date gcc versions, this has to be a pretty small set of users with completely benign consequences for them.

LaurensVoerman:
Hi Robert,
modified as requested.
Regards, Laurens.

PR #135, opened 2016-09-18
Honor the override flags of texture attributes in SharedStateManager

scrawl:
The SharedStateManager had a bug where the OverrideValue flags of texture attributes are ignored, and set to ::ON no matter what the previous flags where.

### DIFF
--- a/src/osgDB/SharedStateManager.cpp
+++ b/src/osgDB/SharedStateManager.cpp
@@ -175,7 +175,11 @@ void SharedStateManager::shareTextures(osg::StateSet* ss)
     const osg::StateSet::TextureAttributeList& texAttributes = ss->getTextureAttributeList();
     for(unsigned int unit=0;unit<texAttributes.size();++unit)
     {
-        osg::StateAttribute *texture = ss->getTextureAttribute(unit, osg::StateAttribute::TEXTURE);
+        const osg::StateSet::RefAttributePair* pair = ss->getTextureAttributePair(unit, osg::StateAttribute::TEXTURE);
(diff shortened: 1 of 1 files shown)


scrawl:
Updated my fix, the previous version had a missing NULL pointer check.

openscenegraph:
I have just reviewed the changes and they fix a bug that I wasn't even aware for which is cool.

While reviewing it occurred to me that once you have the pointer the AttributePair you should be able to simply do:

   pair->second = textureFromSharedList;

This would be faster and require less code, the maps being modified wouldn't need any reallocations as well.

scrawl:
That doesn't work because the getTextureAttributePair returns a const pointer. Should a non-const version be added?

openscenegraph:
Good point. It's not generally safe to pass back a non const pointer as a user could modify the pair.first and break the map.  A const cast and modifying the pair.second will be safe, not elegant, but safe.

scrawl:
In my opinion a minor optimization is not worth the const_cast. If you want to go that route feel free to apply it on top of my changes but as far as I'm concerned this PR is finished.

openscenegraph:
I have a look at the implementing using the pair approach and my earlier assertion that the const Pair get was required because of the look up is not correct, the pair isn't used as a key at all, so it's perfectly safe to return a non const Pair.  I'm currently adding this to osg::StateSet.

openscenegraph:
I have just checked in a fix using the RefAttributePair approach I suggested above commit is : 7e0c30b8843d440ee24a231dd376fa749f5fd698

PR #137, opened 2016-09-22
fix error in CmakeLists: CMAKE_CXX_COMPILER_VERSIO

LaurensVoerman:
Hi Robert,
I failed to restore a test for cmake, and this was merged in my previous commit. Someone made a comment about this - no idea who that was. Anyway - here's the correct version.
Regards, Laurens.


filnet:
I made that comment. Nice that you saw it :)

PR #140, opened 2016-09-24
Bind only if VertxArrayObject has been generated

fwiesel:
Hi,

the current code of the `vertex_array_object` branch calls `glBindVertexArray​(0)`, which in the core profile is an immutable state, so it still works in the compatibility mode by accident.
It happens in the line, which is already marked with a TODO.

My suggestion to fix it by some defensive programming: The idea being, that the id `0` is the default state and is never generated by `glGenVertexArrays​`.
You can reach the default state only by `VertexArrayState::unbindVertexArrayObject()` and not also by `VertexArrayState va; va.bindVertexArrayObject()`, which seems to me a questionable way to do it.

The current state breaks code as the given [Example](https://github.com/fwiesel/vertexarrayfunctest/blob/master/main.cpp), where the `VAO` is created in a camera `DrawCallback` outside the draw function, which otherwise still on the stable OSG version.

Cheers,
  Fabian

### DIFF
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -527,7 +527,6 @@ inline void Drawable::draw(RenderInfo& renderInfo) const
         return;
     }
 
-    // TODO, add check against whether VAO is active and supported
     if (state.getCurrentVertexArrayState()) state.getCurrentVertexArrayState()->bindVertexArrayObject();
 
 
--- a/include/osg/VertexArrayState
+++ b/include/osg/VertexArrayState
@@ -164,7 +164,7 @@ public:
 
         void deleteVertexArrayObject();
 
-        inline void bindVertexArrayObject() const { _ext->glBindVertexArray (_vertexArrayObject); }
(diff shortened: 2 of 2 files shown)


openscenegraph:
I have reviewed your changes and feel that the additional check is inappropriate.  Calling glBindVertexArray​(0) unbinds the vertex array object which is perfectly legal.  

If there are two adjacent VertexArrayState dispatched but the first has a VAO associated with and the second has no VAO associated with it, binding the second VertexArrayState will correctly unbind the VAO, if you add the check you suggest then we'd need to have another check elsewhere to call unbdinVertexArrayObject(), so we'd have additional checks for now actual gain in functionality, so we'd loose performance for no gain.

fwiesel:
Are you interested in a different proposal, which preserves backwards compatibility and avoids an unnecessary `glBindVertexArray​(0)`?

Or you'd you rather see all vertex array objects being managed via `VertexArrayState`?

PR #141, opened 2016-09-24
VAO id is a GLuint

fwiesel:
There is an implicit cast to GLint in the getter which I guess is not intentional.


openscenegraph:
Well spotted, now merged.

PR #146, opened 2016-09-30
cmake fixes for Visual Studio failing to find optimized.lib

LaurensVoerman:
Hi Robert,
I had a few problems with Visual Studio trying to link to a "optimized.lib" and a "debug.lib". 
I traced those back to a change in cmake stuffing into the XYZ_LIBRARY the full optimized;XYZ.lib;debug;XYZd.lib string. I solved this by using the XYZ_LIBRARY_RELEASE if present.
I found some of the changes in cmake to be from mid 2015, but I can't find a specific version. I think my version is safe for all cmake versions.
Regards, Laurens.


robertosfield:
What is the situation if Collada is built with debug Zlib and the OSG links it to a release version?  Look at your changes it looks to make the assumption that Callada is built in release mode and needs release libs - I think this is a reasonable assumption but it's worth clarifying.

LaurensVoerman:
Hi Robert,
I added the debug zlib, an found out that the if statement in the previous version did not work at all. The git log got a bit messed up by merging in the current master, but I think it will still merge properly. If thats not the case I can make a new branch and a new pull request.
Regards, Laurens.

LaurensVoerman:
I managed to clean up the log with a rebase and (forced) push.
Regards, Laurens.

robertosfield:
On 6 October 2016 at 10:24, LaurensVoerman <notifications@github.com> wrote:

> I managed to clean up the log with a rebase and (forced) push.
> Regards, Laurens.
>

I will have a bash with the PR today see how I get on.

Cheers,
Robert

PR #149, opened 2016-10-07
fix visual studio 2015 warning C4250

LaurensVoerman:
Hi Robert,
recent changes to Callback and Eventhandler cause a warning in visual studio
warning C4250: 'osgViewer::InteractiveImageHandler': inherits 'osgGA::EventHandler::osgGA::EventHandler::asEventHandler' via dominance
warning C4250: 'osgVolume::PropertyAdjustmentCallback': inherits 'osgGA::EventHandler::osgGA::EventHandler::asEventHandler' via dominance

the log message in git for the change:
Date: 10/6/2016 7:27:23 PM
Message: Fixed crash and double call bug in handle event callbacks attached to Drawable

Regards, Laurens.

### DIFF
--- a/include/osgViewer/ViewerEventHandlers
+++ b/include/osgViewer/ViewerEventHandlers
@@ -512,6 +512,8 @@ public:
 
     META_Object(osgViewer, InteractiveImageHandler);
 
+    virtual const EventHandler* asEventHandler() const { return osgGA::EventHandler::asEventHandler(); }
+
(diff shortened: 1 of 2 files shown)


openscenegraph:
Hi Laurens, just so happens that I'm do a Windows build right now and have fixed these.  My changes have the const and non const versions of the asEventHandler() methods, but otherwise is the same solution as you went for.  These fixes are now checked into git master.

LaurensVoerman:
Hi Robert,
thanks, I just noticed my missing non const version. Back to synchronizing
again for me.
Regards, Laurens.

PR #154, opened 2016-10-11
hyperthreading detection for windows, distribute threads over actual cores.

LaurensVoerman:
Hi Robert,
the windows version of SetProcessorAffinityOfCurrentThread now works on the main thread too, and results in a performance drop for my windows machine (with hyperthreading enabled), because the main thread is bound to core0/thread0 an graphics to core0/thread1. As main and graphics actually overlap, this has a performance penalty. I implemented a simple workaround for windows, assigning the actual core - and leaving the thread choice to the os.
I cannot find a decent solution for linux. I managed to get the cpuid opcode to return some result describing the capabilities of the cpu, but not to tell me if hyperthreading actually is enabled. So for now I have left this with a fallback implementation.
Regards, Laurens.

### DIFF
--- a/include/OpenThreads/Thread
+++ b/include/OpenThreads/Thread
@@ -33,7 +33,14 @@ namespace OpenThreads {
  *  Note, systems where no support exists for querying the number of processors, 1 is returned.
  *
  */
-extern OPENTHREAD_EXPORT_DIRECTIVE int GetNumberOfProcessors();
+extern OPENTHREAD_EXPORT_DIRECTIVE unsigned int GetNumberOfProcessors();
+/**
+ *  Get the number of pysical processor cores and the number of threads per processor
+ *
+ *  Note, systems where no support exists for querying the number threads, GetNumberOfProcessors() and 1 is returned.
+ *
+ */
+extern OPENTHREAD_EXPORT_DIRECTIVE unsigned int GetNumberOfHtProcessors(unsigned int *HtCount);
 
 /**
  *  Set the processor affinity of current thread.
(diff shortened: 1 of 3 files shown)


openscenegraph:
Hi Laurens,

I have just done a first pass review of the changes and I agree with the need to detect the actual CPU core count and hyperthreading and utilize this when configuring the affinity to use when setting up osgViewer.  I don't feel the approach you have taken is ideal though.  

I'd rather than have a means of detecting the hyper-threading and letting the application/osgViewer decide upon what to do about it,  For OpenThreads::Affinity and setting affinity I'd suggest we stick to it's CPU relating to logical cores and not have any attempt internally in OpenThreads to set case down to physical processor.

The osgViewer::ViewerBase::configureAffinity() method needs to be modified so it can query for the presence of hyper-threading and do it's own mapping to physical cores where appropriate. So main threads goes to 0, any draw cull threads then are assigned to 2 onwards in intervals of 2 if there is hyper-threading.  This increment would need to be in sync with the logical vs physical core relationship.  

The API for finding out whether there is hyper-threading of some sort will need some though. I'd be inclined to have a method that returns a container of logical to physical core mapping or a std::pair<int,int> for logical and physical cores.

This topic would be best discussed on osg-users mailing list/forum as others may well have their own insights/knowledge to contribute.  In particular detecting hyper-threading under Linux, OSX etc. is something we'd need to add as well as Windows.

openscenegraph:
Hi Laurens,  I have just done a search of the web on the topic of detecting hyper-threading and it looks like we may be able to use __cpuid assembly to query it directly from the CPU.

PR #156, opened 2016-10-14
unlock cpu's 32-63 on 64 bit windows

LaurensVoerman:
Hi Robert,
the current windows implementation uses a 32 bit affinity and a 32 bit shift - constraining the mask to only 32 cpu's. Documentation says 64 bit windows can use 64 cpus in a processor group - sadly I don't have a machine to test this on.
Regards, Laurens.

### DIFF
--- a/src/OpenThreads/win32/Win32Thread.cpp
+++ b/src/OpenThreads/win32/Win32Thread.cpp
@@ -570,9 +570,10 @@ size_t Thread::getStackSize() {
 static int SetThreadAffinity(HANDLE tid, const Affinity& affinity)
 {
 	unsigned int numprocessors = OpenThreads::GetNumberOfProcessors();
-	std::cout << "setProcessorAffinity() : affinity.activeCPUs.size()=" << affinity.activeCPUs.size() << ", numprocessors=" << numprocessors << std::endl;
+	std::cout << "SetThreadAffinity() : affinity.activeCPUs.size()=" << affinity.activeCPUs.size() << ", numprocessors=" << numprocessors << std::endl;
 
-	DWORD affinityMask = 0x0;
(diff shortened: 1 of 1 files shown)


src/OpenThreads/win32/Win32Thread.cpp
```diff
@@ -570,9 +570,10 @@ size_t Thread::getStackSize() {
 static int SetThreadAffinity(HANDLE tid, const Affinity& affinity)
 {
 	unsigned int numprocessors = OpenThreads::GetNumberOfProcessors();
-	std::cout << "setProcessorAffinity() : affinity.activeCPUs.size()=" << affinity.activeCPUs.size() << ", numprocessors=" << numprocessors << std::endl;
+	std::cout << "SetThreadAffinity() : affinity.activeCPUs.size()=" << affinity.activeCPUs.size() << ", numprocessors=" << numprocessors << std::endl;
 
-	DWORD affinityMask = 0x0;
+    DWORD_PTR affinityMask = 0x0;
```
robertosfield: Indentation looks like it's been changed, tab vs spaces issue?

robertosfield:
Hi Laurens,

What happens on a Win32 build?  Will the DWORD usage still fine?

In my quick review the indentation looked to be wrong on a couple of the new lines.

Robert.

LaurensVoerman:
Hi Robert,
the original has a lot of tabs, in the line you point out I changed a tab
to 4 spaces.

In the win32 build a DWORD_PTR is 32 bit, matching the second argument
of SetThreadAffinityMask(HANDLE, DWORD_PTR).
32 bit windows only supports 32 logical cores, so that should work out just
fine.
Regards, Laurens.

PR #157, opened 2016-10-14
modified thread affinity & reconfig on setThreadingModel

LaurensVoerman:
Hi Robert,
In your previous commit you moved the threading config to a different function, causing the treads not to be configured after switching to a different threading model. This PR fixes that with a call to setUpThreading in setThreadingModel. 

other changes:
- set the first cull thread to the same cpu as the main thread - those should not overlap
- the database pagers got a very weird mask - assign a set of actual free cpu's

Regards, Laurens.

### DIFF
--- a/src/osgViewer/ViewerBase.cpp
+++ b/src/osgViewer/ViewerBase.cpp
@@ -139,6 +139,7 @@ void ViewerBase::configureAffinity()
 
     if (requiresCameraThreads)
     {
+        availableProcessor--;//cull can share a core with main
         Cameras cameras;
         getCameras(cameras);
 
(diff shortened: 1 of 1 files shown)


robertosfield:
I think these issues are best split into separate PR's.  The setUpThreading change looks reasonable as the does the change for sharing the first camera thread with the first CPU.  Possibly the same could be done for the first graphics thread in the culldraw threading case, though this would require some more code logic.  The third change w.r.t mask for the database pager, I didn't use the same evens, odds layout of cores used for the other threads as the consequence of hyper threading affecting database paging isn't significant - these are threads that have high IO latency so having a core a bit more contended isn't a big deal.  So I allowed the database pager threads to occupy all the remaining threads, without risking sitting on cores that the main or cull/draw sit upon.

LaurensVoerman:
Hi Robert,
I made the split pull requests, this one can be closed.

for the databasePagers: After assigning main - cull -cull - gpu threads to
vcores 0,2,4 and 6
the databasepagers got an affiniymask for vcores 1,2,3,4.
The PR changes that to the free vcores: 1,3,5,7 - as I think this was the
intention. I am not at all sure
this is the best solution, a lower priority might cause less frame drops,
but I do not have a test ready to test this.
Regards, Laurens.

PR #158, opened 2016-10-17
first cull thread can share a core with main thread

LaurensVoerman:


LaurensVoerman:
-- just to save a previous remark.
>Robert Osfield Oct 14 (3 days ago)
> The setUpThreading change looks reasonable as the does the change for sharing the first camera thread with the first CPU. Possibly the same could be done for the first graphics thread in the culldraw threading case, though this would require some more code logic.

openscenegraph:
Hi Laurens, I'm back looking at pull requests.  I'm looking at the configureAffinty() I think it'd be OK to share the same process for culldraw and cull threads so logic for this would need to be a bit different then in your submission.  I've made the following commit to do this:

https://github.com/openscenegraph/OpenSceneGraph/commit/137b596943a9a1ef853da7a3ba444adee48e2d68

The above commit makes this PR moot so will close it.

PR #160, opened 2016-10-17
assign free vcores to DatabasePager

LaurensVoerman:
-- repeating the relevant parts of the discussion on the previous PR:
Robert Osfield notifications@github.com Oct 14 (3 days ago)
to openscenegraph., me, Author 

> I think these issues are best split into separate PR's. 
> The third change w.r.t mask for the database pager, I didn't use the same evens, odds layout of cores used for the other threads as the consequence of hyper threading affecting database paging isn't significant - these are threads that have high IO latency so having a core a bit more contended isn't a big deal. So I allowed the database pager threads to occupy all the remaining threads, without risking sitting on cores that the main or cull/draw sit upon.

for the databasePagers: After assigning main - cull -cull - gpu threads to vcores 0,2,4 and 6 
the databasepagers got an affiniymask for vcores 1,2,3,4. 
The PR changes that to the free vcores: 1,3,5,7 - as I think this was the intention. I am not at all sure
this is the best solution, a lower priority might cause less frame drops, but I do not have a test ready to test this.
Regards, Laurens.

### DIFF
--- a/src/osgViewer/ViewerBase.cpp
+++ b/src/osgViewer/ViewerBase.cpp
@@ -184,14 +184,15 @@ void ViewerBase::configureAffinity()
 
         OSG_NOTICE<<"  databasePagers = "<<databasePagers.size()<<std::endl;
 
-        availableProcessor = availableProcessors[availableProcessor % availableProcessors.size()];
-
         OpenThreads::Affinity databasePagerAffinity;
+        //if any free CPU's remain, assign the databasePagers to those
+        for (; availableProcessor < availableProcessors.size(); ++availableProcessor) databasePagerAffinity.add(availableProcessors[availableProcessor]);
+
         for(DatabasePagers::iterator itr = databasePagers.begin();
             itr != databasePagers.end();
             ++itr)
         {
-            (*itr)->setProcessorAffinity(OpenThreads::Affinity(availableProcessor, numProcessors-availableProcessor));
+            (*itr)->setProcessorAffinity(databasePagerAffinity);
         }
     }
 }


robertosfield:
I need to make the logic a bit more sophisticated for the allocation of database threads, but this PR as it stands isn't it.  

The intention of the current implementation is that database threads would be assigned to logical cores that don't overlap with the physical cores  that the main viewer threads are running on.  So for a hyper threaded system one might have the main thread is only CPU 0, graphics thread on 2, which maps to physical cores 0 and 1 which need to be kept clear from database threads.  In this set up the ideal place to put the database threads is on pysical cores 2 and 3, which maps to logical cores 4,5,6,7.  This is what current scheme achieves for this particular CPU/viewer setup up combination.

However, in the set up you described with two camera threads the number of logical cores utilized wraps around and causes the logic to fail miserably.

What is required is a scheme that is aware of logical/physical set up and does a review of what the needs are for the viewer and how well this might map to the available physical/logical cores.  When we have more viewer threads then keep threads on separate physical cores is no longer possible so different schemes will be most optimal.  I haven't yet spent enough time on the topic to know what the schemes will need to be.

LaurensVoerman:
Thanks Robert, 
I now see where my confusion started. This PR can be closed - it's not an improvement over your current code. 
I don't think this can be done very well without a full hyperthreading detection. 
My attempts to detect hyperthreading on linux are currently stuck on the fact that CPUID tells me the processor is hyperthreading capable - but on my linux systems it's disabled in the bios. 
Regards, Laurens.

openscenegraph:
I'll leave the PR open as a reminder that this is an issue that needs resolving. 

The CPUID providing full CPU capabilities rather than active capabilities is a bit of issue.  What does the OpenThreads::GetNumOfProcessors() return?  Is this a correct number of cores active?

openscenegraph:
I am now closing this PR, but haven't yet implement hyper threading detection, I have put this on my TODO list.

PR #168, opened 2016-11-24
Fix IncrementalCompileOperation leaking program state

scrawl:
The IncrementalCompileOperation when compiling programs sets the GL state's current program object, but does not reset this state afterwards. A scene graph with no shaders would incorrectly render *with* a shader if you compiled some shaders in the background with the IncrementalCompileOperation.

Note that prior to OSG 3.5.4, this bug wasn't visible because it was masked by another bug - no program objects were compiled at all, see commits 8f68da89d7fb75264fde742d59ea888415b489a6 & 06cb31a3d265a63c6db693120abec3c9fac591f3.

Side note: is there any reason why we do

<pre>
    _program->apply(*compileInfo.getState());
</pre>
Instead of
<pre>
    _program->compile(*compileInfo.getState());
</pre>
Is that to workaround drivers that want to defer the compilation until the program is actually used? In that case, I don't think just setting the program object will do it, you'd need to draw something with it as well.

As a cleaner fix I'd propose changing apply() to compile(), unless we've got any test data that suggests using apply() would make a difference.

### DIFF
--- a/src/osgUtil/IncrementalCompileOperation.cpp
+++ b/src/osgUtil/IncrementalCompileOperation.cpp
@@ -725,7 +725,15 @@ void IncrementalCompileOperation::operator () (osg::GraphicsContext* context)
         }
     }
 
-    //glFush();
+    osg::State* state = context->getState();
+    if (state->getLastAppliedProgramObject())
+    {
+        osg::GLExtensions* extensions = state->get<osg::GLExtensions>();
+        extensions->glUseProgram(0);
+        state->setLastAppliedProgramObject(0);
+    }
+
+    //glFlush();
     //glFinish();
 }
 


openscenegraph:
Sorry the delay reviewing this change, It's taken me a while to get a long enough window to properly review the topic.  I've done code review and and am still learning/thinking about the issue and how best to tackle it.  

My current inclination is change the _program->apply(*compileInfo.getState()); to _program->compileGLObjects(*compileInfo.getState());  as I can't see a reason why apply is used instead in this context.  I suspect the relevant code was written before the Program::compile() was fully fleshed out, but since the code was written 6 years ago it's a bit too far back for me to know exactly what the thinking was at the time.

FYI, the State::setLastAppliedProgramObject() exists to allow osg::State::apply() to wire up uniforms to the correct program object.  The IncrementalCompileOperation doesn't need to do any work with uniforms so I can't see a need to set it, so calling osg::Program::apply() shouldn't be required.

I have checked in the change from Program::appply() to compileGLObjects() to git master, I don't have any paged database too hand which contain Program's so can't easily test this change.  Could you test it on your dataset, or post a link to your dataset so I can test it out.  If things work out OK I'll put this fix into OSG-3.4 and 3.2 branches.

openscenegraph:
Sorry the delay reviewing this change, It's taken me a while to get a long enough window to properly review the topic.  I've done code review and and am still learning/thinking about the issue and how best to tackle it.  

My current inclination is change the _program->apply(*compileInfo.getState()); to _program->compileGLObjects(*compileInfo.getState());  as I can't see a reason why apply is used instead in this context.  I suspect the relevant code was written before the Program::compile() was fully fleshed out, but since the code was written 6 years ago it's a bit too far back for me to know exactly what the thinking was at the time.

FYI, the State::setLastAppliedProgramObject() exists to allow osg::State::apply() to wire up uniforms to the correct program object.  The IncrementalCompileOperation doesn't need to do any work with uniforms so I can't see a need to set it, so calling osg::Program::apply() shouldn't be required.

I have checked in the change from Program::appply() to compileGLObjects() to git master, I don't have any paged database too hand which contain Program's so can't easily test this change.  Could you test it on your dataset, or post a link to your dataset so I can test it out.  If things work out OK I'll put this fix into OSG-3.4 and 3.2 branches.

scrawl:
Unfortunately I don't remember where I saw the issue, it wasn't a particular database but rather my entire program in a specific state.

Merging back to 3.2/3.4 doesn't seem super important, because of this bit:
> Note that prior to OSG 3.5.4, this bug wasn't visible because it was masked by another bug - no program objects were compiled at all, see commits 8f68da8 & 06cb31a.

But you can do it anyway if you feel like it, shouldn't cause any harm.

openscenegraph:
I don't think your PR is needed to be merged now I've made the change from Program::apply() to Program::compileGLObjects().  Could you review my change?

https://github.com/openscenegraph/OpenSceneGraph/commit/9478d272637aefc3d573dcb6c70c0f4dc039d9dd

If you feel that your change is no longer required let me know or close the commit. Thanks.

scrawl:
Yes I've seen it, looks perfect so will close this PR.

PR #173, opened 2016-12-13
Improvements for the OBJ writer plugin

etlapale:
Two improvements for the OBJ writer plugin are provided. The first allows support for a StateSet to be present in an osg::Group rather than directly associated to a node when converting to OBJ. It solves an issue where an empty `.mtl` file is generated. The second allows user to specify a precision when writing an OBJ file, rather than the default 6 digits. It also defaults to double max precision.


etlapale:
Wrong commit.

PR #176, opened 2016-12-15
cmake out of tree build

matomer:
mkdir build
    cd build
    # Shouldn't this be `cmake ..` ?
    cmake ../OpenSceneGraph

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ directory alongside the OpenSceneGraph use:
 
     mkdir build
     cd build
-    cmake ../
+    cmake ..
     make
     sudo make install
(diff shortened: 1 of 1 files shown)


openscenegraph:
I can only presume you have misread the text above the part of the README.md that you have read, the original text is correct. The only way for your modifications to be correct would be if you created the build directory inside the OpenSceneGraph directory, rather than alongside as is specified in the README.md and honoured in the cmake entries in this README.

matomer:
Sorry, that's correct I missed the part about creating the directory alongside OpenSceneGraph.

PR #178, opened 2016-12-21
fix weird looking stats on empty scene.

LaurensVoerman:


openscenegraph:
Using your fix as inspiration I've implemented something similar, with a small change of applying the setGlobalDefaults to the StatsHandler's _camera to be consistent with how osgViewer::View manages it's setGlobalDefaults.

The change I made is:
https://github.com/openscenegraph/OpenSceneGraph/commit/ff8b5536a56e5c9cb325a682862eaee7092c281c

PR #184, opened 2016-12-29
BUG FIX - Wrong calculation of terrain normals in GeometryTechnique

hwiesmann:
The partial derivative for the x-coordinate has been wrongly calculated: only a single sided derivative was calculated.


hwiesmann:
To remove a compiler warning that unused functions are present the two local functions have been blended out for OS X versions older than 10.6.

PR #186, opened 2017-01-07
Addition of documentation for LOD and PagedLOD

hwiesmann:
Documentation has been added for their default constructors. Furthermore, the consequences of different center modes have been explained. A comment regarding the setting of the radius has been fixed.

### DIFF
--- a/include/osg/LOD
+++ b/include/osg/LOD
@@ -36,6 +36,11 @@ class OSG_EXPORT LOD : public Group
 {
     public :
 
+				/** Default constructor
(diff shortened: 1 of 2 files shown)


hwiesmann:
What is the standard you require? No tabs or tabs equal to 4 spaces?

openscenegraph:
The OSG uses spaces for indentation so that the files look the same on all editors.  Your changes to spaces fixes the indentation issue so I think we a good to merge :-)

PR #187, opened 2017-01-12
Improve OSG_NOTIFY macro

rickyviking:
The current notification macros do not work (at least when compiled with Visual Studio) if used in if-else clauses without braces.

In the commit I've simply removed from the OSG_NOTIFY macro definition the isNotifyEnabled() check.
The same test is performed anyway in the notify() method called by the macro, and as added value the macro works fine in if/else clauses without braces.

To test the old and new behavior, I attach a modified version of osgversion application, with the macros locally redefined to reduce the compilation time.
[osgversion.zip](https://github.com/openscenegraph/OpenSceneGraph/files/701698/osgversion.zip)
This is just to show the purpose of the PR and not to be included in the commit.

### DIFF
--- a/include/osg/Notify
+++ b/include/osg/Notify
@@ -79,7 +79,7 @@ extern OSG_EXPORT std::ostream& notify(const NotifySeverity severity);
 
 inline std::ostream& notify(void) { return notify(osg::INFO); }
 
-#define OSG_NOTIFY(level) if (osg::isNotifyEnabled(level)) osg::notify(level)
+#define OSG_NOTIFY(level) osg::notify(level)
 #define OSG_ALWAYS OSG_NOTIFY(osg::ALWAYS)
 #define OSG_FATAL OSG_NOTIFY(osg::FATAL)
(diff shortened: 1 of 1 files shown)


robertosfield:
The if () statement is there to enable optimizers to compile out the code.

rickyviking:
Ah, but can the optimizer compile out such statement given that the singleton's _notifyLevel can be set and changed at runtime?

robertosfield:
Even without the optimizer having the if statement avoids setting up the ostream and any of the work passing the data to the ostream when the notification level of the message is the below the output level.

The optimizer come in when you compile the OSG with a fixed notification such that all messages are ignored, this is something you might do on an deployed system where messages are of no use. The compiler will see the if statement is always false and optimize away the notify statement completely so there is no CPU cost of member used.

rickyviking:
Got it.
Maybe then it's just worth to document that to use the macros within if/else blocks the braces are mandatory.

PR #188, opened 2017-01-15
osgGA : OrbitManipulator - disable mouse interacting with flag

realitian:


openscenegraph:
The StandardManipulator::_flags variable doesn't encompass general mouse interaction so it's inappropririate to use it as it does,

realitian:
Hi, Thanks for review it,
I had thought it was not good way, but I want the any way for flagging to disable only camera mouse interaction when picking object.
actually when I pick the object, the camera had flipping eventually which won't occur.
p.s. the platform was iOS.

PR #198, opened 2017-02-07
Add cmake options for disabling plugins from the build

scrawl:
Added option ```BUILD_OSG_DEPRECATED_SERIALIZERS``` (default ON)

Added option ```BUILD_OSG_PLUGINS_BY_DEFAULT``` (default ON). By default, all plugins are built and individual plugins can be disabled by setting ```BUILD_OSG_PLUGIN_<NAME>``` to 0. If ```BUILD_OSG_PLUGINS_BY_DEFAULT``` is set to 0, each plugin to be built must be enabled explicitely by setting ```BUILD_OSG_PLUGIN_<NAME>``` to 1.

The ```BUILD_OSG_PLUGIN_<NAME>``` options will not actually show up in cmake-gui - due to the sheer number of plugins I thought it was better to have a clean GUI by default and just manually add each one you need.

The motivation/use case for these switches is two fold. First of all, with plugins that depend on external libraries with constantly changing API it is hard to ensure they always build for everyone, and having the simple option to disable non required plugins will be welcome in such cases.

Second, I'd really like to see a reduction in build time when I'm testing osg changes for my app, and disabling plugins that are useless to me helps a lot.

Here are build timings for make -j4 (in both cases with BUILD_OSG_EXAMPLES=0 and BUILD_OSG_APPLICATIONS=0)

<pre>
default
12m56.614s
</pre>

<pre>
-DBUILD_OSG_PLUGINS_BY_DEFAULT=0 -DBUILD_OSG_DEPRECATED_SERIALIZERS=0 -DBUILD_OSG_PLUGIN_BMP=1 -DBUILD_OSG_PLUGIN_DDS=1 -DBUILD_OSG_PLUGIN_TGA=1 -DBUILD_OSG_PLUGIN_JPEG=1 -DBUILD_OSG_PLUGIN_PNG=1 

7m27.359s
</pre>

I've also been working on an option to disable non-essential Node Kits, which got me a further reduction in build time down to 3m56.085s. However I have not committed this yet due to some complications with examples/plugins that depend on node kits in a non optional way.


openscenegraph:
I will need more time to properly review this change  before I make any decision on it, so for now I'll just leave this PR and come back to it when I have time to do the proper review.  

One thing I did wonder of a quick review was whether we could have an optional user edited text file that can be used as input to decide what optional components of the OSG are compiled.I don't know where we can get CMake to parse a text file and set up BUILD_PLUGIN_x to OFF/ON based on the contents of thus text file.

scrawl:
Not sure what kind of text file you had in mind but you can edit CMakeCache.txt if that helps - that contains all settings you've passed in via command line or were otherwise cached by cmake.

aitormoreno:
You may use `cmake -C`:
[Check documentation](https://cmake.org/cmake/help/v3.6/manual/cmake.1.html)

> -C <initial-cache>
>     Pre-load a script to populate the cache.

PR #200, opened 2017-02-08
Move bindVertexArrayObject to osg::State to avoid null vao from being bound redundantly on every draw call

scrawl:
When not using VAOs, OSG would redundantly bind a null VAO before every draw call:

redundant_vao

To solve this, I moved the bind method into osg::State so we can check against the last bound VAO.

With this change applied my program is seeing a performance improvement in the draw thread of about 4%.

### DIFF
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -518,7 +518,7 @@ inline void Drawable::draw(RenderInfo& renderInfo) const
 
         State::SetCurrentVertexArrayStateProxy setVASProxy(state, vas);
 
-        vas->bindVertexArrayObject();
+        state.bindVertexArrayObject();
 
         drawInner(renderInfo);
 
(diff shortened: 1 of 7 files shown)


openscenegraph:
I understand the motivation behind the change but on a first pass it looks to me that there will be an additional reference overhead when using VAO's, so you've helped non VAO usage, but hindered VAO usage.  Once I have some time I will have  think about trying to avoid extra overheads in the two usage cases.  I'll leave the PR up as reminder that the issue is unresolved, but for now not feel the PR is the best way to address the issue.

scrawl:
I can't think of a way to get around the if test. Would it help performance to set up the glBindVertexArray function in the State constructor, as its done for other functions, and then do ```_glBindVertexArray(vao)``` instead of ```_glExtensions->glBindVertexArray(vao);``` ?

openscenegraph:
I have just checked in similar changes to this PR, which not identical will achieve the same result.  https://github.com/openscenegraph/OpenSceneGraph/commit/ffbc1167de4d3761a4c7574add76db50e3098901

PR #207, opened 2017-02-24
fix OpenCascade library list for use with shared libs; added list and…

LaurensVoerman:
fix OpenCascade library list for use with shared libs
added list and options for static lib

btw: this is not working with debug libs, not tested as I did not yet try to build the opencascade in debug config.


openscenegraph:
Hi Laurens, I had took quite a while to work out which dependencies were key to critical to building against OpenCascade so am surprised you have come across a different set, removing some that I found necessary and adding ones that I didn't need.  This makes me wonder if there is a a versioning issue at play beyond static vs dynamic libs.   

What platform are you building on and what version of OpenCascade?

I'm building with Kubuntu 16.04 and OpenCascade built from source, version is 7.1.0 beta.

As I'm concerned merging your changes will break the build for those building with the same OpenCascade version as myself I'll need to create a branch a test your changes out on it.

LaurensVoerman:
Hi Robert,
I am working on windows 10 , with visual studio 2015 (=vs14 update 3), and
I compiled OpenCascade 7.1.0 from source.
according to their web page this version has "Publication date:
25 November 2016" , I found no indication this is a beta version.

I can find that for the dynamic link list the last two added library's are
called from the private and never used function _healShape, that could
explain why g++ won't need those,
OPENCASCADE_TKTopAlgo, OPENCASCADE_TKShHealing.


the other new library's seem to provide required functions:
OPENCASCADE_TKLCAF OPENCASCADE_TKBRep OPENCASCADE_TKMesh

as to the four lib I commented out for dynamic linking,
OPENCASCADE_TKSTEPAttr OPENCASCADE_TKSTEPBase OPENCASCADE_TKSTEP209
OPENCASCADE_TKSTEP
I suppose they should be reinstated - the linker will find out if they are
not needed. They need to be build anyway because TKXDESTEP.dll depends on
them.

So I'll change the commit, re-enabling the lines:
#    ${OPENCASCADE_TKSTEPAttr}
#    ${OPENCASCADE_TKSTEPBase}
#    ${OPENCASCADE_TKSTEP209}
#    ${OPENCASCADE_TKSTEP}

Regards, Laurens.

openscenegraph:
I have tested on my Linux system and everything builds and runs fine so it looks like my concerns were unfounded, so have gone ahead an merged. Cheers.

PR #213, opened 2017-03-06
Bug/osg fixes

marchelbling:
Hi Robert, please find

* Compilation fixes:
    * inventor plugin where a commented variable is asserted
    * implicit conversion from `ref_ptr` ro raw pointer (when using cmake flag `REF_PTR_IMPLICIT_OUTPUT_CONVERSION`)
* Bug fixes:
    * 3ds plugin read (see http://forum.openscenegraph.org/viewtopic.php?t=16482 for discussion)
    * missing array types in array merge (typically an issue when running the optimizer on scenes that have geometries with vertex color)
* Clean up:
    * remove usage of `offset` attribute in `MergeArrayVisitor`; this might be needed in non-public code so I may remove this commit if you prefer)

### DIFF
--- a/src/osg/Geometry.cpp
+++ b/src/osg/Geometry.cpp
@@ -133,7 +133,7 @@ void Geometry::configureBufferObjects()
         itr != arrays.end();
         ++itr)
     {
-        osg::Array* array = *itr;
+        osg::Array* array = itr->get();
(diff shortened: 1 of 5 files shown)


filnet:
+1 for the 3d fix that addresses http://forum.openscenegraph.org/viewtopic.php?t=16482

PR #214, opened 2017-03-06
fix bug in osgb/osgt/osgx readerwriter osgParticle

LaurensVoerman:
causing the datavariance in  osgParticle-SizeInterpolator and -ColorInterpolator to disappear.


LaurensVoerman:
just like the commit for the head revision I reverted the change on the code line writing the TextureTile properties:
REMOVED CHANGE:  line 112
-    os << os.PROPERTY("TextureTile") << p.getTileS() << p.getTileT() << p.getNumTiles() << std::endl;
+    os << os.PROPERTY("TextureTile") << p.getTileS() << p.getTileT() << p.getNumTiles() -1 << std::endl;

PR #215, opened 2017-03-06
Submit serializer particle

LaurensVoerman:
read endbracket for fileversion 145:
osgconv spaceship.osgt spaceship2.osgt; osgconv spaceship2.osgt spaceship3.osgt; osgviewer spaceship3.osgt 
shows the partice engines have stopped working, due to the missing read of an endbracket in file version 145 and up.

### DIFF
--- a/src/osgWrappers/serializers/osgParticle/Particle.cpp
+++ b/src/osgWrappers/serializers/osgParticle/Particle.cpp
@@ -36,7 +36,7 @@ bool readParticle( osgDB::InputStream& is, osgParticle::Particle& p )
     if ( hasInterpolator )
     {
         is >> is.BEGIN_BRACKET;
-        p.setAlphaInterpolator( is.readObjectOfType<osgParticle::Interpolator>() );
+        p.setSizeInterpolator( is.readObjectOfType<osgParticle::Interpolator>() );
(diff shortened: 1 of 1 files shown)


LaurensVoerman:
Hi Robert,
I stuck it in there to stop the number from increasing on every read /
write, changing the file diff while I was working on the other stuff. I
forgot to follow up on that, the problem might be on the reader side or
with strange variable naming.
I think it's the best option to leave the -1 there for now, making the
output match the input without changing the result of existing files, and
maybe later check and fix what's actually going on.
Regards, Laurens.

openscenegraph:
We'll need to get to the bottom of the -1 before merging as I don't wish to introduce a bug to hide a bug, the right fix will be to find the original bug and fix it and not muddy the waters by introducing a hack.

LaurensVoerman:
Agreed,
I removed the change on line 117:
-    os << os.PROPERTY("TextureTile") << p.getTileS() << p.getTileT() <<
p.getNumTiles() << std::endl;
+    os << os.PROPERTY("TextureTile") << p.getTileS() << p.getTileT() <<
p.getNumTiles() -1 << std::endl;


Regards, Laurens.

PR #219, opened 2017-03-15
Ensure no declarations appear before an #extension directive

manisandro:
Mesa refuses to compile sharers where `#extension` directives appear somewhere in the middle of the shader text:

    error: #extension directive is not allowed in the middle of a shader

The relevant part of the PR is just

    +    // move insPos to next line as long as line begins with an #. This ensures declarations appear
    +    // after any possible #extension directives
    +    for(int sourceSize = source.length(); insPos < sourceSize && (source[insPos] == '#' || source[insPos] == '\n'); ) {
    +      insPos = source.find( '\n', insPos) ;
    +      insPos = insPos != std::string::npos ? insPos+1 : sourceSize;

The rest is just renaming `declPos` to `insPos` since it is more correct.

For reference, the Khronos OpenGL wiki states [1]: "You should put these [#extension] definitions before any other language features, but after the version declaration"

[1] https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#Extensions

### DIFF
--- a/src/osg/State.cpp
+++ b/src/osg/State.cpp
@@ -1193,11 +1193,11 @@ namespace State_Utils
         return replacedStr;
     }
 
-    void replaceAndInsertDeclaration(std::string& source, std::string::size_type declPos, const std::string& originalStr, const std::string& newStr, const std::string& qualifier, const std::string& declarationPrefix)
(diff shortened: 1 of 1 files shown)


manisandro:
Argh never mind, code got added to handle this since 3.5.5 which I was testing against.

PR #221, opened 2017-03-18
GLES3 support improvements on iOS

tomhog:
Here are changes to get gles3 working stand alone on iOS. There was some work done previously but it worked by also including gles2 which I don't want. I've tried to keep support for that approach by setting the OPENGL_PROFILE var to GLES2+GLES3 but I haven't tested it yet. I have tested GLES1, GLES2 and GLES3 individually

It's mostly just adding ifdefs for OSG_GLES3_AVALIABLE where it was already being used for GLES2.

Also made improvements to the iphone viewer example, has support for GLES1, 2 and 3 with some basic demo shaders etc.

Updated the build instructions in the readme


examples/osgviewerIPhone/iphoneViewerAppDelegate.mm
```diff
@@ -33,10 +41,19 @@ @implementation iphoneViewerAppDelegate
 
 @synthesize _window;
 
+//
```
openscenegraph: This change suggest to me that we need to change the osg::BuildShapeGeometryVisitor implementation so it does generate GL_QUADS to start with.  Could you post an this problem as an github Issue for the OSG to make sure it's not missed.

examples/osgviewerIPhone/iphoneViewerAppDelegate.mm
```diff
@@ -175,32 +199,41 @@ virtual bool handle (const osgGA::GUIEventAdapter &ea, osgGA::GUIActionAdapter &
...
                         case osgGA::GUIEventAdapter::TOUCH_STATIONERY:
-                            _drawables[j]->setColor(osg::Vec4(0.8,0.8,0.8,1));
+                            color = osg::Vec4(0.5,0.5,0.5,1);
                             break;
                             
                         default:
                             break;
                             
                     }
+#if defined(OSG_GLES2_AVAILABLE) || defined(OSG_GLES3_AVAILABLE)
```
openscenegraph: This change ties this code to a specific shader implementation that uses color as a uniform.  The osg::State class has inbuilt support for aliasing fixed function vertex attributes like gl_Vertex, gl_Color etc. to osg_Vertex, osg_Color etc, my suggest would be that osg_Color should be used in the shader instead, rather than hardwiring other code to a particular shader.

examples/osgviewerIPhone/shaders.h
```diff
@@ -0,0 +1,113 @@
+#pragma once
```
openscenegraph: I feel that it would be more consistent to use the osg_Color in the vertex shader and pass this as a varying to fragment shaders rather than use a custom uniform.

tomhog: Hi Robert

Thanks for getting to this so quickly.

To start with I did try using osg_Color in the shader programs but osg Shape and Text use a single color vertex with the BIND_OVERALL mode. This isn't supported on GLES you have to have a color value per vertex. So I went for this color uniform approach which isn't ideal but at  it works.

I'm struggling to find a reference that states the BIND_OVERALL won't work, but i'm not 100% on what it does in terms of OpenGL calls, had a quick dig but got lost in array dispatcher land.

openscenegraph: There are wider problems is BIND_OVERALL isn't functioning correctly, these need to be looked into and resolved.  Just checked GLES3 specs and the required glVertexAttrib4fv function is there:

https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glVertexAttrib.xhtml

It is worth noting that I am current working on changes to osgText that will provide better out of the box support for osgText on GL3 and GLES2/3.  It may be that all your custom code for osgText will be not required.  Keep an eye on the textrefactor_3 branch.

For this PR I would suggest removing the shader/color/text specific changes from the PR so there aren't any items that are controversial so it can be merged as is.  We then can work towards resolving the final issues, such as making sure the new osgText branch works fine.

tomhog: Looks like I may have been suffering from sleep deprivation as now the Bind overall color seems to be working. 

I'll get the shaders changed and retest gles2 and 3 thanks for the help

tomhog: Ok I've made changes to use vertex colors and tested gles2 and 3 all working. Sorry about that, nice it's working closer desktop now though.

openscenegraph: Thanks for the revisions. End of day here so I'll have a look on Monday.

I have just checked in the GL3+GL2 shaders into osgText, to be found in src/osgText/Font.cpp.  I'm hoping these will work for GLES3+GLE2.  These are now in textrefactor_3 branch

openscenegraph: I am beginning another review, I've checkout out a branch.  One element I feel needs to be looked at is the need for custom shaders for text.  I have now checked in the text refactor changes into master so the osgText now sets up shaders for GLE2/GLES3 as GL3 core profile. I think we should be able to avoid some of the changes you've made.  Could you try your changes out on master and see if you can roll back the local shader additions.

tomhog: Those changes worked and I've removed the local text shaders from the example.

I did also make it so GLES3 uses the GL3_SHADER path in the Font shader code. For this to work I just had to define a different version for GLES and also make it swap the texture component used from the glyph texture. Hope that's not a step too far, just wanted to properly test gles3. 

The gl2 shader path did work for both gles2 and gles3

tomhog: looks like the travis build failed due to max time limit being reached. Is that something you can re trigger? can't see the changes I made making the build take any longer than usual

openscenegraph:
Just done another code review of the the latest changes.  Now looks good.  And will go ahead and merge.  Thought I'd just add one last.  Oh my goodness, what a huge amount of work has gone into this! Kudos for getting the GLES3/iOS port done :-)

PR #223, opened 2017-03-20
Serializer for osg::TextureBuffer and osg::Texture2DMultisample

MPursche:
Hello,

I implemented serializers for the TextureBuffer and Texture2DMultisample classes.
The TextureBuffer serializer supports buffer data that is stored as osg::Image or osg::Array which should be sufficient for most use cases.

Greetings
Marcel

### DIFF
--- a/src/osgWrappers/serializers/osg/TextureBuffer.cpp
+++ b/src/osgWrappers/serializers/osg/TextureBuffer.cpp
@@ -48,7 +48,7 @@ static bool writeImage( osgDB::OutputStream& os, const osg::TextureBuffer& textu
 
 static bool checkBufferData( const osg::TextureBuffer& texture )
 {
-    return texture.getBufferData()->asArray() != NULL;
+    return texture.getBufferData() != NULL && texture.getBufferData()->asArray() != NULL;
 }
 
 static bool readBufferData( osgDB::InputStream& is, osg::TextureBuffer& texture )
--- a/src/osgWrappers/serializers/osg/TextureBuffer.cpp
+++ b/src/osgWrappers/serializers/osg/TextureBuffer.cpp
@@ -1,0 +1,98 @@
+#include <osg/TextureBuffer>
+#include <osg/Array>
+#include <osgDB/ObjectWrapper>
+#include <osgDB/InputStream>
+#include <osgDB/OutputStream>
+
+static bool checkImage( const osg::TextureBuffer& texture )
+{
+    return texture.getImage() != NULL;
+}
+
+static bool readImage( osgDB::InputStream& is, osg::TextureBuffer& texture )
+{
+    if ( is.isBinary() )
+    {
+        osg::ref_ptr<osg::Image> image = is.readImage();
+        texture.setImage(image.get());
+    }
+    else
+    {        
+        is >> is.BEGIN_BRACKET;
+        osg::ref_ptr<osg::Image> image = is.readImage();
+        texture.setImage(image.get());
+        is >> is.END_BRACKET;    
+    }    
+    
+    return true;
+}
+
+static bool writeImage( osgDB::OutputStream& os, const osg::TextureBuffer& texture )
+{
+    const osg::Image* image = texture.getImage();    
+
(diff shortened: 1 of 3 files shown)


src/osgWrappers/serializers/osg/TextureBuffer.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    return true;
+}
+
+REGISTER_OBJECT_WRAPPER( TextureBuffer,
+                         new osg::TextureBuffer,
+                         osg::TextureBuffer,
+                         "osg::Object osg::StateAttribute osg::Texture osg::TextureBuffer" )
+{
+    ADD_USER_SERIALIZER(Image);            // _bufferData as osg::Image
```
openscenegraph: Would it not be possible to just write out BufferData as a object, and have BufferData serializers as this would avoid overlap between writing out BufferData as an image as well as writing it out as an array.

MPursche: I looked into using the _osg::InputStream::readObject_/_OutputStream::writeObject_ methods:
It seems to work fine for Array data but there is an issue if _osg::Image_ is used. The serializer for _osg::Image_ does not store the image data itself if storing data inline is enabled. The logic for this is implemented in the _OutputStream::writeImage_ method and not in the serializer itself. Storing image data inline comes in handy if special image formats are used(float textures, integer textures).
So it would be necessary to adapt the serializer for osg::Image as well.

openscenegraph: What is the issue with writeImage()?  It uses the osg::Image::getWriteHint() to decide whether to write out a separate or inline image data.  If you want the image to be inlined then you just need to set the hint. From include/osg/Image
       enum WriteHint {
            NO_PREFERENCE,
            STORE_INLINE,
            EXTERNAL_FILE
        };

        void setWriteHint(WriteHint writeHint) { _writeHint = writeHint; }
        WriteHint getWriteHint() const { return _writeHint; }

MPursche: There is no issue with the writeImage() method. Setting the write hint works as expected when this method is used.
The problem lies in the writeObject() method. I tried to simplified the code in the pull request by using readObject() and writeObject() to read and write the bufferData for both cases as you suggested. This works for Arrays but not for Images as the image data is not written by the method. Image data is only written if writeImage() is used.

openscenegraph: I'm not the original author of the serializers so will need to look up the code to find out the reasons that it's been done this way, my current inclination is that we should expand the Image serializer so that it can handling inlining the data, this may require using the scoped versioning to achieve this.  I believe we need to fix this problem properly rather than introduce a workaround into the TextureBuffer serializer.

Could you split out the two serializes from your PR into two PR's.  The Texture2DMultisample part looks good to merge right away.  The TextureBuffer one we need to resolve the issues with serailizing images first, then this will simplify the TextureBuffer serialize.

openscenegraph:
I have now merged a prelimary osg::TextureBuffer serializer : 

https://github.com/openscenegraph/OpenSceneGraph/commit/e74f8f4bbe4853936c88aefaacf9b2c034479e46

I have left it simple and just serialized the TextureWidth value and BufferData object. I modified the osgforest example to use and write out the TextureBuffer with an osg::Vec4Array instead of an osg::Image and the serialization worked fine.

However when using the original osg::Image usage with TextureBuffer the serialization just serialized the image it doesn't write out any internal data even when I set the WriteHint to STORE_INLINE.  Clearly this is a bug/unimplementated feature in the serializer support for inlined Images.  This is a separate topic the TextureBuffer though so I think logging this as an issue to fix is the right thing to do rather than workaround this issue by placing code into the TextureBuffer serializer.

openscenegraph:
I have now created an github Issue for the problem with serialization of osg::Image so will close this PR now as once the osg::Image serialization issues are resolved that new TextureBuffer serialization should work without modifications.

PR #231, opened 2017-03-27
BufferObject/BufferData Serializers and involved updated arrays serializers too

mp3butcher:
here are serializers for BuffererData/BufferObject as well as involved Arrays
to integrate consider adding osg::BufferData as osg::Array serializer
i don't know if dso is good due to this commit i've done to integrate associate version management (i keep these changes for a future pr)
https://github.com/mp3butcher/OpenSceneGraph/commit/f0ae9d71608b07c7b5a88954f8f5a3807d7e21cc
Further for the bo serializer to be usefull, you should consider removing the osg::Geometry FinishLoadingCallback that forbid BufferObject serialization (it reset all arrays's bos to null)

### DIFF
--- a/src/osgWrappers/serializers/osg/Array.cpp
+++ b/src/osgWrappers/serializers/osg/Array.cpp
@@ -34,8 +34,14 @@ struct ResizeArray : public osgDB::MethodObject
 REGISTER_OBJECT_WRAPPER( Array,
                          0,
                          osg::Array,
-                         "osg::Object osg::Array" )
+                         "osg::Object osg::BufferData osg::Array" )
 {
+
+
+    {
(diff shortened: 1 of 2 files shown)


mp3butcher:
Perhaps an osgDB option would be a solution in order to (or not) serialize arrays bufferobject configuration.
Is it worth i code it?

openscenegraph:
I have done a full review of the changes and didn't feel that implement serialization for BufferObjects the most appropriate way. 

The commit is:
https://github.com/openscenegraph/OpenSceneGraph/commit/70e7cb886bec4427e7da50422163c9ee7f756268

This commit should allow users to set the BufferObject settings and sharing parameters.  If I've missed something please raise it on the osg-users mailing list/forum so we can get to the bottom of what's missing.

PR #241, opened 2017-04-05
add a fix in lua (seams index in lua begin at 1)

mp3butcher:
add matrix value object support too..
NB I have removed type control in getelements(int pos, int numElements, int type) const for debug purpose
perhaps it should be uncomment but can't remember if lua_pop affected by the index stuff


openscenegraph:
Could you split the index offset change from the matrix fix.  

The index offset issue is one I considered during implementation of the LuaScriptEngine and decided that being consistent with the OSG was more important than being consistent with Lua.  My expectation is that most OSG/lua users will be primarily OSG C++ users and use lua just for light scripting.  All OSG documentation is also using standard 0 as the base of all arrays.  Gven these factors I don't feel it's best to change the indexing across to Lua conventions when accessing OSG data structures.

mp3butcher:
The two are linked and can't be separated: if you don't integrate the index fix, matrix value object are wrong....

openscenegraph:
I've just explained my reasoning for trying to maintain the same indexing for OSG classes in C++ as for the same class/objects in Lua, I don't know what more I can say.  If you disagree fine, but I'm not going to go merge something that I considered carefully and decided against during the design and implementation.

mp3butcher:
My point was just to fix a bug: getElements fails systematically
If you want to keep it..It's your decision

PR #247, opened 2017-04-19
fix osg::Image::computePixelFormat and optimize osgexample//bindlesstext

RealityTracer:


openscenegraph:
I've just done a code review of the changes and it looks like something that's mid debugging and not ready at all for merging.  There is debugging code in there that shouldn't be needed at all, and suggestive that you've done something odd with your build.  The indentation is all over the shop.  Long winded comments that are only relevant for debugging notes.  Replacing the use of ref_ptr<> with C* is also a red flag, and is suggestive that your understanding of how memory should be managed needs a bit more refining.  This is topic for discussion on osg-users mailing list rather than within PR's as PR's should all be able submissions that are well tested and ready for merging with major changes.

I would also suggest in future PR's keep different sets of changes as separate PR's - if the changes are orthogonal to each other there is no reason to lump them together.  This is important as not all PR's are ready to be merged, while others can be merged as is without delay.

PR #248, opened 2017-04-22
judge the ''GLBufferObject' if really is a VBO'

RealityTracer:
when we attach an other type GLbufferobject such as SSAO UBO etc to an array of a geometry, the orginal src of "setArray" will bind this non-vbo glbufferobject as vbo,then APP crashing occurs no matter  whether the geometry open the _useVertexBufferObjects flag.So I fixed this bug to manage it ignoring if the _useVertexBufferObjects flag is true or false.

### DIFF
--- a/src/osg/VertexArrayState.cpp
+++ b/src/osg/VertexArrayState.cpp
@@ -663,8 +663,18 @@ void VertexArrayState::setArray(ArrayDispatch* vad, osg::State& state, const osg
             GLBufferObject* vbo = isVertexBufferObjectSupported() ? new_array->getOrCreateGLBufferObject(state.getContextID()) : 0;
             if (vbo)
             {
-                bindVertexBufferObject(vbo);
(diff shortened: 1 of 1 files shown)


RealityTracer:
I test and check this bug in my customed osg3.4 version, which is fixed in osg::State::setXXXPointer. After modifying, everything goes ok in my env.Then I go to VertexArrayState inosg-master,which only needed to be modified in setArray function in my eye.I have not tested the fixed dlls in osg-master.

openscenegraph:
I am rejecting this submissions as a dynamic_cast<> in code that can get called many thousands of times a frame would destroy performance - it's *really* bad solution.  If you have a problem with your particular usage case then please discuss this on the osg-users mailing list/forum, once we understand and can reproduce the issue we should be able to come up with a sounder solution that hamper performance for everyone just because a niche usage case didn't get to work in particular circumstances.

PR #250, opened 2017-04-24
Turned off lighting in examples/osggeometry

cxw42:
Long-time lurker, first-time contributor.  This is fairly minor, but I noticed that the image and the points in examples/osggeometry are very hard to see if you view them close to edge-on.  This PR turns off lighting for both of those geodes so they are equally visible at any angle.  I hope this helps!  Please let me know if you have any questions or comments.  Thanks!

### DIFF
--- a/examples/osggeometry/osggeometry.cpp
+++ b/examples/osggeometry/osggeometry.cpp
@@ -510,8 +510,10 @@ osg::Node* createScene()
         geode->addDrawable(polyGeom);
     }
 
+    geode->getOrCreateStateSet()->setMode(GL_LIGHTING,
+                    osg::StateAttribute::OFF | osg::StateAttribute::PROTECTED);
     return geode;
-}
+} //createScene()
 
 
 // define a node callback to animate a transform as a cycle along the y axis, between 0 and 2.0.
@@ -543,9 +545,10 @@ class MyTransformCallback : public osg::NodeCallback
 
         float               _angular_velocity;
 
-};
+}; //class MyTransformCallback
 
-
+/// Create a quad that sits behind the geometry from createScene()
+/// and shows the image from the OpenGL docs that geometry duplicates.
 osg::Node* createBackground()
 {
 
@@ -625,22 +628,22 @@ osg::Node* createBackground()
     polyGeom->setStateSet(stateset);
 
 
-    // create the Geode (Geometry Node) to contain all our osg::Geometry objects.
+    // create the Geode (Geometry Node) to contain the quad's osg::Geometry
     osg::Geode* geode = new osg::Geode();
 
     // add the points geometry to the geode.
     geode->addDrawable(polyGeom);
 
-    //return geode;
+    geode->getOrCreateStateSet()->setMode(GL_LIGHTING,
+                    osg::StateAttribute::OFF | osg::StateAttribute::PROTECTED);
 
     // create a transform to move the background back and forward with.
-
     osg::MatrixTransform* transform = new osg::MatrixTransform();
     transform->setUpdateCallback(new MyTransformCallback(1.0f));
     transform->addChild(geode);
 
     return transform;
-}
+} //createBackground()
 
 int main(int, char **)
 {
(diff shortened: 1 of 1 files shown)


examples/osggeometry/osggeometry.cpp
```diff
@@ -510,8 +510,10 @@ osg::Node* createScene()
         geode->addDrawable(polyGeom);
     }
 
+    geode->getOrCreateStateSet()->setMode(GL_LIGHTING,
+                    osg::StateAttribute::OFF | osg::StateAttribute::PROTECTED);
     return geode;
-}
+} //createScene()
```
filnet: not sure the osg coding convention requires to comment method closing braces

examples/osggeometry/osggeometry.cpp
```diff
@@ -543,9 +545,10 @@ class MyTransformCallback : public osg::NodeCallback
 
         float               _angular_velocity;
 
-};
-
+}; //class MyTransformCallback
 
+/// Create a quad that sits behind the geometry from createScene()
+/// and shows the image from the OpenGL docs that geometry duplicates.
```
filnet: what is the meaning of triple slashes ?

cxw42: filnet, I was looking at the [doxygen manual](https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#cppblock) recently and noticed that `///` works the same as `/**` to introduce a documentation block.  My personal style is to avoid `/* */` so that I don't have any nested-comment issues, so I decided to use `///` going forward in my own code.  If you prefer `/**`, just let me know --- I'm happy to change it.

examples/osggeometry/osggeometry.cpp
cxw42: filnet, thanks for looking at this!  Happy to change it if you wish.  I have personally found this sort of comment helpful, especially when using `less` or another viewer that doesn't have jump-to-matching-brace.

examples/osggeometry/osggeometry.cpp
robertosfield: I dislike commenting close brackets as it really doesn't help with readability when you have a modern editor as it'll match brackets for you, and takes away from the clarity of the actual functional code that is doing stuff - the important stuff for understanding code.  I review lots of code so clarity is really important.

As for disabling lighting, I think this is personal preference, the example is meant to be about creating osg::Geometry rather than state, other examples exist to manage state so I don't see a particular value in adding stuff for the example to beautify the on screen result from the perspective from one user.

cxw42: I agree that disabling lighting is not related to the functionality of `osg::Geometry`.  I am thinking back to about two years ago, when I was just starting to learn OSG.  I remember being very confused, as a raw newbie, about what I now know is the functionality split between `Geometry`, on the one hand, and `osgViewer::Viewer` plus the automatic light source indicated by `osg::View::LightingMode`, on the other.  I think removing lighting from this example would have reduced my confusion, or at least told me there was more going on that I needed to look into.

My apologies for not being more descriptive in the PR text.  Regardless, thanks very much for your consideration!

cxw42:
Would 1ee01b6  be more acceptable?  I removed the closing-brace comments and added some descriptive text clarifying that the lighting change is not part of the `Geometry` proper.  I'm trying to put myself back in my newbie mindset from a couple of years ago and imagine what would have been helpful to me then.  If this doesn't do it, I'll let the matter drop and consider that I have learned some useful things about the project :) .

examples/osggeometry/osggeometry.cpp
filnet: cxw42 thanks for the clarification about /// (I did not know about it). I agree about nested /* */.
I also promote the use of // vs /* */ at work.

cxw42:
@filnet  @robertosfield  Any thoughts, yea or nay?  If nay, I won't worry about keeping this branch up to date with master any more :) .

openscenegraph:
I have been busy completing other work so have put aside reviewing submissions over the past week, just wrapped this up so will get back reviewing submissions now.

PR #251, opened 2017-04-25
osgText::Glyph: Explicitly set GL_TEXTURE_MIN_FILTER to address disap…

emminizer:
…pearing glyphs on some Intel cards.

### DIFF
--- a/src/osgText/Glyph.cpp
+++ b/src/osgText/Glyph.cpp
@@ -373,6 +373,7 @@ void GlyphTexture::apply(osg::State& state) const
             glyphsWereSubloading.clear();
 
             glPixelStorei(GL_UNPACK_ALIGNMENT,1);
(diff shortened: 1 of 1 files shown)


openscenegraph:
I am about to merge the fix, but as this applies to only the OpenSceneGraph-3.4 branch the question is open about what to do about OSG master.  The osgText master has a seen a major rewrite to has affected the texture setup, the relevant part to this PR has been simplified and whether the changes in this PR to help with Intel chips/drivers will be required or not is not resolved.  Could you try out OSG master your system and see if things work.  If things work or not could you let me know.

emminizer:
Hi Robert.  I just tested master.  It appears to be OK without this change (which is required in 3.4.0).  So I think this issue can be considered completely resolved, at least from my perspective.

openscenegraph:
Thanks for the testing, good to hear master is working fine for you.

PR #252, opened 2017-04-28
Avoid call of SetupThreads in DatabasePager Constructor 

remoe:
This solves a bug with some shared library configurations. Sometimes the DatabasePager Singleton is initialised before one can call other configuration options like "setNumOfDatabaseThreadsHint". But OSG initialise it here:

https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osgDB/DatabasePager.cpp#L1555 

So this call in the constructor is not needed.

### DIFF
--- a/src/osgDB/DatabasePager.cpp
+++ b/src/osgDB/DatabasePager.cpp
@@ -1110,10 +1110,6 @@ DatabasePager::DatabasePager()
     // test of setting the database threads affinity.
     // _affinity = OpenThreads::Affinity(4,4);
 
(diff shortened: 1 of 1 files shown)


remoe:
This call instantiate the first DatabasePager in the app, before anyone can setup the configuration:

https://github.com/openscenegraph/OpenSceneGraph/blob/10d8ac7f65b177a3a92a28e2f976413c9c641489/src/osgDB/DatabasePager.cpp#L1243

PR #254, opened 2017-05-04
Topic gles3 ios

tomhog:
Hi

Here are a few more tweaks to GLES3 build, turned out I had forgotten to test texturing but noticed it wasn't working once osgText was switch to use the standard Texture system.

Also some minor improvements to the iOS example


tomhog:
Hmm seems the travis build timed out

tomhog:
Thanks

PR #256, opened 2017-05-18
Open scene graph 3.4

JasonWinston:


robertosfield:
What are doing with this PR Jason?  I can't work out what you mean by a massive PR, with no commentary.  Did you make this PR by mistake?

openscenegraph:
HI Jason, I'm rejecting this PR as it just looks like a massive merge of modifications that are in master in the OpenSceneGraph-3.4 branch.  This really isn't appropriate.  

The only changes that are appropriate for merging with the OpenSceneGrah-3.4 branch are ones that have close to zero risk of breaking existing build+functionality and do not affect the ABI.  One has to do make decisions about this on a per commit basis, not wholesale.  So if you want a specific commit from master merged into OpenSceneGraph-3.4 then make a PR for this, or bring up the topic for discussion on the osg-users mailing list/forum as this the official route for support.

PR #267, opened 2017-06-22
Add osgemscripten example

kornerr:
This is a working example to display a red cube with OSG in a web browser using Emscripten.

The example builds OpenSceneGraph as a static library. The example uses SDL to create rendering window and OpenGL graphics context.

The pull request also modifies GLExtensions file to use SDL2, which is shipped with Emscripten, to get OpenGL function addresses.


openscenegraph:
Interesting work, I look forward to diving in and playing with it.  I haven't tried Emscripten before so likely will need to assistance with trying things out.  I might be best to fire up a discussion on osg-users mailing list/forum so that others can pitch in with testing and general discussion on the topic.

RIght now I'm rushed off my feet so don't have time at the movement to dive in with testing of this PR.  Unfortunately it won't be till mid July before I'm able to devote any a decent amount of time on the open source side as I've got a full on week of client work then a two week family holiday.

kornerr:
Here's a step-by-step instructions (they are for the same example, but come from the guide sources): https://github.com/OGStudio/openscenegraph-cross-platform-guide/tree/master/1.10.SampleWeb

Here's the result (open it!): https://ogstudio.github.io/openscenegraph-cross-platform-guide

openscenegraph:
Thanks for the link, I've now merged the changes so step 1.10.5 of the above guide is now redundant :-)  

I will tag a dev release today to make it easier to specify which version of the OSG onwards will support your emscripten work.

kornerr:
Great! Thanks!

PR #268, opened 2017-06-26
Corrected Image::getColor method when using texture coordinates

dliepelt:
Please correct me if I'm wrong, but I think there is a tiny error in the Image::getColor method when using texture coordinates for reading pixel values.

Let’s assume we would only have 2x2 pixel sized image, we get the following structure if the lower left corner is the origin:

p2 p3
p0 p1

According to the current implementation the computed pixel coordinate is always 0 for any texture coordinate in the range 0.0 to < 1.0.
`unsigned int s = osg::clampTo(int(texcoord.x()*float(_s-1)), 0, _s-1);`
For example: s = 0.9  * (2 -1=1) = 0.9 -->0 after float to int conversion. 
This means Image::getColor() will always return p0 in this case, however, I would expect to obtain pixel p3 when choosing a texture coordinate with U and V < 0.5.

If do the same with a 3x3 image...

p6 p7 p8
p3 p4 p5
p0 p1 p2

my assumption would be again, choosing a texture coordinate UV(> 0.66, >0.66) should return p8
s = 0.7 (3-1=2) = 1.4 --> 1 after float to int conversion, so we would get p4 instead.

### DIFF
--- a/src/osg/Image.cpp
+++ b/src/osg/Image.cpp
@@ -1937,9 +1937,9 @@ Vec4 Image::getColor(unsigned int s,unsigned t,unsigned r) const
 
 Vec4 Image::getColor(const Vec3& texcoord) const
 {
-    unsigned int s = osg::clampTo(int(texcoord.x()*float(_s-1)), 0, _s-1);
-    unsigned int t = osg::clampTo(int(texcoord.y()*float(_t-1)), 0, _t-1);
-    unsigned int r = osg::clampTo(int(texcoord.z()*float(_r-1)), 0, _r-1);
+    unsigned int s = osg::clampTo(int(texcoord.x()*float(_s)), 0, _s-1);
+    unsigned int t = osg::clampTo(int(texcoord.y()*float(_t)), 0, _t-1);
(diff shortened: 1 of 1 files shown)


openscenegraph:
Unfortunately I don't have time to look into the issue at present as I'm just wrapping up before a two week trip.  A quick answer would be that how to interpret the mapping correctly would depend upon the wrap mode being used in the texture associated with the osg::Image.  So to handle things correctly their should be multiple get methods for each wrapping mode, or to pass in the wrapping mode and have the get function sort it out.

openscenegraph:
I'm back reviewing submissions again. 

The issue of Image::getColor() not knowing about different texture clamp modes and how they affect how texcoords should be mapped to pixels in an osg::Image is something that this submission doesn't resolve, it only makes a different assumption to the one that the original code makes. 

So merging this submission would fix one usage case only to break others, The best way to fix the Image::getColor(..) methods would be to remove the texcoord based ones completely and have a separate getColor() function that takes texture wrap modes into account.  It might be osg::getColor(texture, image, Vec3&) would be appropriate, or one that passes in the GL wrap modes.  A discussion about this on the osg-users mailing list/forum is probably the best place to thrash out what the best API would be,

PR #274, opened 2017-07-22
add IndirectDraw feature and MultiDrawIndirect PrimitiveSets

mp3butcher:
IndexedPrimitiveFunctors lacks a basevertex feature, it make MDElementsI::accept() methods unimplementable for IndirectedDrawPrimitives.
I added a basevertex parameter to IndexedPrimitiveFunctors::drawElements methods
virtual void drawElements(GLenum mode,GLsizei count,const GLubyte* indices, const GLuint basevertex = 0)
but it induces a cascade of file changes so I would be glad to have your opinion before composing a PR for this (other choices could be passing basevertex by ref or make basevertex a functor attribute...)

### DIFF
--- a/examples/osggpucull/DrawIndirectPrimitiveSet.cpp
+++ b/examples/osggpucull/DrawIndirectPrimitiveSet.cpp
@@ -21,21 +21,18 @@
 
 void DrawArraysIndirect::draw(osg::State& state, bool /*useVertexBufferObjects*/) const
 {
+    osg::GLBufferObject* dibo = _indirectCommandArray->getBufferObject()->getOrCreateGLBufferObject( state.getContextID() );
+    state.bindDrawIndirectBufferObject( dibo );
+
 // if you want to see how many primitives were rendered - uncomment code below, but
 // be warned : it is a serious performance killer ( because of GPU->CPU roundtrip )
+#if 0
+    osg::Drawable::Extensions *dext = osg::Drawable::getExtensions( state.getContextID(),true );
+    int* tab = (int*)dext->glMapBuffer(GL_DRAW_INDIRECT_BUFFER,GL_READ_ONLY);
+    int val = _indirect/sizeof(int);
+    OSG_WARN<<"DrawArraysIndirect ("<<val<<"): "<< tab[val] << " " << tab[val+1] << " " << tab[val+2] << " " << tab[val+3] << std::endl;
+    dext->glUnmapBuffer(GL_DRAW_INDIRECT_BUFFER);
+#endif
+    state.get<osg::GLExtensions>()->glDrawArraysIndirect( _mode, (const GLvoid *)(dibo->getOffset(_indirectCommandArray->getBufferIndex())+_indirect ));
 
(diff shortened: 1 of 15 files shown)


mp3butcher:
Thanks @jtorresfabra for pointing the squash feature
I lack of experience with advanced git usage...

openscenegraph:
I have just done two passes of the various changes and it's rather over-whelming, just too many changes packed into a single PR to make easy or practical to merge as-is. 

There a number of changes that I small improvements to code unrelated to DrawIndirect functionality, these should be separate commit's and PR;s.

There are oddities, like calling a simple packed data structure a command (GLDrawArraysIndirectCmd),  I don't understand the motivation for this, I don't we can be sure others will be confused too.  I also don't understand the types of C style declaration of this struct, the rest of the OSG doesn't use this C style declaration for structs so no need to start using it here.

The right way to get features like this introduced is to do it incrementally rather than push a whole lot of changes all at once, some parts are untroversial and can be merged right away other parts will need discussion.  This big PR's like this just make it impractical to tease apart all the different changes and different motivations behind the changes, it' also limits our ability in the future to trace changes and any possible regressions associated with them.

There are just too many issues wrapped up in this PR for me to merge so I'm going to close this PR.  I don't object to the addition of the functionality covered by this PR, just the manner of large PR and some of the implementation details and wish to avoid merging stuff that will need amending right away to fix oddities.

With big feature additions like this it may make sense to create a branch for their development and have PR's to build up to the final PR that merges that branch into master.  I'm open to creating a branch for this purpose.

mp3butcher:
-There are oddities, like calling a simple packed data structure a command (GLDrawArraysIndirectCmd),
command is the didacted term in the spec and I wrapped up as it is defined
    typedef  struct {
        uint  count;
        uint  instanceCount;
        uint  first;
        uint  baseInstance;
    } DrawArraysIndirectCommand;
-With big feature additions like this it may make sense to create a branch for their development
I think this PR is not so controversal and i can reproduce my dev steps dividing this pr in 3 
1)add MDI binding buffer directly
2 new files (MDI) + mod gpucull examples and  CMakeLists GLdefines and PrimitiveSet
2)add IndirectBufferObject and add it as State attribute (to benefit both from the osg bufferqueuemanagment and the minimization of glbindbuffer)
3x2 files changes State  BufferObject and MDI + mod on osggpucull
3)remove IndirectDrawBufferBinding in order users don't confused on which way to do indirect draw in osg IndirectPrimitiveSet would be the only way
2 files mod (BufferBindings)
Does it sound reasonnable?

openscenegraph:
Just looked the the ARB_draw_indirect specs and I see the derivation of the naming you adopted now.  It's a bit of an odd choice of naming from Khronous.  However, I'd rather stick with something a bit odd but at least consistent with the spec than come up with a more sensible name. So the please stick with the names you went from, just fix the C style declaration.  The export will only be needed if the struct/class isn't entirely implemented in the header.

Your suggests steps 1 through 3 look reasonable to me.  Tasks that would need to be done in addition to these (as a follow up submissions) would be IO support for the new classes, and what happens when doing things like geometry queries such as CPU based intersection tests.

PR #278, opened 2017-07-27
Mdi

mp3butcher:
1)add DrawIndirectBufferObject as State variable in order to minimize call to glBindBuffer(GL_DRAW_INDIRECT_BUFFER,X)
2)Promote PrimitiveSetIndirect's VBO to a target explicited DIBO (in order to benefit from BOSet queue management -in GLBufferObject::bindBuffer():if (_set) _set->moveToBack(this);-)

### DIFF
--- a/examples/osggpucull/osggpucull.cpp
+++ b/examples/osggpucull/osggpucull.cpp
@@ -189,7 +189,7 @@ struct IndirectTarget
     }
     void endRegister(unsigned int index, unsigned int rowsPerInstance, GLenum pixelFormat, GLenum type, GLint internalFormat, bool useMultiDrawArraysIndirect )
     {
-        osg::VertexBufferObject * indirectCommandbuffer=new osg::VertexBufferObject();
+        osg::DrawIndirectBufferObject * indirectCommandbuffer = new osg::DrawIndirectBufferObject();
         indirectCommandbuffer->setUsage(GL_DYNAMIC_DRAW);
         indirectCommands->setBufferObject(indirectCommandbuffer);
 
--- a/examples/osggpucull/osggpucull.cpp
+++ b/examples/osggpucull/osggpucull.cpp
@@ -181,18 +181,16 @@ struct IndirectTarget
         : maxTargetQuantity(0)
     {
         indirectCommands    = new osg::DefaultIndirectCommandDrawArrays;
+        indirectCommands->getBufferObject()->setUsage(GL_DYNAMIC_DRAW);
     }
     IndirectTarget( AggregateGeometryVisitor* agv, osg::Program* program )
         : geometryAggregator(agv), drawProgram(program), maxTargetQuantity(0)
     {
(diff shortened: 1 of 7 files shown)


mp3butcher:
@robertosfield :
Before to make further change I would like to know if you have any objection adding basevertex parameter (with default value 0) to Primitive*Functor::drawElements methods
it would involve changes in:
#       modifié :         include/osg/PrimitiveSet
#       modifié :         include/osg/TemplatePrimitiveFunctor
#       modifié :         include/osg/TemplatePrimitiveIndexFunctor
#       modifié :         include/osg/TriangleFunctor
#       modifié :         include/osg/TriangleIndexFunctor
#       modifié :         include/osg/TriangleLinePointIndexFunctor
#       modifié :         include/osgUtil/Statistics
#       modifié :         src/osg/Drawable.cpp
#       modifié :         src/osgPlugins/3ds/WriterNodeVisitor.cpp
#       modifié :         src/osgPlugins/dxf/DXFWriterNodeVisitor.cpp
#       modifié :         src/osgPlugins/obj/OBJWriterNodeVisitor.cpp
#       modifié :         src/osgUtil/Statistics.cpp

openscenegraph:
I don't think forcing baseVertex support out across parts of the OSG is a good design or implementation approach.  Basevertex is very much a niche bit of OpenGL functionality, and personally I consider it a bit of hacky part of OpenGL .  If you really want to mess around with baseVertex then then thing to do would be to offset the vertex array pointers before you call drawElements, this way you isolate the parts of the OSG that have baseVertex built in to handle the case rather than pushing it out across the whole OSG.

mp3butcher:
You're right it's better design (I haven't really think of implementing it)

openscenegraph:
I'm not going to merge the MultiDrawIndirect branch with master if it breaks the API and forces major changes out to the rest of the OSG and to client applications.  You might be feel that baseVertex functionality is worth it, but it really is a niche feature of OpenGL&OSG that will only be used by a very small number of users for very specific purposes, it's will never be a general purpose feature that most users will be. 

For me to merge code that will break the compile of end user applications it has to have real value for majority of end users and worth to them to go through the pain of migrating their code.  Often end users don't migrate wholesale, they often try to keep their applications working with older versions of the OSG as well as newer ones, the more inconsistencies the more different this is for them.

If not resolving this baseVertex works with primitive functors then so be it, again it's a niche feature, it's not going to effect the majority of the OSG users.

When modifying the OSG code base please keep in mind that knock on effect it'll have to the rest of the OSG, we need to minimize the breakages in API and functionality.

openscenegraph:
I have been doing a quick code review and have started tweaking several of the files to improve the readability and consistency with the rest of the OSG.  I have also spotted a few switch bugs where break; were missing so have added these.  Given these change it will be worth checking making sure you sync before making further modifications.

mp3butcher:
@robertosfield 
Thanks you for correcting my errors

PR #281, opened 2017-07-28
Adding basevertex to functor

mp3butcher:
sorry for useless commits
I forgot to merge and then cannot rebase(squash) these commits


openscenegraph:
I've done a quick review but it just underlines why I baseVertex functionality is such an ugly hack.  Instant red flags are raised about the whole block of functionality introducing a extra hacks across a wide range of code.

The code you have written forces an extra add to every single primitive index calculation in the the functors, this is an extra cost that *ALL* OSG users would have to pay for the convenience of tiny number of OSG users.  This is *BAD* solution.

I don't mind adding niche features to the OSG as long there isn't an intrusive cost to normal OSG usage, the baseVertex implementation as you are pushing doesn't fit with this, not even close, this is not a good direction to the take the OSG so I'm not willing to merge this approach.

mp3butcher:
@robertosfield 
I can duplicate drawelements methods and dispatch according if(_basevertex)...does it seams right to you?

openscenegraph:
Ahhh noooooooo, it's a just a horrible mess having to push baseVertex out, the feature is fundamentally a hacky workaround for specific usage cases in OpenGL, having this hacky mess spill out any further than it absolutely has to in the OSG is something we have to avoid.  The pain you are feeling trying to make it fit is down to the fundamental nature of this hacky OpenGL feature.  It's not a good fit, as we're finding out it's not orthogonal functionality that is easy to add without affecting other code.

I'm happy to not to expect this hacky part of OpenGL to have not have full support on the OSG side.

I think one way to progress would be to create a dedicated and simple example that illustrates how draw instanced and base vertex functionality would be expected to be used in the OSG, then this could be reviewed and discussed and then we can start thinking about this type of usage can be smoothly supported by the OSG.  I don't know yet how sensible it is to even expect any the functors to work with draw indirect.  

For example we have this issue with modern OpenGL usage - if you have a vertex or geometry shader that modifies vertex positions and meshing then the CPU based geometry queries have a real problem - you can't do standard CPU based intersection testing on them.  The OSG doesn't have a solution for this conundrum yet.  It's a problem we've had ever since ARB_vertex_program came about over a decade ago, we haven't yet got a solution for, but still for the most part OSG users have got by with this limitation.

In the case of baseVertex, it's far more niche than vertex shaders so potentially an area we can just mark as not fully supported for certain types of OSG usage and this won't serverly impact users.  If we can find a low cost, non intrusive way of generalizing the functionality then you'll get a thumbs up from me. 

I wouldn't hold up rolling out the functionality just because of the open issue of functors not supporting a niche OSG usage case.

--

As an aside, I'm doing code review and am wondering how you intended DrawIndirectBufferObject to be used, I haven't spotted any usage of it directly in the examples yet so I don't know yet how specifically you intend to use it.  It looks to be like DrawIndirectBufferObject and VertexBufferObject are similar in nature - both with similiar addArray/removeArrray etc. methods.  This makes me think it may be possible to subclass DrawIndirectBufferObject from VertexBufferObject.  This will keep the API consistent and avoid some code duplication.

mp3butcher:
I haven't any answer for the difference between OSG/CPU representation and GPU one. I have always avoided to dirty GPU generated vram bo and then ram BufferData (appart for reset stuff). From my point of view it's user charge to know OSG is not aware of GPU buffer changes. if user want to use OSG with them he will have to readback them 
(perhaps an flag in BufferObject could help him-like a cpudirty flags that trigged GLBufferObject to readback to its associated BufferData 
or better (even if it increases glBindBuffer calls) the cpudirty flags could be itself a BufferData whose BufferObject would be autobound by osg on each program in order to be write by GPU using SSBO -even if it would arbitrary enhance glsl spec and provoke a write conflict among GPUthreads- 
)

_I don't know yet how sensible it is to even expect any the functors to work with draw indirect._
as long as gpu and osg representation are the same it's safe. So it can be usefull to have for ex basevertex feature for BBcomputation at least for static geometries (which is my main focus I admit)


 _I'm doing code review and am wondering how you intended DrawIndirectBufferObject to be used,_
https://github.com/openscenegraph/OpenSceneGraph/commit/104b2dfc7217fe2d895d19af63cdddd339e5fb14
in this commit I promote the used VBO to DIBO in order to have correct _target when bindBuffer is called
https://github.com/openscenegraph/OpenSceneGraph/commit/a45c4c8f3986f0c7a2b0680f89d5f84bc548894b
In this commit I force commandbufferarray to have a dibo at construction so there's no need for user to set it himself

_This makes me think it may be possible to subclass DrawIndirectBufferObject from VertexBufferObject._ 
their _target differs so I decide polymorphism wasn't the good way to procede

openscenegraph:
W.r.t OSG/GPU representation issues for vertices and meshes, the solution has to be CPU based for performance and threading reasons - the round trip to the GPU is extremely expensive and has to be done from a thread with the appropriate graphics context current.  I think the solution in the end would be to have callbacks or perhaps even scripts using these callbacks to do the conversion from source vertex+primitive data to transformed vertex and primitive data.   This is a big topic though, not one to resolve here.   I gave this as example where we don't have solve all the OSG's usage cases in one go, so requiring other parts of the OSG support baseVertex is not essential right away.

For the bounding box issue you mention one can use the compute bound callback support into Drawable, same as we do right now when vertex shaders that modify vertex positions.

I am still not clear on the motivation for DrawIndirectBufferObject.  Your explanation just makes me think we can do away with it and have users setTarget on the VertexBufferObject as is.

mp3butcher:
_I am still not clear on the motivation for DrawIndirectBufferObject. Your explanation just makes me think we can do away with it and have users setTarget on the VertexBufferObject as is._
Indeed you can but setting target of a vbo can be considered as hacking;)

mp3butcher:
And concerning conversion from source vertex+primitive data to transformed vertex and primitive data, it may be impossible since most of the out stream spcification as the tendency to become more and more arbitrary

PR #294, opened 2017-08-02
add resize methods method in IndirectCommandArrays ..etc..

mp3butcher:
*add resize methods method in  IndirectCommandArrays (it allows user to work without casting to the concrete class result of MDI.get IndirectCommandArray())
*move getTotalDataSize in  IndirectCommandArrays interfaces
*comply with other DrawElementsXXX removing parameters in mdi constructors and add several method ( allow use of osgUtil::DrawElementTypeSimplifer on these)
*MDI support in osgUtil::DrawElementTypeSimplifer

### DIFF
--- a/include/osg/PrimitiveSetIndirect
+++ b/include/osg/PrimitiveSetIndirect
@@ -31,6 +31,8 @@ public:
     IndirectCommandDrawArrays(const IndirectCommandDrawArrays& copy,const CopyOp& copyop/*=CopyOp::SHALLOW_COPY*/) :
         BufferData(copy, copyop) {}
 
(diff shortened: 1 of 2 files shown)


openscenegraph:
The osg::DrawElements base class has a series of *Elements() methods that include reserveElements(..) and resizeElements(), the IndirextCommandArrays already adopts some of this language convention, so I'm inclined towards adopting this for all the array related sizing and access rather than follow the std::vector<> naming for resize().

Thoughts?

mp3butcher:
I'm ok with ..will do the mods..soon

mp3butcher:
done

PR #302, opened 2017-08-17
Support specifying GL version, profile and context flags on GLX

manisandro:
This PR contains initial work to support requesting a specific GL context, inspired by how Qt does it [1], in the attempt to get OSG running on Linux with intel/mesa drivers which does not provide a compatibility context.

I'm however stuck at the fundamental question how the glContextVersion should be specified in the traits. Should there be a default, handled directly in GraphicsWindowX11::init, perhaps controllable by an environment variable, which takes effect unless the user explicitly specified one in the traits?

Currently running the default osg example programs fails with shader compilation error messages like

    0:1(10): error: GLSL 3.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.00 ES, 3.00 ES, 3.10 ES, and 3.20 ES

because by default it does not create a core profile context with the required GL/GLSL version.

[1] https://github.com/qt/qtbase/blob/5.9/src/plugins/platforms/xcb/gl_integrations/xcb_glx/qglxintegration.cpp#L167

### DIFF
--- a/include/osgViewer/api/X11/GraphicsWindowX11
+++ b/include/osgViewer/api/X11/GraphicsWindowX11
@@ -37,6 +37,7 @@ class OSGVIEWER_EXPORT GraphicsWindowX11 : public osgViewer::GraphicsWindow, pub
             _parent(0),
             _window(0),
             _visualInfo(0),
+            _fbConfig(0),
             #ifdef OSG_USE_EGL
             _eglDisplay(0),
             _eglSurface(0),
@@ -176,6 +177,7 @@ class OSGVIEWER_EXPORT GraphicsWindowX11 : public osgViewer::GraphicsWindow, pub
         Window          _parent;
         Window          _window;
         XVisualInfo*    _visualInfo;
+        GLXFBConfig     _fbConfig;
 
         #ifdef OSG_USE_EGL
         EGLDisplay      _eglDisplay;
--- a/src/osgViewer/GraphicsWindowX11.cpp
+++ b/src/osgViewer/GraphicsWindowX11.cpp
@@ -42,6 +42,8 @@
 #include <X11/extensions/Xinerama.h>
 #endif
 
+#include <sstream>
+#include <cstdlib>
 #include <unistd.h>
 
 using namespace osgViewer;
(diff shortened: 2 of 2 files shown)


manisandro:
Actually there seems to be an existing logic to set the default context version via `DisplaySettings::setGLContextVersion` etc, but for that to work consistently it requires that the `DisplaySettings` instance is passed to the `Traits` constructor, which does not always appear to be the case. Should that be changed?

manisandro:
FWIW, with the change below, I'm finally able to get osgEarth running again with my mesa drivers (provided I set the context version and profile in the `DisplaySettings` before any context is created), but I don't know if this can break other stuff?

    diff --git a/src/osg/GraphicsContext.cpp b/src/osg/GraphicsContext.cpp
    index b8fce4436..e07e228cb 100644
    --- a/src/osg/GraphicsContext.cpp
    +++ b/src/osg/GraphicsContext.cpp
    @@ -246,6 +246,10 @@ GraphicsContext::Traits::Traits(DisplaySettings* ds):
                 overrideRedirect(false),
                 swapMethod( DisplaySettings::SWAP_DEFAULT )
     {
    +    if (!ds)
    +    {
    +       ds = DisplaySettings::instance().get();
    +    }
         if (ds)
         {
             alpha = ds->getMinimumNumAlphaBits();

mp3butcher:
I haven't review the pr and can't reproduce your issue because have no osx installed but, just to be sure check:
What are you cmake build configuration, check OSG_XXX_AVAILABLE flags? I believe to remember (but can't be sure... ) on macos you have to disable GL1/2 an GL_FIXED_FUNC in order to use gl>=3 context...

manisandro:
It's not a matter of the cmake configuration, the problem is that in the current master the X11 backend simply does not even try to request a GL3+ context because the code to do so is not there.

mp3butcher:
ok so you can try to use osgViewer command line option in order to set context properly
   osgViewer::Viewer viewer(arguments); with arg =  --gl-version 3.3

mp3butcher:
In fact you may be right what context version return glXCreateContext on osx just for curiosity?

openscenegraph:
I have done a quick review and it looks I strongly suspect it'll back the build for older systems so won't be safe to merge as is.  Next week I'll have a look at the PR as it is and have a look to what approach would be appropriate.  

This type of work would be discussed on the osg-users mailing list/forum as we need to get as many people involved in testing on as wide range of environments.  Creating a branch for it might be the best way of helping this too.  

Could you start up a topic on osg-users about the original issues and proposed changes.

mp3butcher:
perhaps i'm wrong but if some if(OSG_GL3_FEATURES) are replaced with #if OSG_GL3_FEATURES
It would preserve eldest systems to  GLX>1.3 code path

openscenegraph:
Given GLXFBConfig usage is an new addition I was a bit concerned about backwards compatibility. I've just had a look at GLXFBConfig and it's dates back to it's introduction GLX 1.3 alongside OepnGL 1.2, so... quite an old feature so should be safe to assume that all OSG GLX system that have it now :-)

I'm now doing a review, I may create a branch for this work, will keep you updated on my thoughts.

openscenegraph:
Hi Guys, 

Using this PR as a base I've done a bit of reworking to streamline the code and to keep the code formatting  consistent with the rest of the OSG, and added a XtFree to make sure the code doesn't leak.  Could you try out the commit that I've just merged with master:

https://github.com/openscenegraph/OpenSceneGraph/commit/8926f0e9c253a1781e6cdbde96ec467324bb105c

I haven't attempt to change the default setting of the GraphicsContext::Traits::glContextVersion do this still defaults to 1.0 for all platforms, how best to set this would be a discussion for osg-users ML/forum.  

I've tested the osgsimplegl3 example and confirm that the appropriate code paths are being selected.

manisandro:
Thanks for merging, I've started a discussion about the traits issue here:
http://lists.openscenegraph.org/pipermail/osg-users-openscenegraph.org/2017-August/006612.html

PR #306, opened 2017-08-19
Add support for type-1 (colour-mapped, uncompressed) targa images to …

AnyOldName3:
…the osgTGA plugin

An example of such an image is the `textures\_rb_fs1_wood1.tga` file in the Morrowind mod here:http://mw.modhistory.com/download-71-10678

This feature has already been merged into the OSG fork recommended for OpenMW, hence why I'm using a Morrowind mod as a test file.

### DIFF
--- a/src/osgPlugins/tga/ReaderWriterTGA.cpp
+++ b/src/osgPlugins/tga/ReaderWriterTGA.cpp
@@ -174,6 +174,22 @@ static int getInt16(unsigned char *ptr)
     return res | (tmp<<8);
 }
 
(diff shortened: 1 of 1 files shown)


openscenegraph:
@AnyOldName3 I'm currently updating the OpenSceneGraph/AUTHORS.txt list of contributors, could provide a human name so I can add this to the list.  Thanks.

AnyOldName3:
It's Chris Djali.

This is the end of an era for me - I've used this username everywhere since I was twelve, and up until now, it's been impossible to link it back to me.

PR #313, opened 2017-08-22
alleviate pressure on treading problems in std::ostream behind osg::Notify

LaurensVoerman:
While chasing a memory overrun causing spurious chashes on shutdown of our program, I found that this point in the code tended to trigger a treading issue in Visual Studio's (2013 & 2015) ostream buffer, as the size() function requests a mutex held by an other databasepager tread that's likely to genarate some osg_info as well.

### DIFF
--- a/src/osgDB/DatabasePager.cpp
+++ b/src/osgDB/DatabasePager.cpp
@@ -728,7 +728,10 @@ void DatabasePager::DatabaseThread::run()
 
         _active = true;
 
-        OSG_INFO<<_name<<": _pager->size()= "<<read_queue->size()<<" to delete = "<<read_queue->_childrenToDeleteList.size()<<std::endl;
+        if (osg::isNotifyEnabled(osg::INFO)) {
+            unsigned int rqSize = read_queue->size();
+            OSG_INFO << _name << ": _pager->size()= " << rqSize << " to delete = " << read_queue->_childrenToDeleteList.size() << std::endl;
+        }
 
 
 


openscenegraph:
I just did a review and something odd must being going on as your change just expands when the OSG_INFO macro already does, OSG_INFO maps to OSG_NOTIFY(osg::INFO) which expands to:

if (osg::isNotifyEnabled(osg::INFO)) osg::notify(osg::INFO)

Which is basically what you modifications do, which makes me wonder why the OSG_INFO macro is not guarding against the doing the output.  

I'd want to get to bottom of this as merging submissions like yours that basically create a workaround by adding lots of code to something that should already be happening.  If we start adding hacks like this where would we stop?

LaurensVoerman:
I understand the confusion, but it's about calling
read_queue->size()
BEFORE starting to write to the osg::Notify stream.
Holding the stream object while waiting for read_queue->size() to aquire
the mutex is causing a problem.
the "if (osg::isNotifyEnabled(osg::INFO))" only helps to not call size() at
all  if we are not writing output at all.

Regards, Laurens.

openscenegraph:
The stream code will only ever be called when the notification level is to INFO, which is not the default. with the default the level is NOTICE so stream code in question shouldn't be invoked.  One only enables INFO when in debugging phases, but this isn't something one would worry about performance.

I remain confused as to why the kludge is required.

LaurensVoerman:
It's because the thread stops to wait on the mutex in the size() function -
while holding the osg::NotifyStream object. The Notify stream inherits
from std::ostream - which is not thread safe - not by the C++ standard, and
not in the microsoft implementation
<https://msdn.microsoft.com/en-us/library/c9ceah3b.aspx> - at least not for
writing.
On the default notify level the original code did not cause any problems at
all, as no output gets written (not sure if the size() function gets called
at all).
Laurens.

openscenegraph:
So you are only seeing an issue when you set the notification level to INFO or DEBUG?

Reading the code it seems like your change won't functionality change whether the stream code will be executed or not, but when the read_queue->size(); doing before any stream operations rather in the middle of a set of stream operations.  I can see that this would change the timing of it.

However, this doesn't get away from the fact we are just tackling about debugging usage of the OSG, adding a cludge to the codebase to just workaround an implementation issues specific to one compiler or one platform is really not a good place to be.  This really isn't a good solution to a problem that is elsewhere.

LaurensVoerman:
That's correct - the issue only occurs at notification level to INFO or higher. My other submission hides this problem as well, except for the interleaved output (output of the tread holding the read_queue mutex is bound to end up in the middle of the line).
I would prefer to have both in the main osg tree, but if you prefer to drop this submission the other one seems to be sufficient to avoid triggering the problem often (not very well tested by me)
Regards, Laurens.

openscenegraph:
Thanks for the explanation.  I don't feel this PR is a good solution, the code add verbosity, makes the code less readable and outward functionality does nothing, it's just a workaround for an issue elsewhere, and for only when debugging.

PR #314, opened 2017-08-22
reduce the need to reallocate the std::ostream buffer behind osg::Notify

LaurensVoerman:
… (causing multitreading issues) by pre-allocating 4095 bytes.

While digging into treading problems I found out that Visual Studio's ostream has a problem if thread A starts some output to the stream, thread B writes enough output to cause the buffer to be resized and thread A then continues to write to the old buffer, while only respecting the length of the new buffer - overrunning the now released old buffer and overwriting other (valid) data. pre-allocating a reasonable buffer helps to minimize the problem - in my case it disappeared.
Regards, Laurens.

### DIFF
--- a/src/osg/Notify.cpp
+++ b/src/osg/Notify.cpp
@@ -61,6 +61,9 @@ struct NotifyStreamBuffer : public std::stringbuf
 {
     NotifyStreamBuffer() : _severity(osg::NOTICE)
     {
+        /* reduce the need to reallocate the std::ostream buffer behind osg::Notify (causing multitreading issues) by pre-allocating 4095 bytes */
+        str(std::string(4095, 0));
+        pubseekpos(0, std::ios_base::out);
     }
 
     void setNotifyHandler(osg::NotifyHandler *handler) { _handler = handler; }


openscenegraph:
Hi Laurens, What you describe looks like bug in VS's ostream implementation.  Are they fixing it?

The workaround you've posted is curious.  Is this a known workaround?

Could you add a comment to the additions to explain why they are needed and what they do, otherwise we have some obscure code in the OSG that outwardly doesn't do anything sensible.

LaurensVoerman:
I think the visual studio team thinks this is not a bug as the standard
does not require the std::ostream to be thread safe.
I'll try to add some code comments tomorrow.
Laurens.

openscenegraph:
I get general purpose streams not be thread safe but it makes sense for std::cerr and std::cout implementations to be thread safe...

LaurensVoerman:
the problem is not in std::cout, it's in std::stringbuf -
(NotifyStreamBuffer inherits from std::stringbuf )
this has come up before
<http://forum.openscenegraph.org/viewtopic.php?t=10533> - there's a little
test program in that post that should be able to show the problem exists on
linux too.
I should try that to see if I managed to hide the problem on linux too.

I also found a proposal for a full fix here
<https://www.mail-archive.com/osg-submissions@lists.openscenegraph.org/msg08698.html>
(your
reply here
<https://www.mail-archive.com/osg-submissions@lists.openscenegraph.org/msg08732.html>
)

Regards, Laurens.

LaurensVoerman:
I've added a comment in the code trying to explain why it's there.
In my first message I mentioned "ostream" - that's not correct. This is all about std::stringbuf.

I found this workaround online somewhere, but cannot find a link to that anymore.
If I remember this properly I changed it from using seekpos to pubseekpos to comply with the standard and to be able to compile with gcc.
Laurens.

openscenegraph:
I have tracked down the original test program that Mattias posted and got it building under Linux and confirmed that it reproduced the crash, albeit not on every run.  Applying Mattias' fix using multiple streambuffer, I'm not happy with Mattias' changes as they are rather intrusive and use of the optional compilation is not a good to maintain.  Perhaps implementing a thread safe stringbuf alternative would be better, though I haven't looked into what this might entail.   The notify implementation is already far more complicated than I ever intended so I'm keen to avoid added extra layers of complication.

I tested your proposed change associated with this PR and I am can't reproduce a crash so far using Mattias' test program, as workaround goes it's pretty minimal.  So thumbs up there.

Might it be good enough?

LaurensVoerman:
Thanks, I cannot get the test program to crash at all on linux (debian 7.7) 

As to a full tread safe version: I created one - by sticking a mutex in the osg::Notify singleton and changing the OSG_NOTIFY define like this:

#define OSG_NOTIFY(level) if (osg::isNotifyEnabled(level)) for (osg::notifyScopeMutexLock notifyLock; notifyLock.locked(); notifyLock.unlock()) osg::notify(level)

beware that this will create a deadlock if notify level is INFO or above at src/osgDB/DatabasePager.cpp line 731 - where the read_queue->size() function requires a mutex while holding the notify mutex.

I'am not to fond of this version as the notify locking in itself synchronizes threads making multi-threading issues even harder to spot, as they will disappear faster with increasing notify levels.
Laurens.

openscenegraph:
Thanks for adding the comment, will merge right away.

You could try a OpenThreads::ReentrantMutex to avoid the deadlock.

PR #316, opened 2017-08-23
fix Visual Studio compile warning C4800: 'unsigned __int64': forcing …

LaurensVoerman:
Hi Robert,
just fixing a few warnings.

### DIFF
--- a/src/osgPlugins/gles/GeometryIndexSplitter
+++ b/src/osgPlugins/gles/GeometryIndexSplitter
@@ -61,7 +61,7 @@ protected:
         }
 
         bool contains(unsigned int v1) const {
-            return subvertices.count(v1);
+            return 0 != subvertices.count(v1);
         }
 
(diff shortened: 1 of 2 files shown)


openscenegraph:
I've just done a quick review. I familiar with the pattern you've used to add the 0 != and the motivation behind it but personally find the code awkward to read, writing code that is less readable but done in way to try and use the compiler to spot mistakes is an odd choice.  Personally I find less readable code to be something that makes it more likely that users will make mistakes so I don't find it a good approach so I'm not happy seeing it spread out through the code base.

I have fixed the warnings in a more conventional and readable way: https://github.com/openscenegraph/OpenSceneGraph/commit/bb3b69962a6cb77b658c1b1bc26941a9fee82a42

Given I've applied fixes I'll now close to PR.

LaurensVoerman:
Thanks. I agree your version looks better.

Regards, Laurens.

PR #333, opened 2017-08-28
osgAnimation: change animationdata owner from riggeometry to rigtransformimplementations

mp3butcher:
Refactoring and add a prepareData in RigTransform interface because datapreparation is technic dependant (even if the 2 have few similitude)
The original design splitted data preparation in 2 steps:
- one when the skeleton is not findable (rig->buildinfluenceset)(rig not in a SG callable by use)
- one when the skeleton is set (rigtrans->init )(rig in a SG not callable by user but fails if no skeleton reachable ) -basically it replace bonename in result  buildinfluenceset with actual boneptr

In the pr i remove the first step and merge it in the second...
If it poses problem
I can rollback putting tempdata in VertexInfluenceMap or RigGeom
and revive RigTansform::init to do the 2nd step


mp3butcher:
@scrawl 
If you can test or review, you're welcome

jtorresfabra:
> Julian I have just done a quick review and it looks like you are changing a few names of methods etc and this raises a red flag of backwards compatibility. You might not rate it highly, but I'm sure lots of other users of osgAnimation will be troubled by the changes.

Totally agree, backward compatibility is a must for some of us. We are open to add / remove stuff version scoped wise, but eliminating backwards compatibility is really a blocker for us.

mp3butcher:
I totally agree not to break backward compatibility
I kept old method for that ( MorphGeometry::morphtransformsoftware and Rig::buildVertexInfluenceSet)
Which class I modify is a user class? I think none

Concerning the copyright stuff i remember to add my name but that's all (i'll check)

jtorresfabra:
@mp3butcher I can do some testing with older models once you feel the PR is good to merge. Right now it iseems more than it is a WIP.  Feel free to ping me.

mp3butcher:
@jtorresfabra if you have older models then you could validate that the behavior hasn't change

jtorresfabra:
@mp3butcher  Yes, that's why I'm waiting the PR to be ready :).

openscenegraph:
@mp3butcher Small changes like s/getName() -> s/getBoneName() are breaks in the API look like they risk breaking the build for no functional benefit.  One can argue that the new name is better, but the bar for justification is raised higher once an API is out in the wild and being used.  If it's niche part that is unlikely to be used by any developers then  the risk is low.

mp3butcher:
so you implies VertexInfluence (from wich getBoneName is) shouldn't be touched
ok for that  will do that in my next commit

openscenegraph:
I'm not the author or a user of osgAnimation so I'm working from the stance of providing general guidance, things like Name vs BoneName is a specific example, but it's a general issue about when it's OK or when it's not OK to change a name.  

In this specific instance it might be OK, or it might not, in general with any class or method name change there is a risk of breaking the build for end users.  If the change relates to internal implementation details then risk of breaking the build is low, if it's a widely used public API then you are certain to break the build.  

With a large number of changes being dropped together it's more difficult to provide guidance at a fine grained level, this is where having others in the community help out review, test and refine changes is useful.  

So please use me highlighting BoneName as an example of a potential issue rather than a declaration that this specific issue needs fixing, it might not if it's an internal implementation detail.  Every single code change needs to be considered as coming with a risk of breaking end user code, the nearer the changes are to the public API the higher the risk of breaking builds.

mp3butcher:
@robertosfield 
this pr follows my previous one. You created the branch without asking about the change s/VertexInfluence/BoneInfluenceSet/ that were breaking the api.
As a result it's hard to say in the diff what is dangerous as it's already based on something potentially dangerous (even if its a simple substitution)...
I rollbacked BoneInfluenceSet and its xetbonename to VertexInfluence and it xetName as it was, so there's no change spreading out of the spectrum of animation technics.
My guess is the easier would be to accept this pr and see in what the branch breaks the api with a diff/master...but you may disagree, it's not a really good practice...

openscenegraph:
@mp3butcher I proposed the branch to mitigate against the risks of changes and give you a bit of freedom to complete the work allowing others to chip in.  I didn't have a chance to review all the details before I pulled your original post into this branch.  There is limit on how much I can review large contributions in a short space of time - you know I have other tasks to juggle, I can't just down tools on everything else on do full reviews as soon as large PR is generated. I am only human.  Having PR's with many extensive changes is hard to deal with and get everything reviewed and all perfect feedback from day one.

mp3butcher:
Raa lovely!! after rewriting the whole code i found it!!!  the bug that drove me crazy since the begining in RigTransformHardware.

scrawl:
I agree with Robert on everything he said and couldn't have said it better myself. An example w.r.t. backwards compatibility that caught my eye was the VertexIndexWeight -> IndexWeight renaming, I pointed this out on the last PR so I guess you missed my comment? This change actually changed the example with it which is an indicator that it might break the build for users too.

That said I'm no longer a big user of osgAnimation, I only use MorphGeometry at the moment and might be moving to a custom version of that in the future, so you can take my comments more as general advice than anything. I don't have a lot of time to test things now and it takes a lot of time for me build OSG after switching a branch but I can give it another review and/or test when you feel like the branch is working well and finalised.

openscenegraph:
@mp3butcher What is the status of this PR now.  It's evolved a bit, should I merge now?

mp3butcher:
I revert IndexWeight to VertexIndexWeigh and It will be good

mp3butcher:
@scrawl ok it's revert  but I would tend to say that if you build yourself you influencemap then you can deal with these minor api changes...

mp3butcher:
@scrawl @robertosfield  @jtorresfabra 
If theres's no more objections I think this pr can be merge on the branch (it's just a branch)
I will open a subject for critics on the forum

jtorresfabra:
@mp3butcher  I don't have any objection on merging on the branch. We can continue the discussion there. 
Good work btw! Let me know whenever is good to start to test.

mp3butcher:
@jtorresfabra it's good enough to do some testing

scrawl:
> @scrawl ok it's revert but I would tend to say that if you build yourself your influencemap then you can surely deal with these minor api changes ...

True, not everyone will build their own InfluenceMap but I actually did do so as part of importing another model format into osgAnimation. I imagine there may be some other users (like procedural animations).

The deal with backwards compatibility isn't about being able to deal with things, of course people are able to, the real issue is this hassle is multiplied by the number of users affected, they may need to introduce version guards if they still intend to compile with older versions of OSG, older versions of the product that don't include this check will no longer compile with new OSG versions, not being able to use git bisect as easily, etc, these are part of the reason why backwards compatibility is important and I very much appreciate that Robert takes it seriously in his project.

In general and in my projects I tend to be skeptical towards any changes with little value added, be it a renaming, refactoring, commenting or whitespace/indentation changes, in a big project you have to consider every change has a hidden cost that may not be obvious at first, like potential API breaks, potential merge conflicts, clottering up review and git history, breaking something on accident or just leading to pointless discussions (not saying this one is).

PR #337, opened 2017-08-30
fix a bug in how vertexattributes are filled

mp3butcher:
That was it all along: 
a bug leading to bad  vertexattribute info in the shader


openscenegraph:
Quick clarification, is this fix meant for master or the osganimation branch?

mp3butcher:
it's a hotfix for master

openscenegraph:
How much more work do you think would be required before osganimation branch would be considered for merging?

mp3butcher:
I don't know it would require testing from the community.
The RigTransform new design differs a bit from the initial so i hope none will have problem with it

mp3butcher:
@AurL @jtorresfabra @cedricpinson 
This bug concern also your osganimation version

PR #340, opened 2017-09-01
add a new method on VertexInfluenceMap and the serializer for MorphTransforfHW

mp3butcher:
1)add a method to remove useless bones on a skeleton
2)update XXXTranformHW serializers with new properties
3) make preparedata skeleton independant again (as it was with the original Rig::buildInfluenceSet() path)

I also merge with master to correct the auto mergeability status of the branch


mp3butcher:
@jtorresfabra @scrawl 
Give your feedback here
http://forum.openscenegraph.org/viewtopic.php?p=71570#71570

mp3butcher:
...and I think that's all..I think I will not make further modification

mp3butcher:
...and I think that's all..I might not make further modifications on this branch

openscenegraph:
@mp3butcher. Just done a quick scan of the changes.  

With additions to serializes we need add version checks to the addition/removals to ensure backwards compatibility of .osgt,.osgb and osgx files.  Have a look at the UPDATE_TO_VERSION_SCOPED(..) in RigGeometry.cpp serializers for an example of this.  Could you add this to your additions to your changes to the serializers. Thanks.

mp3butcher:
done

mp3butcher:
@robertosfield 
I was wondering if using codeinjection would it be better than current marmelade with current shaders prioritization (morph then skinning). It would allow more flexibility....
But codeinjection is tagged as old (deprecated?) in osgshadercomposition example in benefit of osgshaderpipeline. 
What's your point of view about that? Is integrate all osg possible effects in "meta über shaders"  (shaders/osgshaderpipeline.*) is the chosen course? it would fastly become huge meta shaders!? Am I right?

openscenegraph:
My plan isn't for a single large osg::Shader, rather having separate osg::Shader that are linked together when required.  #pargma(tic) shader composition allows you to optionally link in shaders so you don't pay the penalty for having huge glProgram's, even if at the C++ programming level you link in lots of osg::Shader to your osg::Program. 

For code injection, you can do it with #pragma(tic) shader composition by using the #define's.

PR #343, opened 2017-09-04
Dynamic changing of 'osgText without any decoration' fix

valid-ptr:
I've found, that osgtext3d example doesn't work
with "osgText::Text::TEXT" Text3D's draw mode.

This PR will fix this.
I think, _texcoords arrays should also be marked as dirty in TextBase.

KOS

### DIFF
--- a/examples/osgtext3D/osgtext3D.cpp
+++ b/examples/osgtext3D/osgtext3D.cpp
@@ -49,23 +49,34 @@ public:
         {
             if (ea.getKey() == osgGA::GUIEventAdapter::KEY_Up)
             {
-                m_Text3D->setCharacterSize(m_Text3D->getCharacterHeight() + 0.1);        // failed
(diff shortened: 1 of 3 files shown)


openscenegraph:
@valid-ptr Just done a quick review.  Could you split the submission into changes to the osgtext3D.cpp and changes to osgText as the former changes aren't specifically related to the bug you are addressing.  One a quick review I was also concerned about whether _texcoords is always set so suitable to be modified, it would be worth double checking to see if this assumption is correct.

valid-ptr:
I've created PR #344 for Text3D only.

I can't 100% understand what you are talking about _texcoords, so I did not touch it .
Also, I want to notice about _colorCoords (BIND_PER_VERTEX) in TextBase. What about it?

KOS

openscenegraph:
I will have a look at the texCoord issue once I get a chance (I'm currently working osgText but related to functionality).  I have merged the other changes.

PR #349, opened 2017-09-20
Update TriStripVisitor.cpp

vkovalev123:
Deleted unnecessary sorting which results in long downloads of large files


openscenegraph:
I have just done a first pass review and the sort line that your removed looks like a key part of the remove duplicates operation so is not appropriate to remove, doing so would completely break the removal of duplicates.  Given this PR introduces a bug to code that looks correct I am closing this PR.

Could you please provide a dataset that illustrates the performance issue you've seen to the osg-users mailing list/forum so that others can test it and look at how to resolve the issue properly.

vkovalev123:
On lines 224-227 there is a loop that fills the array in an ordered form without duplicates. Any file which contains many points illustraites this problem(over 1 million)

PR #350, opened 2017-09-21
allow for multiple --login arguments

LaurensVoerman:
current code only adds the details of the first --login on the command line


openscenegraph:
I have reviewed your changes and rather than merge these I have add a Registry::getOrCreateAuthenticalMap() and now use this in osgviewer.cpp to add the multiple logins support.  This approach simplifies the code and makes it more readable. Commit is: ddf8c68d267669290a8dc08a7f7d2ea2f1ece1b4

PR #351, opened 2017-09-21
VNC: try to find password for host if no password for host:port is found

LaurensVoerman:
when using a different port like localhost:5901.vnc the password needed to be set with "--login localhost:5901 unusedUsername password".
with this change the password for localhost is used if no password for localhost:5901 is found.
Regards, Laurens.


openscenegraph:
Hi Laurens, I have merged your changes but had to add some spacing and put the { on a separate line to make the code more readable and consistent with the style used elsewhere in the OSG.  The commit of the style/readability fixes is:

https://github.com/openscenegraph/OpenSceneGraph/commit/042a357278411b7c5c2ee91fd0f3089080499d3d

PR #357, opened 2017-10-04
fix normalization bug (happen too early )

mp3butcher:
not sure if this bug  affects master too ...
if a bone is not found in the bone map it's removed but vertexggroup is not normalized after that


openscenegraph:
I have merged the changes to the osganimation branch, but the new code wasn't consistent with the rest of the OSG or easy to read so I've made some further changes that can be found in the commit : eb5f46d..7868b42

Issue #360, opened 2017-10-10
Pragma(tic) shader composition generates invalid shader text with CRLF line endings


siu:
We are refactoring some code to load shaders using the new pragma shader composition. Some more context to understand the problem:

 * We load the shader text from a text file
 * This comes from a git repository and git (by default) converts the line endings to the platform line endings. On windows CRLF and on unixes LF.
 * This works fine if we don't use the new setDefine(..., ...)
 * I am testing on nvidia hardware + drivers

As soon as we use defines the shader code is split in several blocks, one with the glsl version, one with the defines and a third one with the rest of the shader source. If the original shader source uses CRLF line endings the produced "#version 150" line becomes "#version 150\r" which does not compile at run-time. Everything works as expected if the line endings are "\r\n".

This is most likely a problem around https://github.com/openscenegraph/OpenSceneGraph/blob/cb2b0a5fe84bf862613b46432226a0201005fced/src/osg/Shader.cpp#L611

I can contribute a fix but I would like to know what is the preferred approach:

1. Convert "\r\n" line endings to "\n" in osg::Shader::setSource(...)
2. Handle "\r\n" in osg::Shader::PerContext::compileShader(...) 
3. "\r\n" are not supported and the documentation of osg::Shader needs to be updated. The client code must provide the correct line endings.

Also we need the fix in the stable version, do you accept patches to 3.4 or only to master?

openscenegraph:
What a pain these type of issues are, platform incompatibilities that stem from stupid decisions by MS and alike many years ago but still plaguing us today... life would be so much simpler if all platforms just adopted the LF standard of Unix.

The Shader.cpp file has diverged a bit between 3.4 and master so we might need to different patches.  My inclination, rather than change user source via a setSource() change, is to fix it in compileShader or in osg::State though I'd have to look at all relevant code to make a decision though.  I am busy with other tasks right now so can't investigate right away.  If you have a fix then generate a PR and I can can review the changes, this will give me a feel form what is likely to be required if not the final solution.

siu:
Thanks for the input, I will see if I can provide a patch.

PR #361, opened 2017-10-11
Fix pragma shader composition when shader source contains windows lin…

siu:
…e endings

There was a bug separating the "#version X" line when the shader source code
contained windows line endings. The produced shader code contained single '\r'
characters which failed to compile in the OpenGL driver (at least on nvidia).
The fix unifies line endings to single '\n' character at the beginning of the
parsing.

The target branch is 3.4, a similar fix can be applied to master.

Addresses issue #360.

### DIFF
--- a/src/osg/Shader.cpp
+++ b/src/osg/Shader.cpp
@@ -19,6 +19,7 @@
  *         Holger Helmich 2010-10-21
 */
 
+#include <algorithm>
 #include <fstream>
 #include <list>
 #include <sstream>
@@ -688,6 +689,8 @@ void Shader::PerContextShader::compileShader(osg::State& state)
     }
     else
     {
+        // Remove \r from windows line endings
+        source.erase(std::remove_if(source.begin(), source.end(), EqualTo('\r')), source.end());
 
         std::string versionLine;
         unsigned int lineNum = 0;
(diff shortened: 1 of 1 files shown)


openscenegraph:
I have just had a look at the changes, it looks like you are removing /r but this makes the assumption that /r are always coupled with /n, but on OSX it's possible to have files with /r only, so it would potentially break OSX.

siu:
Ouch, I didn't know that, I will give it a second spin.

siu:
I pushed a second commit with a different solution. It doesn't show in github though :octocat:

openscenegraph:
Thanks change now looks safe.  I'll merge this, but I'd like to change the code to an in place replace of '\r\n" with " \n" as this will be more efficient that creating and deleting copies on the heap.

siu:
Sure, thanks for merging.

openscenegraph:
I have implemented the in place replacement, the commit is 6496c304f99e162d4f45ddc886b96f82eb7d30b2, it's also now merged with both the OpenScenGraph-3.4 branch and master.  I've tested it on my Linux with a file converted to Widows and Mac endings and it works as far as I can tell, could you test it on your Windows system? Thanks.

siu:
I tested 3.4 on Windows, it works well now.

PR #367, opened 2017-10-13
copySubImage support for block compressed images, added support for astc compressed type.

LaurensVoerman:
-added support for variable block compression sizes
-added types introduced with GL_KHR_texture_compression_astc_hdr extention
-added copySubImage support to copy compressed blocks (aligned with block footprint only)

-check added to copy to give a warning instead of a crash if new failed
-guard added to Image::getColor to return (1,1,1,1) for unsupported compressed textures (=not dxtc)

TODO/MISSING
-Image::flipHorizontal no guard/support for any compression
-Image::flipVertical() very crude OSG_NOTICE for compressed textures, only POT dxt textures will actually be flipped

### DIFF
--- a/src/osg/Image.cpp
+++ b/src/osg/Image.cpp
@@ -241,11 +241,17 @@ Image::Image(const Image& image,const CopyOp& copyop):
     {
         unsigned int size = image.getTotalSizeInBytesIncludingMipmaps();
         setData(new unsigned char [size],USE_NEW_DELETE);
-        unsigned char* dest_ptr = _data;
-        for(DataIterator itr(&image); itr.valid(); ++itr)
+        if (unsigned char* dest_ptr = _data) 
         {
-            memcpy(dest_ptr, itr.data(), itr.size());
-            dest_ptr += itr.size();
+            for(DataIterator itr(&image); itr.valid(); ++itr)
+            {
+                memcpy(dest_ptr, itr.data(), itr.size());
+                dest_ptr += itr.size();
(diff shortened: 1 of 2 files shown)


src/osg/Image.cpp
```diff
@@ -241,11 +241,17 @@ Image::Image(const Image& image,const CopyOp& copyop):
     {
         unsigned int size = image.getTotalSizeInBytesIncludingMipmaps();
         setData(new unsigned char [size],USE_NEW_DELETE);
-        unsigned char* dest_ptr = _data;
-        for(DataIterator itr(&image); itr.valid(); ++itr)
+        if (unsigned char* dest_ptr = _data)
```
openscenegraph: How far back in C++ does putting variable declarations in if statements go?  It seems like a bit of risk for breaking the build on older systems.

scrawl: There's a new feature in C++17 that allows you to put multiple declarations in one if statement, but the simple `if (type var = ....)` has always been here as far as I know. I never had an issue with that.

And by the way, this syntax is already used in OpenSceneGraph [here](https://github.com/openscenegraph/OpenSceneGraph/blob/3b1212fed59b1024aab289bf7837056c559913e1/examples/osgmultiplemovies/osgmultiplemovies.cpp#L599).

openscenegraph: Thanks for spotting the existing usage, tells us it should be safe to merge :-)

PR #368, opened 2017-10-18
backport of object cache fixes

rhabacker:


openscenegraph:
Ralf, my aim with stable releases is to where possible to maintain the ABI, this PR changes to the public API of the classes so will change the ABI.

PR #369, opened 2017-10-18
Add a flag to manage the inertia on camera from user code.

touchline3d:
A method is added to enable or disable the inertia applied on the camera when it is moved with the mouse.

Indeed, if the mouse is moved while a mouse button is pressed, the camera translates, rotates or zoom. If the mouse button is released while the mouse is still moved, the camera continues its previous transformation in a linear way (~inertia).

As this behavior may not be wanted in some applications, the inertia can be deactivated from user code thanks to this new method.

The default value of the inertia matches with the previous implementation of concrete camera manipulators for compatibility matter.


openscenegraph:
This PR is not required as the intended functionality is already built into osgGA.  Simply use StandardManipulator::setAllowThrow(false).

PR #371, opened 2017-10-22
Removal of compiler warning for Cocoa builds

hwiesmann:
if protocol is not defined a compiler warning is raised with respect to line 1726:

`        [NSApp setDelegate: [[CocoaAppDelegate alloc] init] ];
`


hwiesmann:
Replace  commas with semicolons to prevent compiler warnings. Actually, in this situation the commas are not harmful, the warning is just annoying.

hwiesmann:
In commit b1655dd17b585820b8ce6f11c991674349b14c7c a required #endif directive has been accidentally removed.

PR #379, opened 2017-11-04
    following guidance from OpenGL Common Mistakes:

mp3butcher:
According https://www.khronos.org/opengl/wiki/Common_Mistakes#Creating_a_complete_texture final sentence:
    mipmaps should be upload with glTexImage2D and not glTexSubImage2D
perhaps _borderWidth should be replace by 0...don't know

### DIFF
--- a/src/osg/Texture.cpp
+++ b/src/osg/Texture.cpp
@@ -2298,9 +2298,8 @@ void Texture::applyTexImage2D_load(State& state, GLenum target, const Image* ima
                         if (height == 0)
                             height = 1;
 
-                        glTexSubImage2D( target, k,
-                            0, 0,
-                            width, height,
+                        glTexImage2D( target, k, _internalFormat,                          
+                            width, height, _borderWidth,
                             (GLenum)image->getPixelFormat(),
                             (GLenum)image->getDataType(),
                             dataPtr + image->getMipmapOffset(k));
@@ -2603,12 +2602,11 @@ void Texture::applyTexImage2D_subload(State& state, GLenum target, const Image* 
                     if (height == 0)
                         height = 1;
 
-                    glTexSubImage2D( target, k,
-                        0, 0,
-                        width, height,
-                        (GLenum)image->getPixelFormat(),
-                        (GLenum)image->getDataType(),
-                        dataPtr + image->getMipmapOffset(k));
+                    glTexImage2D( target, k, _internalFormat,                          
+                            width, height, _borderWidth,
+                            (GLenum)image->getPixelFormat(),
+                            (GLenum)image->getDataType(),
+                            dataPtr + image->getMipmapOffset(k));
 
                     width >>= 1;
                     height >>= 1;


robertosfield:
It looks like the modified code will only work if you use osg::Image with an PixelBufferObject as you drop the dataPtr+ term when passing the pointer to the image data. When using PBO's the code set the dataPtr to 0, when not using PBO's the dataPtr is set to the base of the osg::Image data.  The mipmap vector in osg::Image should only contains offsets, not absolutely pointers.

mp3butcher:
What?!!
 dataPtr + image->getMipmapOffset(k))
and
             image->getMipmapData(k))
are equivalents....
What's controversal in this submission?! 
Since all this time this only working path was auto mipmap (even tex2D).... It's a bit of a shame....
Please consider more carefully this pr

robertosfield:
The code isn't equivalent.  The code exists in Texture,cpp specifically to handle the possible use of PBO's, you can't just use the data pointers directly in this case.  As things stand this PR introduces a bug so absolutely not acceptable.  The original dataPtr+image->getMipmapOffset(k) is correct and should be re-instated in your code.

The glTexSubImage vs glTextImage usage is a curious one, it's something that was a standard way of doing mipmaps back in the late nineties/early 2000's when the code was written. The glTexSubImage route assumes that the texture object has been created with all it's mipmpaps and you are just filling them in, or that the driver is able to allocate automatically.  With the advent of PBO's and glTextureStorage the Texture code has ended up rather more convoluted than it was in the early days, there is chance there has been a regression, or perhaps its just the modern drivers aren't as lenient as they once were,

mp3butcher:
`  uint width=4;

    osg::TextureCubeMap *cubemap=new osg::TextureCubeMap();
    typedef std::vector<osg::ref_ptr< osg::Image> > VecMipMap;
    VecMipMap mips[6];
    uint maxLevel = log(width) / log(2);
    uint offset = 0,size;
    osg::Image::MipmapDataType mipoffsets;
    for( uint l=0; l<=maxLevel; ++l) {
        if( l!=0)mipoffsets.push_back(offset);
        size= 1<< maxLevel - l;
        offset+=size*size*4;
    }
    osg::Image * cubeface[6];


    for( uint face=0; face<6; ++face) {
        cubeface[face]=new osg::Image;
        unsigned char * data1=new unsigned char[offset];
        cubeface[face]->setImage(width,width,1,GL_RGBA,GL_RGBA,GL_UNSIGNED_BYTE,data1,osg::Image::USE_NEW_DELETE);
        cubeface[face]->setMipmapLevels(mipoffsets);
        for( uint l=0; l<=maxLevel; ++l) {

            size= 1<< maxLevel - l;
            osg::Image * im=new osg::Image();
            unsigned char * data=new unsigned char[size*size*4];
            memset(data,0, size*size*4);

            for(uint i=0; i<size*size; i++) {
                data[i*4+l%3]=255;
                data[i*4+3]=255;
            }
            im->setImage(size,size,1,GL_RGBA,GL_RGBA,GL_UNSIGNED_BYTE,data,osg::Image::USE_NEW_DELETE);
            mips[face].push_back(im);
        }
    }

    for( uint face=0; face<6; ++face) {
        cubemap->setImage( face, cubeface[face] );
        uint l=0;
        for(std::vector<osg::ref_ptr< osg::Image> >::iterator mit=mips[face].begin(); mit!=mips[face].end(); ++mit,++l)
        {
            //  osgDB::writeImageFile(*(*mit),"cubeface.png");
            memcpy(cubeface[face]->getMipmapData(l),(*mit)->data(), (*mit)->getTotalDataSize());

        }
    }

///according gl spec all mipmap mest be discribed

    osg::Geometry * geom=osg::createTexturedQuadGeometry(osg::Vec3(-1.0, -1.0,0.0),osg::Vec3(2.0, 0.0,0.0),osg::Vec3(0.0, 2.0,0.0));

    cubemap->setWrap(osg::Texture2D::WRAP_S, osg::Texture2D::REPEAT);
    cubemap->setWrap(osg::Texture2D::WRAP_T, osg::Texture2D::REPEAT);
    cubemap->setWrap(osg::Texture2D::WRAP_R, osg::Texture2D::REPEAT);
    cubemap->setFilter(osg::Texture::MIN_FILTER, osg::Texture::LINEAR_MIPMAP_LINEAR);
    cubemap->setFilter(osg::Texture::MAG_FILTER, osg::Texture::LINEAR);
    geom->getOrCreateStateSet()->setTextureAttributeAndModes(0,cubemap);

    geom->getOrCreateStateSet()->setMode(GL_LIGHTING,osg::StateAttribute::OFF);
    osg::ref_ptr<osg::Group> rot=new osg::Group;
    char lodovertime[]="uniform samplerCube tes;\
    uniform float osg_SimulationTime;\
    void main(){\
    gl_FragColor=textureLod(tes,gl_FragCoord,mod(osg_SimulationTime*0.25,2));\
    }";
    osg::Program*prog=new osg::Program;
    prog->addShader(new osg::Shader(osg::Shader::FRAGMENT,lodovertime));
    geom->getOrCreateStateSet()->setAttributeAndModes(prog);
    rot->addChild(geom);
    osgDB::writeNodeFile(*rot.get(),"test.osgb");
    viewer.setSceneData(geom);
    viewer.run();`

mp3butcher:
here's a cubemap sample that doesn't work...(I extract it from other stuff so it lacks includes)

robertosfield:
Thanks for the example and returning the getMipmpaOffset() line.  In principle this PR now looks ready to merge, I want to test and see the problem first hand before I merge though - I don't have any doubt about the problem, but I want to know how the problem appears so that if others in the past or in the future come across similar issues we can more quickly spot the problem and provide advice about the fix. Thanks.

mp3butcher:
thanks for ur time

openscenegraph:
I have now created an test program around your code snippet.  Without the changes I see a black quad, with the PR changes I see a colour quad that slowly changes colour.  So I'm guessing this means that we have a fix so I'll merge.

On glitch that we still need to resolve is that the test program emits a GL error during the compile stage:

Warning: detected OpenGL error 'invalid enumerant' at StateSet::compileGLObejcts() compiling texture attribute TextureCubeMap

I get this GL error before and after applying the PR.  I haven't yet worked out the cause of this.

openscenegraph:
I have been looking into the GL error and found that it's caused by the glTextureStorage2D call prior to the glTexImage2D calls, the glTextureStorage2D is being called for each face of the cubemap so enum's GL_TEXTURE_CUBE_MAP_POSITIVE_X. GL_TEXTURE_CUBE_MAP_POSITIVE_Y etc.

I suspect that glTextureStorage2D is being misused for this context, this may have a bearing on why the glTexSubImage2D calls weren't working.  I will now look into the API docs of glTextureStorage2D to double check what should be done here.

openscenegraph:
I have now done more testing and research and found out that this PR is actually the wrong fix, the PR actually introduces a new bug rather than fix the original bug.  

The original bug causing the mipmapped cubemap error was calling glTextureStorage2D for each face with the face target enum rather than once with GL_TEXTURE_CUBE_MAP.  When using glTextureStorage2D is entirely correct to use glTexSubImage2D and *wrong* to the glTexImage2D.  By implementing the fix to the glTextureStoage2D I've been able to get rid of the GL error and get the correct result with the test case.

The OpenGL docs of glTextureStorage + cubmap can be found on the kronous page:

https://www.khronos.org/opengl/wiki/Common_Mistakes

"When using glTexStorage2D instead of glTexImage2D, you should call glTexStorage2D once with the target GL_TEXTURE_CUBE_MAP, then make calls to glTexSubImage2D to upload data for each face."

Grrr.... should have stuck with my original instincts, a fully investigate the issue rather than prematurely merge a buggy PR.  I will now revert this commit and work on getting a proper fix checked in.

openscenegraph:
I have now reverted this PR and checked in the correct fix with commit : 

https://github.com/openscenegraph/OpenSceneGraph/commit/dfec052eb983c10467faeae16c9cb7dca5b45379

This fix is now checked into master and the OpenSceneGraph-3.4 branch.

mp3butcher:
Sorry for this premature pr...I could have investigate a bit further before pr....

PR #382, opened 2017-11-10
Osganimation

mp3butcher:
Seams all is good...
Ready for merging


mp3butcher:
thanks for the review

openscenegraph:
Hi Julian,  I have done a code review of the changes in the PR, I spotted a few code style issues that impaired readability of the code so I have made amendments for this, and spotted on instance of a if () else {} block where then second block code was outside the }, so would have resulted in a bug.  I've merged these.  I have also changed the version check for the serailizer additions so they new are only done for SO version 152.  I've bumped the SO version to 152 in master to mesh with this.

With the changes now merged with master we still need a bit of vigilance for any issues that might arise in the community as the code base will now get wider exposure.  Thanks for you efforts on osgAnimation.

openscenegraph:
Hi Julian,

Once I merged this osganimation PR I generated a Covertiy Scan and it found 8 new defects.  Could you have a look at these?  I will have a look at them myself but I feel it would be good for you to see what issues Coverity picks up in your code changes so you can learn about the subtle errors that can creep into code.

Robert.

mp3butcher:
Hi I can remove leak covertiy warning in HWmorphTrans (replacing ptr wit ref_ptr should do the tricks) but I would like your opinion on coverity warnings in osgAnimation serialisers methods wrapping: 
coverity tells to control the result of the dynamic_cast<reinterpret_cast<>(void*)> but AFAIK when we know for sure the type of the object so controlling the dyn_cast result would be redondant (am i right).

openscenegraph:
Hi Julien,

I have merged fixes for these warnings.  Using ref_ptr<> and add if () was all that was required.

Robert.

PR #388, opened 2017-11-27
VAS "propertization": a possible road to VAS-less drawable

mp3butcher:
1st and 2nd commit/
It's a beginning of answer for the desactivation of Drawable VAS
This design actually already helps me a lot to implement tricks like "pingpong TransformFeedback buffer set" or "baseVertex primsets"

3rd commit/
Add ComputeDispatch and modify example
(i added a vas property test in draw Drawable::draw...perhaps there's better:
I might interact with state in PCProgram::useprogram in order to disable vao but i don't know where to reenable it ...perhaps at the end of CD::drawimpl ...
)

@robertosfield 
Would like your opinion on it

### DIFF
--- a/examples/osgcomputeshaders/osgcomputeshaders.cpp
+++ b/examples/osgcomputeshaders/osgcomputeshaders.cpp
@@ -63,7 +63,7 @@ int main( int argc, char** argv )
     // Create a node for outputting to the texture.
     // It is OK to have just an empty node here, but seems inbuilt uniforms like osg_FrameTime won't work then.
     // TODO: maybe we can have a custom drawable which also will implement glMemoryBarrier?
-    osg::ref_ptr<osg::Node> sourceNode = osgDB::readRefNodeFile("axes.osgt");
+    osg::ref_ptr<osg::Node> sourceNode = 0;//osgDB::readRefNodeFile("axes.osgt");
     if ( !sourceNode ) sourceNode = new osg::ComputeDispatch( 512/16, 512/16, 1 );
     sourceNode->setDataVariance( osg::Object::DYNAMIC );
     sourceNode->getOrCreateStateSet()->setAttributeAndModes( computeProg.get() );
--- a/examples/osgcomputeshaders/osgcomputeshaders.cpp
+++ b/examples/osgcomputeshaders/osgcomputeshaders.cpp
@@ -20,7 +20,7 @@
 // This example can work only if GL version is 4.3 or greater
 
 #include <osg/Texture2D>
-#include <osg/Geometry>
+#include <osg/ComputeDispatch>
 #include <osg/Geode>
 #include <osgDB/ReadFile>
 #include <osgGA/StateSetManipulator>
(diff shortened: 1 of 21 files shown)


openscenegraph:
I have just done a first pass review and a bit stunned, so many code changes, the code readability in so many places being compromised.  21 files changed to add a narrow set of new functionality. It's hard to love these type of changes.

If you wanted to introduce a container for a buffer of the VertexArrayState object then the appropriate thing to do would be to create a new class VertexArrayStateBuffer/VertexArrayStateSet or similar name.

Because there are so many code changes simply because you chose a different naming scheme it's much harder for me to review and see the actual functional changes rather than cosmetic.  

Some of the commits mixes stuff up, between the different stages of work so while I'd like to just pull out the ComputeDispatch class and merge right away it's not straight-forward.  Could you generated a PR for just the ComputeDispatch header and source file/CmakeList changes.  Don't worry about the VAS redundancy I think I can probably fix this relatively easily by overrriding the Drawable::createVertexArrayState() and returning NULL and then making sure the rest of the Drawable code doesn't implode on this change.

Being able to reuse VertexArrayState is whole different topic, this might be done via a VertexArrayStateBuffer or having a callback that can override the createVertexArrayState() method so users can implement their own custom behaviour - including sharing.

As a general guide, when making changes to the OSG it's best to think about how to minimize the number of name, line and file changes required, this not only avoids breaking 3rd party code, it also leads to much easier to follow code as it evolves over time.  

Occasionally it necessary to do sweeping changes but these have to be very carefully considered, the wider the changes the higher the hurdle of justification that you must apply.

mp3butcher:
I admit I mixed 2 prs here:
- VertexArrayStateBuffer sharing stuff (I'd rather use VertexArrayStateBuffer/VAS instead of VAS/PCVAS naming to minimize renaming)
-ComputeDispatch related stuff (your way through createVAS is much more elegant..i admit i was tired when began to hack Drawable::draw() knowing it was bad)
You don't seam against VertexArrayStateBuffer property and sharing so there's no point trying to testify it with a not-so-fitted-feature (ComputeDispatch)

openscenegraph:
Will you be able to create a PR for ComputeDispatch today?  I'd like to tag 3.5.9 today.

W.r.t having a VertexArrayStateBuffer to replace the std::buffer_object< osg::ref_ptr<osg::VertexArrayState>, it's something I considering in my original implementation but decided to keep things simple, and to avoid an extra indirection associated with getting to the VertexArrayState object.  Adding a VertexArrayStateBuffer will add an extra step so will be slower w.r.t CPU overhead, something I worked really hard to try a minimize when I was implementing the VAS functionality.

So... while I'm not completely against the approach of having a VertexArrayStateBuffer object if we can provide flexibility required another route that doesn't add an overhead for normal OSG usage then I'm keen to explore such possibilities first.

mp3butcher:
I've made the pr for ComputeDispatch ...

Concerning 'VertexArrayStateSet' property, 
I've renamed VAS/PCVAS to respect same convention as Program/PCProgram it seams clearer to me:
Readability after renaming was the purpose of the 1st commit
If you prefer VASSet/VAS i will to that but not convinced...

As you mentions, It can't be done without an extraindirection,... the classic flexibility cost...I don't know if it would involve huge penalty..

PR #389, opened 2017-11-28
add ComputeDispatch class

mp3butcher:
revoke glDispatch in PCProgram::useProgram
update example

remove numGroups both in Program and its serializer (@robertosfield  retrocompat not maintained)

### DIFF
--- a/examples/osgSSBO/osgSSBO.cpp
+++ b/examples/osgSSBO/osgSSBO.cpp
@@ -10,7 +10,7 @@
 #include <osg/StateAttributeCallback>
 #include <osg/Texture2D>
 #include <osg/Geometry>
-#include <osg/Geode>
+#include <osg/ComputeDispatch>
 #include <osgDB/ReadFile>
 #include <osgGA/StateSetManipulator>
 #include <osgViewer/Viewer>
@@ -167,6 +167,7 @@ class ComputeNode : public osg::PositionAttitudeTransform
 
 public:
 
+    osg::ref_ptr<osg::ComputeDispatch>                _computeDispatch;
     osg::ref_ptr<osg::Program>                        _computeProgram;
     osg::ref_ptr<osg::Shader>                        _computeShader;        //compute and write position data in SSBO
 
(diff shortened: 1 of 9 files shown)


mp3butcher:
perhaps createVAS should be overiden in ComputeDispatch....

openscenegraph:
I have done a first pass run through, mostly the changes look good. 

Breaking the backwards compatibility of serailizer isn't acceptable though so we'll need to come up with a fix. The way I would tackle this is to read dummy values and discard them, possible emitting a warning.  You can remove elements from serializer after specified versions. I'm happy to tackle this work.

openscenegraph:
Hi Julian,

The ComputeDispatch PR is now merged, thanks for you efforts on this.  I've added back on serilalizer support for compute groups parameters to the Program serializer to retain backwards compatibility.  It would be worth you looking how I did it so you know how it's done in future.  The commit is:

https://github.com/openscenegraph/OpenSceneGraph/commit/c0a276e85023927cd056a28a028523ff709b8492#diff-b253cee269678a0c9d82052e07d0272c

openscenegraph:
Dang, haven't done a before and after test w.r.t the SSBO warning yet...

openscenegraph:
I just tested the OpenSceneGraph-3.4 branch and it has the same error.  I had to fix the shader file search paths in this branch to get as far a running with all the shaders so there is a fix checked into this branch for this.

With these fixes to finding the correct shaders I see exactly the same errors that your reported for the osgSSO example, it looks to me like the shaders are wrong.  Now looking into it.

openscenegraph:
The GLSL errors in the osgSSO example are down to some vec3's being mixed in with float maths, it's buggy for sure.  I'm not the original author of this example or shader so I don't know what the intention is in the shaders.  Looking at the logs it was written by Marcus Hein back in 2014.

openscenegraph:
I have hacked the SSO shaders a bit to avoid the mixing of types, I don't know the original intent behind the maths so likely took things away form the original intention but at least it compiles and runs cleanly.  An update to OpenScenGraph-Data is required to get this fix.

PR #392, opened 2017-11-28
Vas propertization

mp3butcher:
more readable pr....

### DIFF
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -195,7 +195,11 @@ class OSG_EXPORT Drawable : public Node
         /** Get the const Shape of the const Drawable.*/
         inline const Shape* getShape() const { return _shape.get(); }
 
+        /** set the vertex array state of this Drawable **/
+        void setVertexArrayStateSet(VertexArrayStateSet* vasset) { _vertexArrayStateSet = vasset; }
 
+        /** get the vertex array state of this Drawable **/
+        VertexArrayStateSet* getVertexArrayStateSet()const { return _vertexArrayStateSet; }
 
         /** Set the drawable so that it can or cannot be used in conjunction with OpenGL
           * display lists.  When set to true, calls to Drawable::setUseDisplayList,
@@ -491,8 +495,7 @@ class OSG_EXPORT Drawable : public Node
         typedef osg::buffered_value<GLuint> GLObjectList;
         mutable GLObjectList    _globjList;
 
-        typedef buffered_object< osg::ref_ptr<VertexArrayState> > VertexArrayStateList;
-        mutable VertexArrayStateList    _vertexArrayStateList;
+        osg::ref_ptr<VertexArrayStateSet>  _vertexArrayStateSet;
 
         ref_ptr<DrawCallback>   _drawCallback;
 };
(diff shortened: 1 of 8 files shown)


openscenegraph:
This PR is a lot clearer. Thanks.

I'm not convinced on the merit of making the VertexArrayStateSet a property as it introduces another object on the heap and another operation that can cause a cache miss.  I tried really hard to avoid these extra overheads when I implement the vertex array object support into the OSG, the implementations that we have wasn't implemented casually, I thought about many different approaches including the one you just have implemented, I chose not to follow it when I implemented VAO because of the CPU overhead.  This overhead hasn't changed.

FYI, one of the metrics I was using in the VAO implementation was trying to close the performance gap between using display lists and non display lists, using moderately large city model for performance benchmarking.  As much as tried I couldn't close the gap, for this class on model display lists are better, but VAO halves the overhead relative to just using VBO's on this type of dataset - large number of relatively small geometries.  The OSG is used heavily in the vis-sim market where such models are standard fodder, it's not an area we can drop the ball on.

For reusing VertexArrayState objects I think one way forward is to have a CreateVertexArrayStateCallback that can assigned to osg::Drawable, and takes over the osg::Drawable::createVertexArrayState() virtual function.  Drawable would then have a createVertexArrayStateImplementaion() that does the actual work with createVertexArrayState() now doing a :

    if (_createVertexArrayStateCallback.valid())  return (*_createVertexArrayStateCallback)(renderInfo);
    else creteVertexArrayStateImplementation()

I believe such a callback could have it's own local cache of VAS's and share these out to the osg::Drawable that share this callback.  This approach would enable the overhead associated with wiring up the shader of VAS to just the create operation, thereafter there would be no overhead relative to the current implementation.

I'm happy to go ahead an implement this as it's pretty lightweight.  No point in me implementing if it doesn't solve the problem you are trying to solve though.  I think it does but I'm not coding your application.

openscenegraph:
I have decided to go ahead an implement the CreateVertexArrayStateCallback approach I mentioned above.  This approach is really lightwegiht and keeps the changes just to the Drawable base class and should enable custom management of the VAS objects such as sharing between Drawables.  I am still testing this functionality but should be able to check it in soon.

mp3butcher:
Great
I'm agree with you a calback design could be better in term of performance.
I haven't thought deeply in the implication in would have in my already coded tricks, but at first glance i think i could adapt.
Perhaps i will have problem with my PingPongTFDrawCallback because i need to swap VASSet of 2 pingpong drawables...

openscenegraph:
I have now checked in my additions, the key parts are Drawable::CreateVertexArrayStateCallback and addition of public accesses to the VertexArrayStateList.   The combination of these two should make it possible to customize how the VertexArrayState objects are managed.

https://github.com/openscenegraph/OpenSceneGraph/commit/a6069c3226a42d6ef29348b9e3ddd8adb5e147fd

mp3butcher:
My first tought on this feature was to propose flexibitility of geoms VAS without having to subclass Geometry, if this flexibility is too costy perhaps it's just invalid.
I doubt this feature can be as usefull as i tought initially ...
- In the form of a prop it allows vas swapping and sharing without having to subclass Geometry
- In the form of a callback it allows only VAS sharing (which i admit has low spectrum of application -except exotics "basevertex primsets" ,i can't see other usecase-)
Perhaps it would be wiser to let user subclass himself than providing a feature he will have difficulty to exploit....

PR #397, opened 2017-12-01
Added misc fixes from cppcheck

jtorresfabra:
Hi @robertosfield, 

Quick first pass on cppcheck linter warnings and errors. Mainly unused variables and some small leaks. 

Thanks

### DIFF
--- a/applications/osgconv/osgconv.cpp
+++ b/applications/osgconv/osgconv.cpp
@@ -740,7 +740,7 @@ int main( int argc, char **argv )
     while(arguments.read("--addMissingColours") || arguments.read("--addMissingColors")) { addMissingColours = true; }
 
     bool do_overallNormal = false;
-    while(arguments.read("--overallNormal") || arguments.read("--overallNormal")) { do_overallNormal = true; }
+    while(arguments.read("--overallNormal")) { do_overallNormal = true; }
 
     bool enableObjectCache = false;
     while(arguments.read("--enable-object-cache")) { enableObjectCache = true; }
--- a/applications/present3D/present3D.cpp
+++ b/applications/present3D/present3D.cpp
@@ -805,8 +805,6 @@ int main( int argc, char **argv )
         viewer.realize();
 
         if (doSetViewer) setViewer(viewer, width, height, distance);
-
-        viewerInitialized = true;
     }
 
 
--- a/src/osg/ApplicationUsage.cpp
+++ b/src/osg/ApplicationUsage.cpp
@@ -286,7 +286,6 @@ void ApplicationUsage::write(std::ostream& output, unsigned int type, unsigned i
         if (needspace) output << std::endl;
         output << "Keyboard and Mouse Bindings:"<<std::endl;
         write(output,getKeyboardMouseBindings(),widthOfOutput);
-        needspace = true;
     }
 
 }
--- a/src/osg/Geometry.cpp
+++ b/src/osg/Geometry.cpp
@@ -570,15 +570,10 @@ void Geometry::setUseVertexBufferObjects(bool flag)
     DrawElementsList drawElementsList;
     getDrawElementsList(drawElementsList);
 
-    typedef std::vector<osg::VertexBufferObject*>  VertexBufferObjectList;
-    typedef std::vector<osg::ElementBufferObject*>  ElementBufferObjectList;
-
     /*if (_useVertexBufferObjects)*/
     {
         if (!arrayList.empty())
         {
-
-            VertexBufferObjectList vboList;
 
             osg::ref_ptr<osg::VertexBufferObject> vbo;
 
@@ -604,8 +599,6 @@ void Geometry::setUseVertexBufferObjects(bool flag)
 
         if (!drawElementsList.empty())
         {
-            ElementBufferObjectList eboList;
-
             osg::ref_ptr<osg::ElementBufferObject> ebo;
 
             DrawElementsList::iterator deitr;
--- a/src/osg/State.cpp
+++ b/src/osg/State.cpp
@@ -1257,7 +1257,6 @@ namespace State_Utils
                 std::string::size_type start_var = pos;
                 ++pos;
                 pos = str.find_first_not_of("ABCDEFGHIJKLMNOPQRTSUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_", pos);
-                std::string var_str;
                 if (pos != std::string::npos)
                 {
 
--- a/src/osgAnimation/StatsHandler.cpp
+++ b/src/osgAnimation/StatsHandler.cpp
@@ -409,7 +409,6 @@ struct ValueTextDrawCallback : public virtual osg::Drawable::DrawCallback
             if (!visitor)
                 return;
 
-            std::string font("fonts/arial.ttf");
             float leftPos = 10.0f;
             float characterSize = 20.0f;
 
@@ -427,7 +426,6 @@ struct ValueTextDrawCallback : public virtual osg::Drawable::DrawCallback
              }
 
             const std::vector<std::string>& channels = visitor->getChannels();
-            std::map<std::string,int> size;
             for (int i = 0; i < (int)channels.size(); i++) {
                 std::string name = channels[i];
                 if (_actions.find(name) == _actions.end()) {
@@ -440,14 +438,11 @@ struct ValueTextDrawCallback : public virtual osg::Drawable::DrawCallback
                     //_actions[name].touch();
                 }
                 _actions[name]._group->setNodeMask(~osg::Node::NodeMask(0x0));
-                size[name] = 0;
                 pos.y() -= characterSize + graphSpacing;
             }
 
             pos.y() -= backgroundMargin;
             osg::Vec3Array* array = static_cast<osg::Vec3Array*>(_background->getVertexArray());
-            float y = (*array)[0][1];
-            y = y - (pos.y() + backgroundMargin); //(2 * backgroundMargin + (size.size() * (characterSize + graphSpacing)));
             (*array)[1][1] = pos.y();
             (*array)[2][1] = pos.y();
             array->dirty();
--- a/src/osgAnimation/StatsVisitor.cpp
+++ b/src/osgAnimation/StatsVisitor.cpp
@@ -77,10 +77,6 @@ void StatsActionVisitor::apply(ActionStripAnimation& action)
     if (isActive(action))
     {
         _channels.push_back(action.getName());
-        double value;
-        std::string name = action.getName();
-        if (_stats->getAttribute(_frame, name, value))
-            name += "+";
         _stats->setAttribute(_frame, action.getName(), action.getAnimation()->getAnimation()->getWeight());
     }
 }
--- a/src/osgDB/OutputStream.cpp
+++ b/src/osgDB/OutputStream.cpp
@@ -860,7 +860,6 @@ void OutputStream::writeSchema( std::ostream& fout )
         ObjectWrapper::TypeList types;
         wrapper->writeSchema( properties, types );
 
-        std::string propertiesString;
         unsigned int size = osg::minimum( properties.size(), types.size() );
         for ( unsigned int i=0; i<size; ++i )
         {
--- a/src/osgGA/TerrainManipulator.cpp
+++ b/src/osgGA/TerrainManipulator.cpp
@@ -126,7 +126,7 @@ void TerrainManipulator::setByMatrix(const Matrixd& matrix)
 
             _rotation.set(0,0,0,1);
 
-            hitFound = true;
+            //hitFound = true;
         }
     }
 
--- a/src/osgGA/UFOManipulator.cpp
+++ b/src/osgGA/UFOManipulator.cpp
@@ -536,11 +536,6 @@ void UFOManipulator::_adjustPosition()
         return;
 
     // Forward line segment at 3 times our intersect distance
-
-
-    typedef std::vector<osg::Vec3d> Intersections;
-    Intersections intersections;
-
     // Check intersects infront.
     osg::Vec3d ip;
     if (intersect(_position,
--- a/src/osgPlugins/bsp/Q3BSPReader.cpp
+++ b/src/osgPlugins/bsp/Q3BSPReader.cpp
@@ -36,8 +36,6 @@ Q3BSPReader::Q3BSPReader()
 bool Q3BSPReader::readFile(const std::string& file,
                            const osgDB::ReaderWriter::Options* options)
 {
-    std::string ext = osgDB::getLowerCaseFileExtension(file);
-
     Q3BSPLoad load_data;
     load_data.Load(file,8);
 
--- a/src/osgPlugins/bsp/VBSPReader.cpp
+++ b/src/osgPlugins/bsp/VBSPReader.cpp
@@ -653,8 +653,6 @@ ref_ptr<StateSet> VBSPReader::readMaterialFile(std::string materialName)
     bool                    found = false;
     ref_ptr<StateSet>       stateSet;
     std::string             shaderName;
-    std::string             texName;
-    std::string             tex2Name;
     ref_ptr<Texture>        texture;
     ref_ptr<Texture>        texture2;
     ref_ptr<TexEnvCombine>  combiner0;
@@ -992,7 +990,6 @@ void VBSPReader::createScene()
     Quat                        yaw, pitch, roll;
     ref_ptr<MatrixTransform>    propXform;
     std::string                 propModel;
-    std::string                 propFile;
     ref_ptr<Node>               propNode;
 
     // Load the materials and create a StateSet for each one
--- a/src/osgPlugins/dae/daeRMaterials.cpp
+++ b/src/osgPlugins/dae/daeRMaterials.cpp
@@ -589,8 +589,6 @@ bool daeReader::processColorOrTextureType(const osg::StateSet* ss,
     }
     bool retVal = false;
 
-    std::string texCoordSet;
-
     //osg::StateAttribute *sa = NULL;
     //TODO: Make all channels process <param ref=""> type of value
     if (channel == osg::Material::EMISSION )
--- a/src/osgPlugins/dae/daeRSceneObjects.cpp
+++ b/src/osgPlugins/dae/daeRSceneObjects.cpp
@@ -361,7 +361,7 @@ osg::Node* daeReader::processLight( domLight *dlight )
     lightmodel->setAmbientIntensity(osg::Vec4(0.0f,0.0f,0.0f,1.0f));
     _rootStateSet->setAttributeAndModes(lightmodel, osg::StateAttribute::ON);
 
-    osg::LightSource* lightsource = new osg::LightSource();
+    osg::ref_ptr<osg::LightSource> lightsource = new osg::LightSource();
     lightsource->setLight(light);
     std::string name = dlight->getId() ? dlight->getId() : "";
     if (dlight->getName())
@@ -520,7 +520,7 @@ osg::Node* daeReader::processLight( domLight *dlight )
         light->setDirection(osg::Vec3(0, 0, -1));
     }
 
-    return lightsource;
+    return lightsource.release();
 }
 
 // <camera>
--- a/src/osgPlugins/jpeg/ReaderWriterJPEG.cpp
+++ b/src/osgPlugins/jpeg/ReaderWriterJPEG.cpp
@@ -689,7 +689,7 @@ class ReaderWriterJPEG : public osgDB::ReaderWriter
               default:
               {
                   OSG_DEBUG << "ReaderWriterJPEG::write_JPEG_file - Error pixel format non supported" << std::endl;
-                return WriteResult::ERROR_IN_WRITING_FILE; break;
+                  return WriteResult::ERROR_IN_WRITING_FILE;
               }
             }
 
--- a/src/osgPlugins/obj/obj.cpp
+++ b/src/osgPlugins/obj/obj.cpp
@@ -231,7 +231,6 @@ bool Model::readMTL(std::istream& fin)
     bool usingDissolve = false;
 
     Material* material = 0;// &(materialMap[""]);
-    std::string filename;
 
     while (fin)
     {
--- a/src/osgPlugins/ogr/ReaderWriterOGR.cpp
+++ b/src/osgPlugins/ogr/ReaderWriterOGR.cpp
@@ -262,8 +262,6 @@ public:
 
     osg::Geometry* multiPointToDrawable(OGRMultiPoint* mpoint) const
     {
-        osg::Geometry* geom = new osg::Geometry;
-
         osg::Geometry* pointGeom = new osg::Geometry();
         osg::Vec3Array* vertices = new osg::Vec3Array();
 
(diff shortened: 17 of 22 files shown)


openscenegraph:
I have just had a quick look at the chances, quite a few look like they could break things, I really need to know what the warning that was being reported for each change to know whether the change is appropriate or not.  

Could you let me know what run cppcheck in the manner that you did so I recreate the warnings+errors.

jtorresfabra:
Ummm what of these changes could break existing code?

I can do separate PR's for each warning if you want. Would that be ok?

openscenegraph:
I can't say exactly which of PR are definitely buggy, but quite a few looked dubious, and would require proper consideration about the complete flow of the application to know whether it was safe.  These changes have all been made without any context so there is no way for me to judge what motivated the change.

 With these "fixes" to warnings it's really easy to fix the "warning" but break the actual code.  I know this from first hand experience - I've made fixes to address these type of issues only to find out later I've broken the code.  Very often this "warnings" from code analysis tools are false positives or pick out completely benign behaviour.  

Generating a separate for each change doesn't help me in this instance.  The only way for me to make a decision on each one is based on what the warning/error report was, then to step through the code and see what the effect would be for each change,

This is why I really need to see the cppcheck output myself.

jtorresfabra:
I understand your point. Still not sure what changes makes you to hesitate, if you can point them to me I can try to explain the change. But is totally understandable that you want to see the output yourself, it will make things clearer. 

So I use a qtcreator as IDE and a plugin for cppcheck. Let me know how can I help.

applications/osgconv/osgconv.cpp
```diff
@@ -740,7 +740,7 @@ int main( int argc, char **argv )
     while(arguments.read("--addMissingColours") || arguments.read("--addMissingColors")) { addMissingColours = true; }
 
     bool do_overallNormal = false;
-    while(arguments.read("--overallNormal") || arguments.read("--overallNormal")) { do_overallNormal = true; }
+    while(arguments.read("--overallNormal")) { do_overallNormal = true; }
```
jtorresfabra: same values in || so one of them not needed

src/osgPlugins/dae/daeRSceneObjects.cpp
```diff
@@ -520,7 +520,7 @@ osg::Node* daeReader::processLight( domLight *dlight )
         light->setDirection(osg::Vec3(0, 0, -1));
     }
 
-    return lightsource;
+    return lightsource.get();
```
rafagaitan: This should be a .release() AFAIK

src/osgPlugins/ogr/ReaderWriterOGR.cpp
```diff
@@ -262,7 +262,7 @@ class ReaderWriterOGR : public osgDB::ReaderWriter
 
     osg::Geometry* multiPointToDrawable(OGRMultiPoint* mpoint) const
     {
-        osg::Geometry* geom = new osg::Geometry;
+        osg::ref_ptr<osg::Geometry> geom = new osg::Geometry;
```
rafagaitan: is geom properly used?, seems to me it's only initialized and then used below only for reading. I think it's not required

src/osgPlugins/dae/daeRSceneObjects.cpp
jtorresfabra: yep agree, fixed

src/osgPlugins/ogr/ReaderWriterOGR.cpp
jtorresfabra: Yes also agree, fixed. Thanks @rafagaitan

openscenegraph:
Communication for this type of work is often far more work than just doing the work, I have plenty of other pressing tasks to get on with so I'm not at all keen to go through some fine grained discussions on dissecting why my gut instinct tells me that this needs further investigation.   I really need to make efficient use of my time.

Seeing the output myself is the most efficient way for me to review stuff.  Is there an easy command line invocation I can use?  I don't have time to go learning a new tool right now.

jtorresfabra:
Hi @robertosfield:

> Seeing the output myself is the most efficient way for me to review stuff. Is there an easy command line invocation I can use? I don't have time to go learning a new tool right now.

$sudo apt-get install cppcheck
$cd path/to/openscenegraph
$cppcheck . --enable=performance --quiet 2> errors.txt

Check errors.txt once it finishes.

openscenegraph:
Thanks for suggestion on cppcheck. I won't have the time today to look at cppceck but when my current workload drops back a bit I'll run it and see is finds.

filnet:
On PR per change is a bit of an overkill. 
But you could do a PR per warning type,, starting with the simple / safe ones (like unused variables for example).

jtorresfabra:
> But you could do a PR per warning type,, starting with the simple / safe ones (like unused variables for example).

Well I meant that. Sorry if I was not clear.

applications/osgconv/osgconv.cpp
openscenegraph: I've merged this one mannually

applications/present3D/present3D.cpp
```diff
@@ -805,8 +805,6 @@ int main( int argc, char **argv )
         viewer.realize();
 
         if (doSetViewer) setViewer(viewer, width, height, distance);
-
-        viewerInitialized = true;
     }
```
openscenegraph: Removing this line is safe with the current rev of the code but it's bad practice given how similar code elsewhere in the main relies upon the viewInitilized = true being executed, without it it would introduce a bug.  The danger in removing this currently redundant operation if  order of code changes slightly or this block is copied an pasted earlier suddenly we'd introduce a bug. 

Finally if (!initialized) { do initialization; initialized = true; } is a common technique, the pattern is balanced, it helps tell the programmer what the code does and when it's complete the job of initialization is done.  Without this any programmer reviewing the code will be left wondering is there anything else to be initialized.

All in all cppcheck in this instance is technically correct from a pure C++ -> machine code basis, but wrong form the perspective that C++ code is a live and breathing record of what the programmer intends for the computer to do, and a part of the dialogue with future programmers reviewing the code.

src/osg/ApplicationUsage.cpp
```diff
@@ -286,7 +286,6 @@ void ApplicationUsage::write(std::ostream& output, unsigned int type, unsigned i
         if (needspace) output << std::endl;
         output << "Keyboard and Mouse Bindings:"<<std::endl;
         write(output,getKeyboardMouseBindings(),widthOfOutput);
-        needspace = true;
```
openscenegraph: Same issue, cpphcheck might be technically correct the needspace is redundant but absolutely the wrong thing to do from a living and breathing body of C++ code that is open to revisions over time - if one adds another one of these blocks but not add the needSpace you'd get a bug.  Far better to leave the redundant assignment.  The consequence performance wise is utterly irrelevant as it's not an inner loop, and only ever called once in very special circumstances.

Here again wasting our time of silly false positives.

src/osg/Geometry.cpp
```diff
@@ -604,8 +599,6 @@ void Geometry::setUseVertexBufferObjects(bool flag)
 
         if (!drawElementsList.empty())
         {
-            ElementBufferObjectList eboList;
-
             osg::ref_ptr<osg::ElementBufferObject> ebo;
```
openscenegraph: These changes look safe, now merged and submitted to master.

src/osg/State.cpp
```diff
@@ -1257,7 +1257,6 @@ namespace State_Utils
                 std::string::size_type start_var = pos;
                 ++pos;
                 pos = str.find_first_not_of("ABCDEFGHIJKLMNOPQRTSUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_", pos);
-                std::string var_str;
```
openscenegraph: Yay!  A bona fida bug, testing for this fix led me to find other bugs elsewhere in this function so I've ended up fixing two bugs.

src/osgAnimation/StatsHandler.cpp
```diff
@@ -409,7 +409,6 @@ struct ValueTextDrawCallback : public virtual osg::Drawable::DrawCallback
             if (!visitor)
                 return;
 
-            std::string font("fonts/arial.ttf");
```
openscenegraph: Removal looks appropriate, fix merged and submitted to master

src/osgAnimation/StatsHandler.cpp
```diff
@@ -440,14 +438,11 @@ struct ValueTextDrawCallback : public virtual osg::Drawable::DrawCallback
...
                 }
                 _actions[name]._group->setNodeMask(~osg::Node::NodeMask(0x0));
-                size[name] = 0;
                 pos.y() -= characterSize + graphSpacing;
             }
 
             pos.y() -= backgroundMargin;
             osg::Vec3Array* array = static_cast<osg::Vec3Array*>(_background->getVertexArray());
-            float y = (*array)[0][1];
-            y = y - (pos.y() + backgroundMargin); //(2 * backgroundMargin + (size.size() * (characterSize + graphSpacing)));
```
openscenegraph: This and the four other lines above look like code that had been experimented with and then left unused.  Assigning the same pos.y value later in the code to multiple vertices looks a bit suspicious so I'm not confident that there isn't a wider bug here and it may be appropriate to re-instate these unused variables at somepoint.  Given this it's not appropriate to remove the code so I have commented it out and checked this into master.

src/osgAnimation/StatsVisitor.cpp
```diff
@@ -77,10 +77,6 @@ void StatsActionVisitor::apply(ActionStripAnimation& action)
     if (isActive(action))
     {
         _channels.push_back(action.getName());
-        double value;
```
openscenegraph: Removal looks approrpiate, code looks to have been experimental but not actually used for anything in the end.

src/osgDB/OutputStream.cpp
```diff
@@ -860,7 +860,6 @@ void OutputStream::writeSchema( std::ostream& fout )
         ObjectWrapper::TypeList types;
         wrapper->writeSchema( properties, types );
 
-        std::string propertiesString;
```
openscenegraph: Remove looks appropriate, checked in.

src/osgAnimation/StatsVisitor.cpp
```diff
@@ -77,10 +77,6 @@ void StatsActionVisitor::apply(ActionStripAnimation& action)
...
-        std::string name = action.getName();
```
openscenegraph: Remove looks appropriate, checked in.

src/osgGA/TerrainManipulator.cpp
```diff
@@ -126,7 +126,7 @@ void TerrainManipulator::setByMatrix(const Matrixd& matrix)
 
             _rotation.set(0,0,0,1);
 
-            hitFound = true;
```
openscenegraph: Another example if (!state) { state = true; ) pattern is used here and not safe to change in case later changes assume that the pattern is maintained with all blocks.  It's not safe to remove, or even comment out.  It's not a performance issue leaving it in.

src/osgGA/UFOManipulator.cpp
```diff
@@ -536,11 +536,6 @@ void UFOManipulator::_adjustPosition()
         return;
 
     // Forward line segment at 3 times our intersect distance
-
-
-    typedef std::vector<osg::Vec3d> Intersections;
-    Intersections intersections;
-
```
openscenegraph: Removal looks appropriate now checked in.

src/osgPlugins/bsp/Q3BSPReader.cpp
```diff
@@ -36,8 +36,6 @@ Q3BSPReader::Q3BSPReader()
 bool Q3BSPReader::readFile(const std::string& file,
                            const osgDB::ReaderWriter::Options* options)
 {
-    std::string ext = osgDB::getLowerCaseFileExtension(file);
```
openscenegraph: Remove looks appropriate, checked in.

src/osgPlugins/bsp/VBSPReader.cpp
```diff
@@ -992,7 +990,6 @@ void VBSPReader::createScene()
     Quat                        yaw, pitch, roll;
     ref_ptr<MatrixTransform>    propXform;
     std::string                 propModel;
-    std::string                 propFile;
```
openscenegraph: These three removal all looks appropriate, checked in.

src/osgPlugins/dae/daeRMaterials.cpp
```diff
@@ -589,8 +589,6 @@ bool daeReader::processColorOrTextureType(const osg::StateSet* ss,
     }
     bool retVal = false;
 
-    std::string texCoordSet;
```
openscenegraph: Remove looks appropriate, checked in.

src/osgPlugins/dae/daeRSceneObjects.cpp
```diff
@@ -361,7 +361,7 @@ osg::Node* daeReader::processLight( domLight *dlight )
     lightmodel->setAmbientIntensity(osg::Vec4(0.0f,0.0f,0.0f,1.0f));
     _rootStateSet->setAttributeAndModes(lightmodel, osg::StateAttribute::ON);
 
-    osg::LightSource* lightsource = new osg::LightSource();
```
openscenegraph: I have merged this change but code all paths and all calls to processLight correct handle the memory so the detection as leak is a false positive.

src/osgPlugins/jpeg/ReaderWriterJPEG.cpp
```diff
@@ -689,7 +689,7 @@ class ReaderWriterJPEG : public osgDB::ReaderWriter
               default:
               {
                   OSG_DEBUG << "ReaderWriterJPEG::write_JPEG_file - Error pixel format non supported" << std::endl;
-                return WriteResult::ERROR_IN_WRITING_FILE; break;
```
openscenegraph: Merged.

src/osgPlugins/obj/obj.cpp
```diff
@@ -231,7 +231,6 @@ bool Model::readMTL(std::istream& fin)
     bool usingDissolve = false;
 
     Material* material = 0;// &(materialMap[""]);
-    std::string filename;
```
openscenegraph: Merged

src/osgPlugins/ogr/ReaderWriterOGR.cpp
```diff
@@ -286,7 +284,7 @@ class ReaderWriterOGR : public osgDB::ReaderWriter
 
         if (pointGeom->getVertexArray())
         {
-            OSG_INFO << "osgOgrFeature::multiPointToDrawable " << geom->getVertexArray()->getNumElements() << " vertexes"<< std::endl;
```
openscenegraph: Whahoo, a genuine leak and associated bug! Merged .

src/osgPlugins/tga/ReaderWriterTGA.cpp
```diff
@@ -405,6 +405,7 @@ int *numComponents_ret)
                         break;
                     default:
                         tgaerror = ERR_UNSUPPORTED;
+                        delete[] formattedMap;
```
openscenegraph: Merged

src/osgUtil/SceneView.cpp
```diff
@@ -735,15 +735,13 @@ void SceneView::cull()
...
-            bool computeNearFar = cullStage(computeLeftEyeProjection(getProjectionMatrix()),computeLeftEyeView(getViewMatrix()),_cullVisitorLeft.get(),_stateGraphLeft.get(),_renderStageLeft.get(),_viewportLeft.get());
-
 
             // set up the right eye.
             _cullVisitorRight->setDatabaseRequestHandler(_cullVisitor->getDatabaseRequestHandler());
             _cullVisitorRight->setClampProjectionMatrixCallback(_cullVisitor->getClampProjectionMatrixCallback());
             _cullVisitorRight->setTraversalMask(_cullMaskRight);
             computeRightEyeViewport(getViewport());
-            computeNearFar = cullStage(computeRightEyeProjection(getProjectionMatrix()),computeRightEyeView(getViewMatrix()),_cullVisitorRight.get(),_stateGraphRight.get(),_renderStageRight.get(),_viewportRight.get());
+            bool computeNearFar = cullStage(computeRightEyeProjection(getProjectionMatrix()),computeRightEyeView(getViewMatrix()),_cullVisitorRight.get(),_stateGraphRight.get(),_renderStageRight.get(),_viewportRight.get());
```
openscenegraph: Errh.... you've just removed the cull traversal for the left eye, this is a massive bug.  All for quietening down a benign and silly cppcheck warning.  You can NEVER EVER EVER assume that tools like cppcheck or Coverity actually know what you code is doing, you have to throughly read the code and understand what each part does and why before you can access whether it's a positive or a false positive, and then you still need to understand the code if it is a positive and you need to amend.

This bug is a good example of what it's a bad practice to bundle sweeping changes all together in a single commit and especially PR.  You have to be really thorough in reviewing code to spot potential bugs, really bug ones can lurk in innocent looking changes.

applications/present3D/present3D.cpp
jtorresfabra: I see your point and agree. Will take this into account for next PR's.

src/osgUtil/Simplifier.cpp
```diff
@@ -813,7 +813,6 @@ class EdgeCollapse
         osg::ref_ptr<Point> edge_p1 = edge->_p1;
         osg::ref_ptr<Point> edge_p2 = edge->_p2;
 
-        TriangleMap  triangleMap;
```
openscenegraph: Merged, also removed the associated TrinagleMap typedef

src/osgPlugins/dae/daeRSceneObjects.cpp
jtorresfabra: Correct me if I'm wrong but if you return before freeing the memory like:
https://github.com/jtorresfabra/osg/blob/71732e55f2311d9614992a2217424e96907b5d3e/src/osgPlugins/dae/daeRSceneObjects.cpp#L381 
or 
https://github.com/jtorresfabra/osg/blob/71732e55f2311d9614992a2217424e96907b5d3e/src/osgPlugins/dae/daeRSceneObjects.cpp#L398

aren't the code leaking?

src/osgUtil/TriStripVisitor.cpp
```diff
@@ -607,7 +607,7 @@ void TriStripVisitor::mergeTriangleStrips(osg::Geometry::PrimitiveSetList& primi
             osg::PrimitiveSet* ps = primitives[i].get();
             // remove null primitive sets and all primitives that have been merged
             // (i.e. all TRIANGLE_STRIP DrawElements)
-            if (!ps || (ps && ps->getMode() == osg::PrimitiveSet::TRIANGLE_STRIP))
+            if (!ps || (ps->getMode() == osg::PrimitiveSet::TRIANGLE_STRIP))
```
openscenegraph: This change shows a mis-reading of the original code, it's a subtle bug that's introduced.  Keeping reading it until you spot why the original code is correct and the change introduces a potential crash.

src/osgUtil/SceneView.cpp
jtorresfabra: Arg I thought I have only deleted the computeNearFar variable! this was my intention, my bad!

src/osgAnimation/StatsHandler.cpp
jtorresfabra: makes sense

openscenegraph:
@jtorresfabra I have now go through all the changes and merged ones which are appropriate, discarded or amend ones that are wrong.  In several cases the proposed changes introduced quiet serious bugs, so thankfully I went through them all individually and double checked the original code.  

I would recommend having a look at all my comments to each one so you can spot which changes are valid and which ones aren't and why not.  There are some important lessons in there.  

Overall it really shows how the tools like cppcheck can highlight valid bugs, other times they produce false positives.  For both positives and false positive warnings that you fix there is very serious danger of introducing new bugs. Sometimes these can be far more serious than any of the bugs that were detected.  

One has to view cppcheck and Coverity as crude tools that are can be used as an aid to skilled programmer, but in no way are they full-proof, and when making any changes to code the programmer must fully understand the code they are changing as it's really easy to introduce serous bugs with innocuous looking changes.

As I have merged all the changes that are appropriate I'm now going to close this PR.

jtorresfabra:
Hi @robertosfield Thanks for the time reviweing the changes, I will try to be more careful next time. 

I added also a couple of comments. 

Also this was a first pass, but as you have now also cppcheck you can see that there are a lot of other warnings, so it could be good to have a look.

src/osgPlugins/dae/daeRSceneObjects.cpp
openscenegraph: You are right.  I am sure I did look over the code looking for early returns, but clearly missed them.  So in this case cppcheck didn't produce a false positive.

As a general rule it's far better to use the ref_ptr<> ptr and ptr.release() approach that you implemented rather than using C pointers just in case the flow of control changes as the code evolves, or in the presence of exceptions.  Most of the new code I write these days uses ref_ptr<> for these reasons.

openscenegraph:
I've spent quite a few hours on this, something I hadn't budgeted for this weekend, I'm not about to go trying to fix mostly completely benign issues that cppcheck throws up.  I have lots of other really pressing work to get on wtih.

jtorresfabra:
Yeah sure, I did not mean you have to do it. Sorry if it sound like that and again thanks.

src/osgUtil/TriStripVisitor.cpp
jtorresfabra: umm not sure why, still I don't see it, but ok. Nevermind on this.

openscenegraph: Second reading, now thing you are right.  The code is an oddity because it's catching the possibility of ps being null and if it's not null then only passing if the Mode is TRIANGLE_STRIP.  The normal pattern is if (ptr && some_test_on(ptr)), so it's really easy to mis-read a block that is if (!ptr || some_test_on_ptr). 

The original code uses the explicit (ptr && some_test_on(ptr) pattern which is sound by itself. The additional !ptr check confuses the code.  The revised code is correct but requires one to take extra care over as it doesn't fit the usual pattern.

The fact that we are even discussing this shows that original code and the revised code are both easy to confuse, I really don't like either case. Possible one could just make the assumption that all primtives[i].get() entries will be non null, so the whole catch of null values could be redundant.

jtorresfabra: Agree, removing !ps is a better fix.

openscenegraph: I have just checked in the removal of !ps ||, so the code now reads like the if usage above and will be safe.

jtorresfabra: @robertosfield I think we introduced a bug here. because if !ps the code should enter to if-block also. The way is fixed introduces a bug, as it will never enter the block.

openscenegraph: The new code is correct, far more readable.

If there a null primitives sets in the data being processed the new implementation handles this fine, it just doesn't remove them from the list.  The new code is far easier to read.  Case closed - I have far more important things to go on with.  The issue is closed.

PR #398, opened 2017-12-01
Improve SharedArrayOptimizer

marchelbling:
When an array is shared within and outside a geometry, it's still worth it and safe
to recreate the reference within each individual geometry.

### DIFF
--- a/src/osgUtil/MeshOptimizers.cpp
+++ b/src/osgUtil/MeshOptimizers.cpp
@@ -1207,8 +1207,8 @@ void SharedArrayOptimizer::findDuplicatedUVs(const osg::Geometry& geometry)
     for(unsigned int id = 0 ; id != geometry.getNumTexCoordArrays() ; ++ id)
     {
         const osg::Array* channel = geometry.getTexCoordArray(id);
-        // test if array is shared outside the geometry
-        if(channel && static_cast<unsigned int>(channel->referenceCount()) == arrayPointerCounter[channel])
+        // test if array is shared inside the geometry
+        if(channel && arrayPointerCounter[channel] > 1)
         {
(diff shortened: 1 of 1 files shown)


openscenegraph:
Looking at the code and the changes I don't automatically have the confidence that you have that it's safe.  This type of change really needs a bit of background, test cases illustrating that it's required, and illustrating the change won't break anything.  

As things stand I just don't the confidence there aren't any adverse consequences to this PR so I'm not going to merge.  If we can establish it is safe then I'll be happy to merge.

marchelbling:
I was the original contributor for the `SharedArrayOptimizer`. The reason of this submission in the first place was that lots of mesh optimizers require to duplicate shared arrays to avoid processing some arrays twice and end up with garbage data but this can end up in unnecessarily heavy scenes.

The `SharedArrayOptimizer` first collects the number of reference count inside the geometry and in the whole scene and recreate references. In its current implementation, it only does so for arrays that are not shared elsewhere in the scene. This is too conservative and will leave some unnecessary duplicated arrays. The PR only changes the condition to deduplicate arrays _within_ geometries. This is safe as this happens after `Geometry::duplicateSharedArrays` has been called.

We've been using this fix on lots of models at Sketchfab without issues. Please close this PR if you're not confident about it.

openscenegraph:
Thanks for the extra info.  Unfortunately I sill don't have a good enough gasp of the issue so can't make any quick decision.

The different possibilities are complex enough that the only way I can properly review this PR is to test the existing code on a dataset that illustrates the problem it's trying to resolve, apply the change and see how it affects things.

openscenegraph:
I have done another code review and now understand how _deduplicateUvs is used and it now agree it looks safe so I'm merging the change,

PR #399, opened 2017-12-01
Do not cull inactive culling drawables

mathieu:
Drawable that had  `setCullingActive(false)` could still be culled without this fix


openscenegraph:
I don't understand why you put the isCullingActive into the a line about updating the near far values.  There is a isCullingActive() earlier in the method already.

PR #400, opened 2017-12-01
Do not cull inactive culling drawables

mathieu:
Drawable that had  `setCullingActive(false)` could still be culled without this fix


openscenegraph:
Toggling culling off is not toggling off the compute of the near/far, it's two very different things so it's inappropriate to put in this extra isCullingActive at this point.  There is already an isCullingActive() above this code block.  If there is an issue with the bounding volume then this should be addressed elsewhere.

PR #402, opened 2017-12-06
add a fix in order setArray  work with an user managed interleaved array

mp3butcher:
I developp an InterleavedGeometry overriding drawverteximpl in order to manage myself vbo I bound before calling manual form of Dispatcher::setArray and come into several problems
- as no array is provided, we should assume user already have bound its custom vbo
- a test an ptr is not fitted as offset in a bo can be 0 so i removed it
I don't know it this particular use if vas modifiedcount should be modified but i let it as it was

### DIFF
--- a/src/osg/VertexArrayState.cpp
+++ b/src/osg/VertexArrayState.cpp
@@ -483,6 +483,14 @@ struct VertexAttribArrayDispatch : public VertexArrayState::ArrayDispatch
         callVertexAttribPointer(ext, new_array, (const GLvoid *)(vbo->getOffset(new_array->getBufferIndex())));
     }
 
+    virtual void enable_and_dispatch(osg::State& state, GLint size, GLenum type, GLsizei stride, const GLvoid *ptr, GLboolean normalized)
+    {
+        GLExtensions* ext = state.get<GLExtensions>();
+ 
+        ext->glEnableVertexAttribArray( unit );
+        ext->glVertexAttribPointer(static_cast<GLuint>(unit), size, type, normalized, stride, ptr);
+    }
+
     virtual void dispatch(osg::State& state, const osg::Array* new_array)
     {
         GLExtensions* ext = state.get<GLExtensions>();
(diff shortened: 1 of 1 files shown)


mp3butcher:
I have in mind mods on Geometry in order to support interleaved arrays with 
```
class AttribDescriptor{
   GLenum type;
   int datasize, dataoffset;
}
BufferDataDescriptor{
   BufferData bd;//subjectof description
   int stride;//elmntSize;
   int min,max;//first and last attrib
   vec<AttribDescriptor> desc; //of size==max-min+1
}
in osg::Geometry {
   vec<BufferDataDescriptor> interleavedBDs;
...implies mods in createVAS compilGLObj and drawverteximpl
};
}

```

mp3butcher:
I have in mind mods on Geometry in order to support interleaved arrays with this kind of stuff

class AttribDescriptor{
   GLenum type;
   uint datasize, dataoffset;
}
class BufferDataDescriptor{
   BufferData bd;//subjectof description
   uint stride;//elmntSize,may fit device mem alignement
   int min,max;//first and last attrib
   vec<AttribDescriptor> desc; //of size==max-min+1
  ///(bd elmt accessors based on given desc)
  T * data(int indexininterleavearray, int attributedescindexindesc);
}
additions in class osg::Geometry {
   vec<BufferDataDescriptor> interleavedBDs;//new property
...implies additions in createVAS compilGLObj and drawverteximpl
};
}

cons:

    It involves possible conflicts when user use same attrib index both with VertexAttribute and BufferDataDescriptor
    underlying bd is assumed to be large enough to support its description
    functors are affected (haven't think about implications here)

Do you think the feature worth changes in osg::Geometry ?

openscenegraph:
I am very busy with prep for a work trip and will be away for a week,  Until I'm back I'll no time for reviews/thinking about topics,

mp3butcher:
sorry for the disturbance..have a nice trip

PR #403, opened 2017-12-07
fix logo reader plugin

mp3butcher:
TODO upgrade to a more modern gl (remove glDrawpixel usage)


openscenegraph:
Rather than updating, I'd be inclined to just remove this old plugin.  It's not a good approach for implementing screen logos.

PR #413, opened 2017-12-25
add flexier Image Unit Binding implementation 

mp3butcher:
allow differents IUBs for a single Texture
(perhaps naming should be changed)
todo if accepted: feature has already been tested but the old impl must be removed)

### DIFF
--- a/include/osg/StateAttribute
+++ b/include/osg/StateAttribute
@@ -207,7 +207,8 @@ class OSG_EXPORT StateAttribute : public Object
             VIEWPORTINDEXED,
             DEPTHRANGEINDEXED,
             SCISSORINDEXED,
-            IMAGE,
+            
+            IMAGETEXTURE,
 
             CAPABILITY = 100
         };
--- a/include/osg/StateAttribute
+++ b/include/osg/StateAttribute
@@ -207,6 +207,7 @@ class OSG_EXPORT StateAttribute : public Object
             VIEWPORTINDEXED,
             DEPTHRANGEINDEXED,
             SCISSORINDEXED,
+            IMAGE,
 
             CAPABILITY = 100
         };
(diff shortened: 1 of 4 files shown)


mp3butcher:
on the same pattern I have a Sampler class in order to use different filtering on the same Texture, if u want...

include/osg/StateAttribute
```diff
@@ -122,6 +122,7 @@ class OSG_EXPORT StateAttribute : public Object
           * that type's value. */
         enum Type
         {
+            IMAGE,
```
openscenegraph: New StateAttribute types should be appended to the enum list, as the values are important for binary compatibility and the serialises.

src/osg/TextureImage.cpp
```diff
@@ -0,0 +1,23 @@
...
+    COMPARE_StateAttribute_Parameter(_access)
+    COMPARE_StateAttribute_Parameter(_format)
+    COMPARE_StateAttribute_Parameter(_layered)
+    COMPARE_StateAttribute_Parameter(_level)
+    return 0;
+}
+
+void ImageTextureBinding::apply(osg::State&state)const{
+    if(_target.valid()){
+      osg::Texture::TextureObject *to= _target->getTextureObject( state.getContextID() );
```
openscenegraph: Standard StateAttribute in a StateSet will be applied by the OSG before a StateSet's Texture atttributes so there will be an order of creation/apply issue with using the TextureObject of a Texture. as this could end up being required before the actual texture is applied. 

I have no answer for this issue as I have only do a quick code review trying to understand the feature, but this one stands out as a red flag that would need resolving somehow.

mp3butcher: I understand the issue
I originally had a target.apply before getTextureObject but removing it still give good result with my complex test case (Sparse Voxel Octree Voxelization) ...:/
will investigate if targets are applied in some way....

mp3butcher:
I saw there are issues with VAS, and understand the priority ... Take your time with this pr

openscenegraph:
I have now merged this submission, but first made a series of changes.  The key one is that renamed the and class name header to BindImageTexture.

Issue #412, opened 2017-12-21
Problems after issue #407 fix (VAO and dynamic Geometry change crash)


valid-ptr:
Robert!

First of all, thanks for your fix (issue #407)!
Now there is another issue when I don't use VAO.

My state setup
```
...
osg::State* state = gc->getState();    /// gc is realized

state->resetVertexAttributeAlias(false, 8);    ///// PROBLEM is here ("true" is ok)

state->setCurrentToGloabalVertexArrayState();// **Robert, please FIXME: typo here**
state->getCurrentVertexArrayState()->assignAllDispatchers();
...
state->setShaderComposer(shaderComposer);
state->setShaderCompositionEnabled(true);
```
Glo**a**bal -> Global

As I remember, there is only one global VertexArrayState when VAO is not used. So, I decided've to assignAllDispatchers to global VertexArrayState of the state.
But for now, state->getCurrentVertexArrayState()->assignAllDispatchers will do nearly nothing?

KOS

openscenegraph:
Could you provide a compilable example rather than a code fragment as you obviously have custom code outhwith this code fragment as standard OSG usage doesn't require custom osg::State manipulation.

As a general comment, use of osg::ShaderComposier is deprecated and I would recommend against using it.  osg::ShaderComposer was an experiment that proved to be a dead end and is only now kept around for backwards compatibility for those would were early adopted.  Please use #prama(tic) shader composition instead.

valid-ptr:
There was no possibility to reassign any attrib. dispatchers for osg::State::_globalVertexArrayState after changes in State's attr. location aliases (via resetVertexAttributeAlias() function etc.)


See PR  #418


PS
Nice osg::ShaderComposier is not guilty :)

mp3butcher:
as Robert says a compilable example could be usefull to understand the issue....

openscenegraph:
I have checked in the typo fix of Global and checked in a fix to the VertexArrayState::assign*() methods that enable it to handle cases where the State::s/getUseVertexAttributeAliasing is toggled.  This resolved issues with test code of my own, but I can't say whether it will help with @valid-ptr's usage case as I don't have a full example that illustrate this exact usage case.  

Could you do a update to OSG master and let me know if it resolves the issues.

PR #418, opened 2017-12-28
Fix for #412 issue (VertexArrayState)

valid-ptr:
VertexArrayState assign***ArrayDispatcher functions: reassign possibility restored

This PR solves the problem of using
osg::State::resetVertexAttributeAlias(**false**) + osg::VertexArrayState::assignAllDispatchers()


openscenegraph:
Reverting a bug fix to fix a regression is not a fix, it's just creating a new regression.  I have fixed this issue commit https://github.com/openscenegraph/OpenSceneGraph/commit/94a1a84a59e360b5a30bd8ff4802e44ffe2ca918

PR #421, opened 2018-01-01
complete Imagebinding integration 

mp3butcher:
- add a serializer
- remove old stuff
- adapt example
Just an observation: Perhaps merging BindImageTexture files with BufferIndexBinding ones could resulting in faster build..

### DIFF
--- a/src/osgWrappers/serializers/osg/Texture.cpp
+++ b/src/osgWrappers/serializers/osg/Texture.cpp
@@ -73,22 +73,31 @@ static bool writeInternalFormat( osgDB::OutputStream& os, const osg::Texture& te
 // _imageAttachment
 static bool checkImageAttachment( const osg::Texture& attr )
 {
-    return attr.getImageAttachment().access!=0;
+    return false;
 }
+
+struct DummyImageAttachment
+{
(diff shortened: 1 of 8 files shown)


src/osgWrappers/serializers/osg/Texture.cpp
```diff
@@ -244,13 +244,13 @@ REGISTER_OBJECT_WRAPPER( Texture,
 
     ADD_FLOAT_SERIALIZER( ShadowAmbient, 0.0f );  // _shadow_ambient
 
-    {
+   /* {
```
openscenegraph: We can't remove elements from the serializers without versioning their removal, otherwise we'd break backwards compatibility.

openscenegraph:
Hi Julien, could you break this PR into three, the first which refines the BindImageTexture class and provides it's serializer, the second the change to the example, the third the removal of the old bind image texture support integrated into osg::Texture.  

With the changes for serializers don't ever just comment out serializer code as this will break backwards compatibility.  The way to do it is remove the serializer additions for this feature after a specific SO version.  This will mean the parameters read will need to be ignored, but this is better than prevent old files from loading.

mp3butcher:
done

PR #422, opened 2018-01-01
add Sampler StateAttribute

mp3butcher:
add Sampler StateAttribute and a simple example adding 2 textures with 2 differents samplers

This PR may be  incomplete :
There's a problems of double setup of the sampleparameters. 

One way  to avoid that would be to promote useSampler in State in order not to applyTexParameter in Texture ...I don't know if it could be a proper design for osg GL3Sampler

any thoughts about it?


openscenegraph:
Just one a quick review and my reaction is the the duality of osg::Texture and osg::Sampler both overlapping so much will likely cause confusions, toggling on/off whether to use the settings will cause problems in debugging applications.  We a really butting up against traditional OpenGL and the OSG classes that mirror them and recent evolution of OpenGL,  

Whether we can maintain clean support for both old OpenGL and new OpenGL is a big question.  Most users will be using old OpenGL and we need to maintain backwards compatibility where possible so the old OpenGL paths have to take priority.

If I were to do it clean room them my first approach would be to have a texture class "have a"  sampler parameters, with these sampler parameters being capable of decoupled when necessary. This texture class would provide the basic texture object set up and then defer to the sample parameters to apply the parameters.

mp3butcher:
I knew this one was bad design... I wrote that long time ago 
Your idea of a Sampler property is obviously better design...

PR #423, opened 2018-01-02
Sampler as Texture Property

mp3butcher:


openscenegraph:
I have just done another quiick run though, I'm happier with how things are laid out now.  There is still the overlap between the sampler parameters in osg::Texture base class and the osg::Sampler.

Being able to attach a Sampler to osg::Texture is a good first step. I'm now wondering if the next natural step would be to have osg::Texture use osg::Sampler to store all the texture sampler parameters, and have all the osg::Texture sampler related property method set the sampler values.  This way you'd retain backwards compatibility and avoid issues with settings becoming inconsistent.  

Where GL4.2 samplers aren't available you'd still need to rely on the old texture sampler property settings, might it be possible to have osg::Sampler use the old methods in some way?

mp3butcher:
after few  thinks I remember why I didn't put Sampler as Texture property in my previous PR:
It's not how samplers work:
A sampler is associated to a textureunit not to a textureobject !!!
with the design used here we can't use the same texture with differents samplers

openscenegraph:
Thanks for explanation.  From the sound of it there is an implicit sampler for each texture object, and a sampler for each unit that uses the texture one if no other one is provided.  If this is how the GL driver implements things then I think the osg::Texture being able to use a Sampler to store it's sampler properties would still be fine.  It wouldn't preclude the osg::Sample being used independently.

mp3butcher:
I can't tell there's a default sampler for each tu, my guess is more:
-if a sampler is bound to the tu it will be used
-otherwise classic sampleparams are used

PR #424, opened 2018-01-03
GL3.3 SamplerObject as StateAttribute

mp3butcher:
I reup the first PR https://github.com/openscenegraph/OpenSceneGraph/pull/422 because its behavior is correct:
A sampler is associated to a textureunit not to a textureobject . As a result, a user shoudl be able to associate a single texture with differents samplers (in different ss of course)
Other design and discussion here: https://github.com/openscenegraph/OpenSceneGraph/pull/423
spec:
https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_sampler_objects.txt
manual:
https://www.khronos.org/opengl/wiki/Sampler_Object
TODO:
- find a way to "fix_deprecated_sample_parameters" in order to autoport Texture's sampleparams to Sampler TexAttribute  (perhaps a dedicated DisplaySetting)
- find a way to remove uncessaries calls to applytexparameter when sampler is used (DisplaySettings? check AttributeStack?)
- integrate Sampler factorization in Optimizer (would avoid unecessary calls to Sample::compare -after a fix_deprecated_sample_parameters for ex-)

My guess is a branch would be welcome (once we agree on the design)


mp3butcher:
oups I made a mistake copy/pasting commit message of 662bea2, should be:
add GLSampler **as StateAttribute**
would required a git rebase -i  ... to reword.
Sorry for that

PR #425, opened 2018-01-04
Camera DrawCallbacks recursivity proposal

mp3butcher:
I added nested callbacks on Camera DrawCallback
It would allow multiple postdrawcallback... Don't know if it's usefull...
I duplicate for all callbacks even if i only use postDrawCallback


mp3butcher:
Thanks, I could have do indentation changes myself but i had rather prefer have your confirmation of its validity first.
Do you want me to squash your commit with mine?

PR #427, opened 2018-01-04
add Node associate on Drawable serializers

mp3butcher:
- add Node associate on Drawable based:
Drawable
Geometry
DrawPixels.cpp
ShapeDrawable.cpp
RigGeometry.cpp
ConnectedParticleSystem.cpp
FadeText.cpp
Text.cpp
Text3D.cpp
TextBase.cpp
- add BufferData associate to Image and subclasses
if you think it might require dso increment prefer https://github.com/openscenegraph/OpenSceneGraph/pull/428


mp3butcher:
oups I broke something

PR #430, opened 2018-01-05
Complete Sampler integration

mp3butcher:
generateSamplerObjects create duplicates samplers but this should be fixed after Optimizer::StateVisitor pass...

### DIFF
--- a/include/osg/StateSet
+++ b/include/osg/StateSet
@@ -519,6 +519,8 @@ class OSG_EXPORT StateSet : public Object
         /** call release on all StateAttributes contained within this StateSet.*/
         virtual void releaseGLObjects(State* state=0) const;
 
+        /** generate Sampler StateAttributes from Textures. */
+        void generateSamplerObjects();
     protected :
 
 
@@ -586,7 +588,6 @@ class OSG_EXPORT StateSet : public Object
         ref_ptr<Callback> _eventCallback;
         unsigned int _numChildrenRequiringEventTraversal;
         void setNumChildrenRequiringEventTraversal(unsigned int num);
-
 };
 
 extern OSG_EXPORT bool isTextureMode(StateAttribute::GLMode mode);
--- a/src/osg/Texture.cpp
+++ b/src/osg/Texture.cpp
@@ -1228,6 +1228,9 @@ Texture::Texture():
             _min_filter(LINEAR_MIPMAP_LINEAR), // trilinear
             _mag_filter(LINEAR),
             _maxAnisotropy(1.0f),
+            _minlod(0.0f),
+            _maxlod(-1.0f),
+            _lodbias(0.0f),
             _swizzle(GL_RED, GL_GREEN, GL_BLUE, GL_ALPHA),
             _useHardwareMipMapGeneration(true),
             _unrefImageDataAfterApply(false),
@@ -1255,6 +1258,9 @@ Texture::Texture(const Texture& text,const CopyOp& copyop):
             _min_filter(text._min_filter),
             _mag_filter(text._mag_filter),
             _maxAnisotropy(text._maxAnisotropy),
+            _minlod(text._minlod),
+            _maxlod(text._maxlod),
+            _lodbias(text._lodbias),
             _swizzle(text._swizzle),
             _useHardwareMipMapGeneration(text._useHardwareMipMapGeneration),
             _unrefImageDataAfterApply(text._unrefImageDataAfterApply),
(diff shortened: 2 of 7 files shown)


mp3butcher:
After few tests it appears
There's a problem due to default Texture::ShadowTextureMode (it's LUMINANCE, it should be NONE)

include/osg/StateSet
```diff
@@ -519,6 +519,8 @@ class OSG_EXPORT StateSet : public Object
         /** call release on all StateAttributes contained within this StateSet.*/
         virtual void releaseGLObjects(State* state=0) const;
 
+        /** generate Sampler StateAttributes from Textures. */
```
openscenegraph: This niche convenience function isn't appropriate for a general purpose class like osg::StateSet, it's presence bloats the API and could confuse users. Most OSG applications won't need to use osg::Sampler, it's not a feature that needs wide exposure.

If you want to have such a helper function then have a help function in the include/osg/Sampler header, or put into osgUtil.

src/osg/Texture.cpp
```diff
@@ -1242,7 +1245,7 @@ Texture::Texture():
             _sourceType(0),
             _use_shadow_comparison(false),
             _shadow_compare_func(LEQUAL),
-            _shadow_texture_mode(LUMINANCE),
+            _shadow_texture_mode(NONE),
```
openscenegraph: _use_shadow_comparison being false means that the_shadow_* parameters will be ignored, there isn't any need to have the mode set to NONE.  It will also potentially break 3rd party applications and the serializers that are based on this original default. This seems to be a change for sake of change without thought for the potential consequences.

src/osgWrappers/serializers/osg/Texture.cpp
```diff
@@ -245,10 +245,11 @@ REGISTER_OBJECT_WRAPPER( Texture,
         ADD_ENUM_VALUE( ALWAYS );
     END_ENUM_SERIALIZER();  // _shadow_compare_func
 
-    BEGIN_ENUM_SERIALIZER( ShadowTextureMode, LUMINANCE );
+    BEGIN_ENUM_SERIALIZER( ShadowTextureMode, NONE );
```
openscenegraph: This change will break backwards compatibility.

openscenegraph:
I'm going to close this PR as there a couple of problem with so I don;t think it's appropriate to merge as is.  

I don't think there should be any StateSet::generateSamplerObjects(), this is just bad design stuffing niche functionality into a general purpose class.  Changing as many different OSG classes as you can is not a good trait.  If you want to change any of the core OSG classes you need to REALLY sure that it's warranted and has good applicability across the user base.  It's frustrating to have to keep going back over this issue of changes being splurged over a wide number of core OSG classes.

I keep having to fix indentation/spacing/placement of brackets in your code.  Having to fix this once or twice is OK, but it's now getting a bit silly.  I try to keep the coding style of OSG similar to make it more readable and consistent, not a hot potch of styles that hurts the eyes and brain trying to flip between.

Also there are two many separate issues would up into this PR. The changes to osg::Texture are independent from osg::Sampler, there is no reason to lump them in.  The changes to osg::Texture add new features and change old ones, these two different motivation for change deserve separate PRs.  In this specific case one set of changes in not controversial and could be merged as is, the other would break backwards compatibility so aren't acceptable as is.

Please remember, that I'm a caretaker responsible to maintaining the code in way that works for the whole community, any time we change core functionality we have to be really certain that there won't be consequences to the rest of the community.  The closer the changes are to the core the more critical this issue is.

src/osg/Texture.cpp
mp3butcher: the problem is in osg::StateSet::generateSamplerObjects:
When Sampler:: _shadow_texture_mode is setted to LUMINANCE it yields GL error

src/osgWrappers/serializers/osg/Texture.cpp
mp3butcher: Accounting that in most of the case ShadowTextureMode is left untouched by the user..it may not have be written in files(default) so changing default value in serializer may fix inconsistancies...
Am I right?
And in the case user have setted it exiplicitly there's no issue

mp3butcher:
I admit I take PRs more as a communication mean to point stuff I'm concern with rather than pure submission... There often are identation problems as most of the time I copy paste from my fork to know if the feature is interesting to merge.
Mea Culpa on this, but if it happens again please tell me to fix before actually do it

For this PR, I admit I coded a bit in the wild as commit steps seamed trivial, but roadmap was blocked with the ShadowTextureMode problem. This explains the lack of order in this pr and I'm not surprised you reject it.


I'll try to avoid this but can't give guarantee

PR #431, opened 2018-01-05
allow read of script from zip

mp3butcher:
Nothin to review...  script was the only unimplemented...


openscenegraph:
All code should be reviewed, not matter how trivial. A single mistaken character can break the code.  I know, I've made such mistakes plenty of times in my career.  

Even when copying and adapting code it's still prone to errors.  I would recommend looking up the Copy And Paste Antipattern.

As far as I can tell the only issues with the code in this PR is the spacing and position of brackets, I've fixed this to make sure the code is consistent with the rest of the code around it.

PR #432, opened 2018-01-06
add some macros to help method wrapping

mp3butcher:
add some macros to help method wrapping
examplify with few osg::Group methods

kinda ugly but
It's working stuff I tested with lua script

### DIFF
--- a/include/osgDB/ObjectWrapper
+++ b/include/osgDB/ObjectWrapper
@@ -283,6 +283,115 @@ protected:
     extern "C" void wrapper_compressor_##CLASS(void) {} \
     static osgDB::RegisterCompressorProxy compressor_proxy_##CLASS(NAME, new CLASS);
 
+/** Method wrapping helpers **/
+template <class T>
+class ObjectParameter{
+    T* _t;
+public:
+    inline void set(T* t)          {         _t=t;                                   }
+    inline void set(osg::Object* t){         _t=OBJECT_CAST<T*>(t);                  }
+    inline T* get()                {         return _t;                              }
+    inline osg::Object *getObject(){         return  get();                          }
+    inline osg::Object *getNew()   {         return get();                           }
+};
+
+template <class T,class U>
+class ValueObjectParameter{
+    T _t; osg::ref_ptr<osg::ValueObject> _u;
+public:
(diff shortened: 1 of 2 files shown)


openscenegraph:
I understand the motivation behind these changes, but I don't think this particular incarnation of changes is the best we can do.

In the shader_pipeline branch I've solved part of the problem a bit better than is done in master but introducing ValueObject::set/getScalarValue() method.  It's probably time I cherry-picked these changes back into master.  This resolves part of the issue with getting index etc. parameters from lua scripts as it allows adaptation between doubles -> int/uint/float etc.  

I have yet provided the marco's to help the wrapping, or rolled it out widely across the various serializers.  The TextureCubeMap.cpp serializer is the one to look at in the shader_pipeline.  

I will have a think about the practicility of cherry picking stuff from the shader_pipeline mode.  I'm still officially on holiday so can't spend much time on this today... Back to work fulltime next week.

openscenegraph:
I have now merged the ValueObject changes from shader_pipeline into master.  I will now have a look at merging the improvements to the lua plugin that are in the shader_pipeline branch too.

openscenegraph:
I have checked into master some improvements to the lua plugin from the shader_pipeline branch.

mp3butcher:
great! I'll test it

openscenegraph:
There is a MatrixfUnfirom and MatrixdUniform in the shader_pipeline branch that subclasses from the osg::UniformBase if that's what you mean.  I don't think I've actually tested it - it's just a typedef and I've used the other variants.

The new osg::*Uniform types are easier to support with lua scripting, for a client project we ended up using the shader_pipeline branch simply because setting up a shader based scene graph was so much easier when using the new uniform types.  

Potentially the UniformBase class and it's new subclasses could be merged with master sooner than the while shader_pipeline branch, but it could break some end user code so I want to make sure I'm 100% happyu with the new UniformBase approach before merging this.

Over the next couple of months my plan is to complete the shader_pipeline work so might just concentrate on this and then merge.

Whether the shader_pipeline branch is suitable for you right now I can't say.  I've been merging master into shader_pipeline so it should be possible for you to at least keep access to the new classes that you've written that have been merged with master.

PR #435, opened 2018-01-10
New camera manipulator(Free FPS camera).

KazimKamilov:
New (and very much needed!) manipulator added. FPS camera + WASD moving.
I hope I did everything right. If not, then I'm sorry, I send a pull request for the first time.

### DIFF
--- a/include/osgGA/FreeCameraManipulator
+++ b/include/osgGA/FreeCameraManipulator
@@ -1,0 +1,64 @@
+#pragma once
+#ifndef __OSGGA_FREECAMERAMANIPULATOR__
+#define __OSGGA_FREECAMERAMANIPULATOR__
+
+#include <osgGA/GUIEventHandler>
+#include <osgGA/StandardManipulator>
+
+
+namespace osgGA
+{
+
+class FreeCameraManipulator : public osgGA::StandardManipulator
+{
+public:
+    FreeCameraManipulator();
+	FreeCameraManipulator(const FreeCameraManipulator& om, const osg::CopyOp& copyOp = osg::CopyOp::SHALLOW_COPY);
+
+    virtual osg::Matrixd getMatrix() const;
+    virtual osg::Matrixd getInverseMatrix() const;
+    virtual void setByMatrix(const osg::Matrixd& matrix);
+    virtual void setByInverseMatrix(const osg::Matrixd& matrix);
+    
+    virtual void setTransformation(const osg::Vec3d& eye, const osg::Quat& rotation);
+    virtual void setTransformation(const osg::Vec3d& eye, const osg::Vec3d& center, const osg::Vec3d& up);
+    virtual void getTransformation(osg::Vec3d& eye, osg::Quat& rotation) const;
+    virtual void getTransformation(osg::Vec3d& eye, osg::Vec3d& center, osg::Vec3d& up) const;
+
+    virtual void home(double h);
+    virtual void home(const osgGA::GUIEventAdapter& ea, osgGA::GUIActionAdapter& us);
+
+	/*
+		Sets camera position.
+	*/
+    void setPosition(const osg::Vec3& position);
+	
+	/*
+		Sets speed of camera move.
+	*/
+	void setSpeed(double speed);
+	
+	/*
+		Sets mouse sensitivity. By default, 0.5.
+	*/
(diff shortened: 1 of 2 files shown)


jtorresfabra:
> I hope I did everything right. If not, then I'm sorry, I send a pull request for the first time.
I would say the commit message could be improved :). Like "Add FPS Camera with WASD moving" instead of "Add files via upload". Reading the commit log one does not know what was that for.

KazimKamilov:
Heh... Next time i will try do all better...

jtorresfabra:
You still can ammend your commit message (git commit --amend)

include/osgGA/FreeCameraManipulator
```diff
@@ -0,0 +1,64 @@
+#pragma once
```
bjornblissing: `#pragma once` is not standard c++ and should be removed. You have already include guards so they serve very little benefit.

include/osgGA/FreeCameraManipulator
```diff
@@ -0,0 +1,64 @@
...
+namespace osgGA
+{
+
+class FreeCameraManipulator : public osgGA::StandardManipulator
+{
+public:
+    FreeCameraManipulator();
+	FreeCameraManipulator(const FreeCameraManipulator& om, const osg::CopyOp& copyOp = osg::CopyOp::SHALLOW_COPY);
+
+    virtual osg::Matrixd getMatrix() const;
```
bjornblissing: There is no need to declare these as virtual since they are already declared as virtual in the base class.

src/osgGA/FreeCameraManipulator.cpp
```diff
@@ -0,0 +1,156 @@
+#include <osgGA/FreeCameraManipulator>
+#include <iostream>
```
bjornblissing: `iostream` is an unnecessary include.

src/osgGA/FreeCameraManipulator.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+	return true;
+}
+
+
+bool FreeCameraManipulator::handleKeyDown(const osgGA::GUIEventAdapter& ea, osgGA::GUIActionAdapter& us)
+{
+	switch(ea.getKey())
+	{
+		case 'w':
```
bjornblissing: Use the `KeySymbols` defined in `GUIEventAdapter` instead of raw chars, i.e. `KEY_W` instead of `'w'` etc...

bjornblissing:
First of all, there is already `osgGA::FirstPersonManipulator` which have similar functionality. Any FPS style controller should probably subclass from that instead.

Secondly, the overall quality of this submission is quite low. I have added some general comments to the corresponding commits to give @KazimKamilov some general advice on how to improve the quality for future submissions.

KazimKamilov:
As soon as I have more free time, I will make the manipulator better.

filnet:
And no need to do one commit per file (one commit is enough).

KazimKamilov:
Hey, give me a break! :)

bjornblissing:
@KazimKamilov, you have only fixed the most rudimentary issues with your code. I understand that everyone have to start somewhere, but in my mind the quality of your PR is not up to the expected standard.

I suggest that you redesign this piece of code from the ground up. Start by sub classing from `osgGA::FirstPersonManipulator` instead of `osgGA::StandardManipulator`. And then only add the functions that are **really** needed to make things work.

_(And you are still doing one commit per file. Please stop doing that...)_

openscenegraph:
I don't understand the need for this manipulator, there is existing FirstPersonManipulator, if there are issues with this then my inclination is to discuss these on how it can be improved rather than creating yet another class that doesn't almost the same thing as another class.  Developers who come along after will be left scratching there heads whether to use FirstPersonManipulator of FPSManipulator.

I also no clue why is called a FreeCameraManipulator.  It's pretty meaningless title with many possible interpretations.

Overall it doesn't feel like a submission that is clearly needed or is ready to merge.  As a general commit, it's perfect OK for users to sublcass their own CameraManipulators and keep them in their own applications, there isn't critical need to push things into the core OSG.  It's only appropriate when there is clear value to having it generally available.

KazimKamilov:
Well, I guess if you don't need this manipulator, then say so and close this discussion. It's pretty simple.
Good day to you'll.

Issue #434, opened 2018-01-08
nvtt plugin -fix for 3.4.0 link error under win32


mp3butcher:
building 3.4.0 with 3rdParty_VS2017_v141_x64_V11_full.7z under win7
https://github.com/openscenegraph/OpenSceneGraph/issues/417

- plugin cmake lacks some external libs
squish.lib bc7.lib bc6h.lib (from libnvtt repo) should be add to cmake
Fix:
i added  
`SET(TARGET_EXTERNAL_LIBRARIES ${TARGET_EXTERNAL_LIBRARIES} ${NVTT_LIBRARY} ${NVBC7_LIBRARY} ${NVSQUISH_LIBRARY} ${NVBC6H_LIBRARY}  ${NVTHREAD_LIBRARY} ${NVIMAGE_LIBRARY} ${NVCORE_LIBRARY} ${NVMATH_LIBRARY})`
 to the plugin TARGET_EXTERNAL_LIBRARIES and it worked at least for Release configuration
(FindNVTT.cmake might define multi configuration cmake libs definitions...)

PR #436, opened 2018-01-10
add missing NVTT libraries to nvtt plugin linker

mp3butcher:
fix for https://github.com/openscenegraph/OpenSceneGraph/issues/434
doesn't work for debug configuration
I think it affects also master git
not 100% ok so may be wait till I mod FindNVTT.cmake


openscenegraph:
I think the right way to handle this would be to change the FindNVTT.cmake to assemble all the libs that needed to be linked to into a single variable that the nvtt plugin uses.  This way if there are platform specific build dependencies then can be kept entirely in the FindNVTT.cmake.

mp3butcher:
i'm currently doing it
ready in the pr in 5 min

mp3butcher:
done

mp3butcher:
heu do you have any idea why there are conflicts?

mp3butcher:
I'll renew that... there's still problems

mp3butcher:
mmh conan bug in fact

mp3butcher:
mmh conan bug in fact

mp3butcher:
heu do you have any idea why there are conflicts?

mp3butcher:
arg debug and optimized are mixed

mp3butcher:
now it's okay i think

openscenegraph:
I merged the changes but unforunately they broke the Linux build so I've tweaked the changes to only require the new additional libs under Windows.  The commit is : ff1acce..1872103

I have checked these changes into OpenSceneGraph-3.4 and master.

Could you test these changes out.  Thanks,

mp3butcher:
ok i relaunch conan but these changes seam not to change win32 behavior...

mp3butcher:
MSVC14 win7 osg3.4.0 full build release OK

mp3butcher:
Don't know if you have something to do with it but
Here's the recipe result built for osg34-MSVC14-Release-x64 (I use cracked tools so can't give any reliability guarantee about generated binaries) 
https://drive.google.com/file/d/1OFfrnU57hitFkMoHcMwMU0H0do-wf1qL/view?usp=sharing

PR #440, opened 2018-01-12
removal of the Geometry FinishedObjectReadCallback

mp3butcher:
that avoid saving custom bufferobject configuration
(it really is a pity that setting a nice bo sharing stuff being destroyed at loading sniff)

### DIFF
--- a/src/osgWrappers/serializers/osg/Geometry.cpp
+++ b/src/osgWrappers/serializers/osg/Geometry.cpp
@@ -109,19 +109,6 @@ ADD_ARRAYDATA_FUNCTIONS( FogCoordData, FogCoordArray )
 ADD_ARRAYLIST_FUNCTIONS( TexCoordData, TexCoordArray, TexCoordArrayList )
 ADD_ARRAYLIST_FUNCTIONS( VertexAttribData, VertexAttribArray, VertexAttribArrayList )
 
-struct GeometryFinishedObjectReadCallback : public osgDB::FinishedObjectReadCallback
-{
-    virtual void objectRead(osgDB::InputStream&, osg::Object& obj)
-    {
-        osg::Geometry& geometry = static_cast<osg::Geometry&>(obj);
-        if (geometry.getUseVertexBufferObjects())
-        {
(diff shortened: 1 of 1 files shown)


openscenegraph:
This callback hack was, perhaps still is, required to avoid a bug that existed previously.  I don't recall the details off the top of my head, we'd need to go look at the osg-users archives and git log for this and associated files.  Once we track down the usage case that required this callback we can then test to see if the original bug is now not an issue so this callback can be removed. 

Until we do this work this submission simply isn't safe to merge.

mp3butcher:
commit 57a0065d08b3aa5b06da54a7b78ea2ae36cb2c9a
Author: Robert OSFIELD <robert.osfield@gmail.com>
Date: Thu Oct 28 14:04:57 2010 +0000

Added use ObjectWrapper's FinishObjectReadCallback to fix VertexBufferObject handling in osg::Geometry
and a call to the TileLoaded callback in osgTerrain.

openscenegraph:
I have done some testing at my end and so far haven't come across a problem with not using the GeometryFinishedObjectReadCallback.  I know if this is just that the underlying issues that required it have been fixed elsewhere in the OSG, or whether I haven't done the right tests to tease out the problem case.  At this point I can't say with 100% confidence that it's safe to remove this hack yet.  Thought I'm inclined to think it'll be OK, and if we do fine it is necessary in the future we'll need to look into the underlying issues and fix this than require this hack.

Given that there is still some uncertainty the ability to re-enable this hack quickly would be useful so rather than remove it completely from the code base I have guarded with an #ifdef that for now is not defined, so in effect the callback is no longer called, and you have edit the code to re-enable it.  The commit is:

https://github.com/openscenegraph/OpenSceneGraph/commit/3e6d118d10fe6379b95cb7ee6dcb9e1524d2add3

With this commit being a bit safer I believe this PR is no longer required so am closing it.

https://github.com/openscenegraph/OpenSceneGraph/commit/3e6d118d10fe6379b95cb7ee6dcb9e1524d2add3

mp3butcher:
great!

PR #443, opened 2018-01-13
osgSampler1 add MinLOD MaxLOD and LODBias to Texture

mp3butcher:
and add them to serializer


openscenegraph:
Thanks for breaking things into separate PR.

PR #444, opened 2018-01-13
Osg sampler2  add static helper func in order to generate Samplers of a Samplerless StateSet

mp3butcher:


openscenegraph:
This example PR includes all the changes to Texture.  Was this meant? 

I've merged the Texture changes, I'm not sure how well git will handle me merging this PR with these same changes... I'm hoping that it git will keep from tangling things up but not 100% sure as I haven't tried it work before.  In two minds whether to just click merge and see what happens....

openscenegraph:
I did a trial manual merge on the command line and it just picked out the Sampler changes without the other commits so have gave me the confidence to go ahead and to the merge via the github webpage and it looks like it worked just fine:-)

PR #450, opened 2018-01-17
proposal for bufferdata readback callback

mp3butcher:
I did that for a while now...
It seams ok 
would require more testing than just the example case

### DIFF
--- a/include/osg/BufferIndexBinding
+++ b/include/osg/BufferIndexBinding
@@ -221,22 +221,25 @@ class SyncBufferDataCallback : public osg::NodeCallback{
 public:
     typedef GLenum BufferDataAccess;
 
-    SyncBufferDataCallback(BufferData* bd):
+    SyncBufferDataCallback(BufferData* bd = 0):
         _bd(bd),
         gpuproduced(false),
         cpuproduced(true),
         _CpuAccess(GL_READ_ONLY),
-        _GpuAccess(GL_READ_ONLY){}
+        _GpuAccess(GL_READ_ONLY),
+        _mappingflags(GL_MAP_READ_BIT){}
 
     void setBufferData(BufferData* bd) { _bd=bd; }
     const BufferData* getBufferData() const { return _bd.get(); }
 
     inline void setHostAccess(BufferDataAccess a) { _CpuAccess=a; }
-    inline BufferDataAccess getHostAccess()const { return _CpuAccess; }
+    inline BufferDataAccess getHostAccess() const { return _CpuAccess; }
 
     inline void setDeviceAccess(BufferDataAccess a) { _GpuAccess=a; }
-    inline BufferDataAccess getDeviceAccess()const { return _GpuAccess; }
+    inline BufferDataAccess getDeviceAccess() const { return _GpuAccess; }
 
+    inline void setMappingFlag(GLbitfield a) { _mappingflags=a; }
+    inline GLbitfield getMappingFlag() const { return _mappingflags; }
 
     inline void setReadBackCamera(Camera* a) { _cam=a; }
     inline Camera* getReadBackCamera() const { return _cam.get(); }
@@ -261,6 +264,7 @@ protected:
     bool gpuproduced, cpuproduced;
     BufferDataAccess _CpuAccess;
     BufferDataAccess _GpuAccess;
+    GLbitfield _mappingflags;
     ref_ptr<Camera::DrawCallback> _dcb;
     observer_ptr<Camera> _cam;
     OpenThreads::Mutex _lock;
--- a/include/osg/BufferIndexBinding
+++ b/include/osg/BufferIndexBinding
@@ -21,6 +21,7 @@
 #include <osg/Export>
 #include <osg/BufferObject>
 #include <osg/StateAttribute>
+#include <osg/Camera>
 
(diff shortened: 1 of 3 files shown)


include/osg/BufferIndexBinding
```diff
@@ -213,6 +214,75 @@ class OSG_EXPORT ShaderStorageBufferBinding : public BufferIndexBinding
...
+public:
+    BufferDataReadBack(SyncBufferDataCallback* u=0): Camera::DrawCallback(), _upcb(u){}
+    inline void setSyncBufferDataUpdateCallback( SyncBufferDataCallback *u){ _upcb=u; }
+
+    virtual void operator () (osg::RenderInfo& renderInfo) const;
+
+    /** read back BufferObject to its BufferData (reimplement for custom behavior) */
+    virtual void readBackBufferData(osg::RenderInfo& renderInfo) const;
+protected:
+    SyncBufferDataCallback* _upcb;
```
openscenegraph: This is just asking for a dangling pointer issue.  From a first run through I haven't been able to see the whole thing clearly enough to suggest exactly what you need to resolve it.  It's not safe as is so will need some form of resolution before it's safe to merge.

mp3butcher:
the  _upcb is a weak pointer as it's setted during SyncBufferDataCallback update traversal
perhaps public setter should be removed and make ReadBufferDataCallback friend of SyncBufferDataCallback is order it access protected _upcb

include/osg/BufferIndexBinding
mp3butcher: the _upcb is a weak pointer as it's setted during SyncBufferDataCallback update traversal.
in my last commit, i removed public setter and made class BufferDataReadBack friend of SyncBufferDataCallback in order it access protected _upcb

mp3butcher:
I redesignes in order not to exibit BufferReadCallback to the user...

mp3butcher:
I change atomic counter example in order to test the 3 different atomic glsl behavior(atomic_counter, ssbo,iamgeapi) and test readback in each case

mp3butcher:
@robertosfield 
(not directly linked to the pr)
I began to integrate unmutable buffer into osg but i came into problem with persistant mapping and stl vector used in osg:
The only solution i found is to interact with underlying allocator...it's forbidden by the standard but the boost::vector implementation allow it....
Would it be wise to introduce a boost dependancy so deep in osg just for the persistant mapping?

openscenegraph:
That's an easy one, no it's not acceptable to introduce a boost dependency into the core OSG.

It's possible to create a new set of osg::Array subclasses that don't use std::vector<> that you keep separate from the core, such as in your own program.

Or at a stretch perhaps just implement a osg::Array implementation that doesn't use std::vector<> at all, just implement the most commonly used std::vector<> methods.

mp3butcher:
a new set of Array would be a clean solution to use custom allocator however glmapping requires a context to be bound in order to allocate dma memory...further i seek a way to use driver pinned memory:
- throught the same interface as regular array (MixInVec interface)
- demand  (user might want treat it like regular owned memory and during compileBuffer pinned memory is created, user content copied then ownership transfered)
I reach the 2nd point without any problem but don't achieve the 1st point
I'll let pinnedmemory accessible trhuoght a dedicated pointer in Array but it will be at user charge not to use vector ...:/  It's seams the easiest compromise....

openscenegraph:
A std::vector<> essentially has a pointer to the memory allocated for the elements, and then either pointers or counters to the reserve and the size.  osg::Array get's access to the base pointer and for passing to OpenGL via a virtual method.

I guess one could implement an osg::Array class that has the pointer directly within in, and then the osg::Vec3Array etc. reuse this pointer casting from GLvoid* to osg::Vec3* etc.

Needing to have the memory allocated by GL complicates things, especially for multi-context systems.  You may actually want the osg::Array to defer to the VBO assigned to it as VBO's have support for multiple GLBufferObjects, one per context.

I haven't thought deeply about the topic though, it's certainly a sticky one....

mp3butcher:
I doubt this callback works on image ... but I can make the same  stuff for Texture readback...

openscenegraph:
I have just done a code review, and oh boy, noooo, this application specific hacks shoudln't be in the core OSG. Perhaps embedded in an example, but even this it's really not at all recommend practice to go round modifying the scene graph during the traversal to go add callbacks.   There has to be a far less hacky way to do what you want to do, but as the changes are so convoluted it's hard to know exactly what to recommend.  This PR system provided by github really isn't the place for such discussions - please raise a discussion topic on osg-users ML/forum.

mp3butcher:
My opinion:
U're right that BufferIndexBuffer is not a fitted place...(i can't remember why i put it there)
I coded this a long time ago but never used it (I always do that manually in drawcalback)
what a pity there's no thread safe way to mod scenegraph..:/
I haven't a complete understanding of the threading interleaving in osg..Do you think a mutual exclusion step (of both updaters and renderers) could be added somewhere in the mainloop in order to provide a standardized way of mod scenegraph?

openscenegraph:
Could you please stop these misusing the PR system for discussions.  If you want to have a discussion the place to do it on osg-users ML/forum.

I really need to unsubtle now as you don't yet seem to have sunk in. I SERIOUSLY FED UP WITH THE WAY YOU ARE ABUSING THE PR SYSTEM.

PR #451, opened 2018-01-18
Updated CMake files to find correct frameworks for iOS builds

aterenzi:
When trying to use CMake to generate the Xcode project targeting iOS, the path to the following framework were wrongly referring to the MacOS versions:

OpenGLES, CoreMedia, CoreVideo, AVFoundation, QuartzCore and SystemConfiguration

So the generated Xcode project could not build successfully.

Modified also the avfoundation plugin because the duration of video files was not read properly.


openscenegraph:
I have done a quick look through the changes and it looks like the cmake find functions aren't handling the rev of iOS/XCode that you are using.  It may be resolvable be cmake updating their scripts to properly support the latest stuff from Apple rather than pushing hardwiring into the OSG side.

aterenzi:
Ok, I will try again using the latest version of cmake, hopefully they can manage to find the correct version for the iOS frameworks.

There’s a second commit though, it is related to a bug of avfoundation plugin, should I send a different request for that?

Thanks.

openscenegraph:
Yes, use separate PRs for separate fixes.  Also when trying to explore what might be the appropriate fix use the osg-users mailing list/forum to see if other using iOS/OSX have any insights.  I don't personally have iOS/OSX system so can't provide this myself.

aterenzi:
I have tried the latest version of CMake, 3.10.2, and I confirm that it still cannot manage to set the correct paths of any of the frameworks I mentioned earlier, it also fails to set the path to zlib (for which I do have a fix as well).

So far, if one wants to build for iOS, even if he follows the platform specific instructions available on the website, he has to manually set every frameworks single path in the CMake cache and even later in the Xcode project in some cases.

I will share this information on the forum to see if others are experiencing the same issues.

src/osgPlugins/avfoundation/OSXAVFoundationVideo.mm
```diff
@@ -345,7 +345,7 @@ CVBufferRef getLastFrame() {
         object:_data->avplayer.currentItem];
     
     
-    _videoDuration = CMTimeGetSeconds([_data->avplayer.currentItem duration]);
+    _videoDuration = CMTimeGetSeconds([[_data->avplayer.currentItem asset] duration]);
```
mp3butcher: not a c99 syntax

mp3butcher:
you closed it? Changes seam motivated at first glance...

aterenzi:
Yes, I closed the request because I realised that other commits were being added to the pull request, my mistake, sorry. I will send separated pull requests for each commit or send the updated files to the mailing list.

PR #462, opened 2018-01-29
fix FindGTA CMake module

fiesh:
The following change (in accordance with how `FindGIFLIB.cmake` does it) is necessary for `libgta` to be found correctly when building with `mxe`.

### DIFF
--- a/CMakeModules/FindGTA.cmake
+++ b/CMakeModules/FindGTA.cmake
@@ -18,40 +18,37 @@ IF(PKG_CONFIG_FOUND)
 
 ENDIF(PKG_CONFIG_FOUND)
 
-IF(NOT GTA_FOUND)
-    FIND_PATH(GTA_INCLUDE_DIRS gta/gta.hpp
-        $ENV{GTA_DIR}/include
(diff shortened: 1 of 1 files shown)


openscenegraph:
I'm confused by the need for this PR, the FindGIFLIB.cmake doesn't have any check for CMake providing a CMake script to find the libs, while our FindGTA.cmake does.  The FindGTA.cmake only falls back to the manual search for headers and libs in the case when the Cmake Find package doesn't succeed.

The way you've changed things effectively ignores that the Cmake find package might find, so why even maintain it.  It strikes me that there is something going on your system that needs further investigation rather than modifying the OSG to workaround unspecified problems. 

 Please use the osg-users mailing list/forum to discuss the issues you have seen then we can discuss what might be amiss and what the best resolution would be.  When you do could you please explain the acronyms you use as not everyone will know what you are talking about otherwise.

I'm closing this PR as changes suggest a problem elsewhere that needs investigation, and the changes that have been made aren't consistent with whole findGTA.cmake script, as they ignore results.

PR #464, opened 2018-01-30
Not for merging yet: Immutable buffers + persistance mapping proposal

mp3butcher:
Here's what I did for the moment on immutables buffer integration 
**Not candidate to merge for the moment
I pr just to make review easier**

In this implementation I made DMAbuffer a new attribute of BufferData.
The problem (if there's one) is that user should be aware to interact through this buffer and not vector interface.
I didn't find a way to use vector interface for both dma and ::new memory allocation policy without violating the standard (interact directly with vector allocator state of a boost::vector)
If this compromise is enough for community requirement I'll continue on this path (and make my PR cleaner)

### DIFF
--- a/src/osg/BufferObject.cpp
+++ b/src/osg/BufferObject.cpp
@@ -1,5 +1,6 @@
 /* -*-c++-*- OpenSceneGraph - Copyright (C) 1998-2006 Robert Osfield
  * Copyright (C) 2012 David Callu
+ * Copyright (C) 2018 Julien Valentin
  *
  * This library is open source and may be redistributed and/or modified under
  * the terms of the OpenSceneGraph Public License (OSGPL) version 0.0 or
@@ -53,7 +54,7 @@ unsigned int GLBufferObject::BufferEntry::getNumClients() const
 GLBufferObject::GLBufferObject(unsigned int contextID, BufferObject* bufferObject, unsigned int glObjectID):
     _contextID(contextID),
     _glObjectID(glObjectID),
-    _profile(0,0,0),
+    _profile(0,0,0,0),
     _allocatedSize(0),
     _dirty(true),
     _bufferObject(0),
@@ -104,7 +105,7 @@ void GLBufferObject::assign(BufferObject* bufferObject)
     }
     else
     {
-        _profile.setProfile(0,0,0);
+        _profile.setProfile(0,0,0,0);
 
         // clear all previous entries;
         _bufferEntries.clear();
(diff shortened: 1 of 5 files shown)


src/osg/BufferObject.cpp
```diff
@@ -120,7 +122,7 @@ void GLBufferObject::clear()
 void GLBufferObject::compileBuffer()
 {
     _dirty = false;
-
+_persistantDMA=0;
```
mp3butcher: oups this should be removed

filnet:
I believe "persistant" is spelled "persistent" in English.

mp3butcher:
close/open to force travis job

mp3butcher:
close/open to force travis job

mp3butcher:
as i said it's an uncleaned and not deeply tested proposal
I'm waiting @robertosfield word in order to clean it an make an example

mp3butcher:
in all case https://github.com/openscenegraph/OpenSceneGraph/pull/464/commits/d740f513b40a07dcc786e8602797ea7274820f7b is cherry pickable directly i think

openscenegraph:
It really feels like you are trying to test me, yet again indentation is not consistent with the rest of the OSG, and you have random other fixes in there for no reason.  I really fed with having to wade through this stuff. It's not the first, second or third time I've raised this issue.  PR's to the OSG main project website are things you really need to take care over, I shouldn't have to be pushing back all the time to fix basic stuff.

As for the actual functionality attempted here. Well pushing context specific data into BufferData which is designed to be graphics context agnostic is broken, the whole point of this design is to allow the subclasses of BufferData to just concentrate on application data and not have to worry about what context are created by the application - it's all about decoupling the scene graph from the viewer.  It's one of the OSG great assets.  And here you won't to throw away a clean design for a niche feature only usable on a GL4 and has absolutely no value for all other GL/GLES versions.

With niche feature think how to implement things in a custom way where the changes are kept local, not sprayed across the core OSG.

Also the change for size_t, if this is a real issue could you post a separate PR for it and an illustrate of what build combinations it may be an issue.  If it's a genuine issue that we'll need to backport this change.

mp3butcher:
As I mentionned everywhere this PR is dirty and not for merging...so I didn't care indentation or lexic issue for the moment...

I really would like to take pr seriously but in most of the case I must discuss first of the pr... and what better way to discuss a pr than a pr ?!

I'd though the cherrypick of https://github.com/openscenegraph/OpenSceneGraph/commit/d740f513b40a07dcc786e8602797ea7274820f7b would be obvious as size_t is a namespaceless type define in string.h (it's just names collisions paranoia talking)

Now that civilities are done let's discuss the center of indecision in this pr: 
how to implement persistent buffer in osg?
I'm agree with you on the global frame but i think you miss the fact that DMAbuffer principal usage is to write the memory in the update thread and commit changes to the renderer thread throught glcommit
It seams strange to call gl from update thread but that how it works..No?

openscenegraph:
You are abusing the PR system for this type of discussion.  

The right place for discussions is on osg-users ML/forum, this way everyone subscribed can see, contribute and benefit the discussion rather only those who happen to follow PR threads.  I have never suggested to you that pushing all these PR's is a good thing, please stop doing it.  

As for indecision, I've got other time critical work, I'm not paid to jump immediately and review your every speculative PR.  I have to prioritize where I spend my time.  I have  even have a family that I occasionally like to spend time with too.  Time replying here is time that is better spent elsewhere.  Please stop misusing the PR system.

mp3butcher:
okay i wil try to explain myself on the forum and put link to my branches on it... (hoping i'll not loose my time with unconcerned users..)

PR #465, opened 2018-02-01
Not for merging: osg graphicobjects homogenization proposal

mp3butcher:
**Not candidate to merge but would require community feeback as i change a lot of things**
sorry in advance for the 16 files changes in one commit :~/
but i need to know if this fits your insights about osg

I'll try to remove non pertinents stuff before you look at it (I fail believing a GLProfileInterface can be usefull)

ChangeLog (in absence of commit history)
- introduce a osg::TextureObject ( buffered_objectPCTOs  [PCTO==Texture::TextureObject] )
- inject a base class  PerContextGraphicObject  (base of BufferObject, TextureObject  (bufferedPCTOs), ProgramsObject ..) that is just a per context GL object container interface
- deprecate Image less Texture by adding an Image with NULL data  and make unrefimageafterapply just deallocate Image (may require further data==0 tests to ensure retrocomp)
- remove buffered_objectPCTOs from Texture to use  the osg::TextureObject  of its Image instead

It fixes the bug without rollback and bring some homogenity ...

If it's okay I'll  redo the pr step by step including involved changes in serializers

### DIFF
--- a/include/osg/BufferObject
+++ b/include/osg/BufferObject
@@ -94,7 +94,7 @@ class State;
 class BufferData;
 class BufferObject;
 
-class BufferObjectProfile
+class BufferObjectProfile : public GraphicsObjectProfile
 {
     public:
         BufferObjectProfile():
@@ -209,13 +209,11 @@ class OSG_EXPORT GLBufferObject : public GraphicsObject
         }
 
         /** release GLBufferObject to the orphan list to be reused or deleted.*/
-        void release();
+        virtual void release();
 
-        inline bool isDirty() const { return _dirty; }
+        virtual GraphicsObjectProfile &getProfile(){ return _profile; }
 
-        void dirty() { _dirty = true; }
-
-        void clear();
+        virtual void clear();
 
         void compileBuffer();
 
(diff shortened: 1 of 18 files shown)


mp3butcher:
@robertosfield 
sorry in advance for the 16 files changes in one commit
but i need to know if this fit your insights about osg

mp3butcher:
After few thought this design suxxx
Texture MUST own tos

mp3butcher:
reopened for the interest of a base class for TextureGraphicObject and BufferObject

openscenegraph:
I really don't understand the motivation behind this proposal, it's just a collection of changes to do things a bit differently and give more operations more lines of code to do the same thing.

The OSG is open source so people can just go changing as many files as they want.  Just because you can doesn't mean it's a good thing to do it.  For a project lead myself having to wade through many files trying to understand the motivation and the safety of all these changes is hideous.

Might I suggest that you reign in your amitions of modifying as many OSG files as possible to achieve non descript aims.  If you have a specific feature you want to implement then use a scaple not a chainsaw.  The OSG is a mature project used by thousands of developers, we need to try and keep it as solid as a base as we can.  Any additions or changes need to justified because they offer real value.

The change of osg::Texture to enforce an osg::Image always being assigned is simply not viable - it'll break loaders and end users applications as any place that code does a texture->getImage()==NULL will silently fail at runtime, the bugs that ensue will be subtle and hard to track down.  Such a change to a mature project would be crazy.

I'm closing this PR because it's a deluge of changes without any clear benefit and huge risks involved in spraying so many changes across so many parts of the OSG.  Really this type of major change is something that shouldn't be a PR on the main repository.  If you want to do crazy experiments and publish them create a branch on your own repository and then call for feedback from the community.

Having such PR's here on the main site just gets in the way of me managing proper submissions that are ready to be merged.  Please remember I'm the caretaker of something that is used by a large and diverse group of developers, this is where my responsibility lies.

mp3butcher:
it happens one time a year: a chainsaw psychotic night rearranging osg.....that how i learned osg...dismantling it one time a year...
I'd though this "mutation" brings some homogeneity and tell a lot about insights i've got...but whithout any community concern..
I should have close this one by myself...

PR #467, opened 2018-02-07
minimum changes to fix BindImageTexture

mp3butcher:
re-add Texture changes of
https://github.com/openscenegraph/OpenSceneGraph/pull/466
and fix issue 
http://forum.openscenegraph.org/viewtopic.php?t=17102
by making  _modifiedcount Texture attribute (see commit log)

### DIFF
--- a/examples/osgcomputeshaders/osgcomputeshaders.cpp
+++ b/examples/osgcomputeshaders/osgcomputeshaders.cpp
@@ -69,8 +69,7 @@ int main( int argc, char** argv )
     sourceNode->setDataVariance( osg::Object::DYNAMIC );
     sourceNode->getOrCreateStateSet()->setAttributeAndModes( computeProg.get() );
     sourceNode->getOrCreateStateSet()->addUniform( new osg::Uniform("targetTex", (int)0) );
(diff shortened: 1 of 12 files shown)


openscenegraph:
Making _modifiedCount part of the base class is broken because you end up having to duplicate the data structure for the array version of texture.

Also I can grasp how 12 changed files could be at all a minimal change set.  You've merged three set of changes into one commit.  

Adding the texture unit as into BindImageTexture just feels like this class is getting more a more tightly coupled state that breaks the orthogonal design that the OSG and OpenGL try to maintain, the fact you are having to do this tight coupling suggests that approach is far from idea and may need revising at a fundamental level. 

I'm closing this PR as it just contains too many different problems.

mp3butcher:
as i mentionned in the commit the semantic of _modifiedCount is not clear:
- in unlayered Texture it refer to texture object dirtiness
- in layered texture it's used for as hint not to upload unchanged image layers

Make  _modifiedcount a Texture attrib make clear to everyone that _modifiedCount represents "texture object modified count" and not a context-unrelated-inner-tricks-not-to-upload-image-already-there

Concerning how I design BindImageTexture, we already have the discussion **it's not a textureatrribute**

PR #468, opened 2018-02-07
fix Visual Studio warnings about inheritance via dominace (C4250)

LaurensVoerman:


openscenegraph:
Hi Laurens,  these warnings tell me that this example is badly written, using multiple inheritance in the class is just inappropriate.  There really should be two classes.

openscenegraph:
I have re-factored the problem class into a dedicated event handler and a nested helper class that subclasses from NodeVisitor.  This avoids the inappropriate use of multiple inheritance and will resolve the warning with it.

The fixe PR is:
https://github.com/openscenegraph/OpenSceneGraph/commit/f767ab0cc967395aa4fe9573f2fbad4a9b762bae

I will close this PR as it's no longer required.

PR #470, opened 2018-02-08
premisses to BindImageTexture fix

mp3butcher:
I followed your advice for getModifiedCount 
http://forum.openscenegraph.org/viewtopic.php?t=17102

### DIFF
--- a/examples/osggpucull/osggpucull.cpp
+++ b/examples/osggpucull/osggpucull.cpp
@@ -376,10 +376,7 @@ struct IndirectTarget
         std::string uniformName = uniformNamePrefix + char( '0' + index );
         osg::Uniform* uniform = new osg::Uniform(uniformName.c_str(), (int)index );
         stateset->addUniform( uniform );
-        stateset->setAttribute(indirectCommandImageBinding);
-        stateset->setTextureAttribute( index, indirectCommandTextureBuffer.get() );
-
-
+        stateset->setTextureAttribute(index, indirectCommandImageBinding);
     }
 
     void addIndirectTargetData( bool cullPhase, const std::string& uniformNamePrefix, int index, osg::StateSet* stateset )
@@ -392,9 +389,7 @@ struct IndirectTarget
 
         osg::Uniform* uniform = new osg::Uniform(uniformName.c_str(), (int)(OSGGPUCULL_MAXIMUM_INDIRECT_TARGET_NUMBER+index) );
         stateset->addUniform( uniform );
-
-        stateset->setAttribute(instanceTargetimagebinding);
-        stateset->setTextureAttribute( OSGGPUCULL_MAXIMUM_INDIRECT_TARGET_NUMBER+index, instanceTarget.get() );
+        stateset->setTextureAttribute(OSGGPUCULL_MAXIMUM_INDIRECT_TARGET_NUMBER+index, instanceTargetimagebinding);
     }
 
     void addDrawProgram( const std::string& uniformBlockName, osg::StateSet* stateset )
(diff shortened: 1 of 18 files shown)


mp3butcher:
i'll reopen this later

include/osg/Texture
```diff
@@ -771,6 +774,7 @@ class OSG_EXPORT Texture : public osg::StateAttribute
           * for all graphics contexts. */
         virtual void releaseGLObjects(State* state=0) const;
 
+        virtual unsigned int &getModifiedCount(unsigned int contextID) const = 0;
```
filnet: Seems like your other changes make it possible to not return a reference (which contradicts the const nature of the method).

include/osg/Texture2DMultisample
```diff
@@ -46,7 +46,12 @@ class OSG_EXPORT Texture2DMultisample : public Texture
         {
           return GL_TEXTURE_2D_MULTISAMPLE;
         }
-
```
filnet: those empty lines should be kept

include/osg/Texture
mp3butcher: the method is const because it concerns a mutable attributes...

include/osg/Texture2DMultisample
mp3butcher: haven't see that..

mp3butcher:
I close this until I achieve to make all examples work with it....I've got a strange bug in gpucull I haven't with others my branches...

mp3butcher:
I've found the problem:
a strange behavior of stateset.getTextureAttribute(tu,type) : member searched is always set to 0 !?

include/osg/Texture
filnet: yes, but your changes make it possible to not return a reference to the mutable.

src/osg/TextureRectangle.cpp
```diff
@@ -202,10 +202,10 @@ void TextureRectangle::apply(State& state) const
         {
             _subloadCallback->subload(*this, state);
         }
-        else if (_image.valid() && getModifiedCount(contextID) != _image->getModifiedCount())
+        else if (_image.valid() && _modifiedCount[contextID] != _image->getModifiedCount())
         {
             // update the modified count to show that it is up to date.
-            getModifiedCount(contextID) = _image->getModifiedCount();
+            _modifiedCount[contextID] = _image->getModifiedCount();
```
filnet: this is one example where getModified() had to return a ref (and does not have to anymore)

openscenegraph:
Really?  After everything I suggested about "what to" and what not do you go follow the path of "what not to do".

If you are modifying dozens of files to implement a niche StateAttribute there is something seriously wrong with the approach.  18 Files changed is a yet another sliurge over lots of the OSG.  The skill of a programmer is measured by the quality of their work not the quantity.

You need to think more of surgeon with a scalpe and not a butcher with a cleaver.  

These are real lessons to take away, to practice in your own code and especially submissions to the OSG.  You will end up a better programmer.

For this particular issue when I will be implement my previous suggestion.  I won't be able to get to it today, but hopefully midweek.

mp3butcher:
Really! I just obey your guidelines...
I did modify a lot of useless calls to getModifiied in the textures because it surely became cost....

openscenegraph:
I have checked in two commits, the first is the addition of a bool Texutre::isDirty(contextID) method and it's implementation in the various Texture subclasses:

https://github.com/openscenegraph/OpenSceneGraph/commit/b8e5f22cb7ad2b24451e55e8e117b9e4a404104f

The second commit is adding usage of the Texture::isDirty(contextID) into the BindImageTexture class, and usage of the State::applyTexturtueAttribute() API to work with the OSG's lazy state updating.  

https://github.com/openscenegraph/OpenSceneGraph/commit/68bb67312e417fc6d93bf8856609587114a380c6

The number of files changed was 8 headers and 1 cpp.  I haven't modified the examples, but this was always a separate commit and am happy for others to pitch in if required.  I haven't changed BindImageTexture to be a TextureAttribute as at this stage I think we can get away without doing this.

I would recommend studying these commits to see how one can efficiently and safely encapsulate functionality that is useful for multiple purposes and it's implementation is kept orthogonal to existing functionality.  All while changing significantly less files. The key principle is to Do more, risk less and do so with less code changes.

examples/osggpucull/osggpucull.cpp
```diff
@@ -858,33 +853,31 @@ struct ResetTexturesCallback : public osg::StateSet::Callback
...
-                img->dirty();
-        }
-        for(it=texUnitsDirtyParams.begin(), eit=texUnitsDirtyParams.end(); it!=eit; ++it)
-        {
-            osg::TextureBuffer* tex = dynamic_cast<osg::TextureBuffer*>( stateset->getTextureAttribute(*it,osg::StateAttribute::TEXTURE) );
-            if(tex!=NULL)
-                tex->dirtyTextureParameters();
+    {
+           osg::BindImageTexture*imb = 0;
+           const osg::StateSet::AttributeList & texlist= stateset->getTextureAttributeList()[*it];
```
mp3butcher: this dirty part is because ss.getTexAttribute(tu,type) fails : member searched is always set to 0 !?

mp3butcher:
Okay!..., sorry,  I unconsciously put bad will on it.. A little more thinking would have surely lead me to this impl...
I would make a pr to fix examples but an other one is previously required to fix:
a strange behavior of stateset.getTextureAttribute(tu,type) : member searched is always set to 0 !?
Does simply add a 3rd (uint member) param with default value seams enough or do i have to add an other method ?

PR #477, opened 2018-02-14
avoid visual studio warning about gethostbyname being deprecated.

LaurensVoerman:
avoid visual studio warning about gethostbyname being deprecated by #define _WINSOCK_DEPRECATED_NO_WARNINGS

I guess gethostbyname will still be supported by windows for a while.

### DIFF
--- a/src/osgPlugins/osc/ip/win32/NetworkingUtils.cpp
+++ b/src/osgPlugins/osc/ip/win32/NetworkingUtils.cpp
@@ -28,7 +28,7 @@
 	WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 #include "ip/NetworkingUtils.h"
-
+#define _WINSOCK_DEPRECATED_NO_WARNINGS
 #include <winsock2.h>   // this must come first to prevent errors with MSVC7
 #include <windows.h>
 #include <stdlib.h>


src/osgPlugins/osc/ip/win32/NetworkingUtils.cpp
```diff
@@ -28,7 +28,7 @@
 	WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 #include "ip/NetworkingUtils.h"
-
+#define _WINSOCK_DEPRECATED_NO_WARNINGS
```
filnet: Would it be possible to do that from the CMakeLists.txt file ?

LaurensVoerman:
That would be something like:
  IF(WIN32)
    ADD_DEFINITIONS(-D_WINSOCK_DEPRECATED_NO_WARNINGS)
  END IF(WIN32)
In this case I chose to add it in the source to get the warning back in
case someone decided to update the osc code.
Laurens.

openscenegraph:
Do you know what the new Windows replacement function would be?

mp3butcher:
int WSAAPI getaddrinfo(
  _In_opt_       PCSTR      pNodeName,
  _In_opt_       PCSTR      pServiceName,
  _In_opt_ const ADDRINFOA  *pHints,
  _Out_          PADDRINFOA *ppResult
);
https://msdn.microsoft.com/en-us/library/windows/desktop/ms738520(v=vs.85).aspx

openscenegraph:
Thanks for the suggestion.  I have looked to getaddrinfo but as I can't compile things easily under Windows and don't have a test case to test any changes out I'll pass on trying to convert the code across.  Long term this will be the way to go.  It's a niche plugin so we can tackle this when need be.

As an interim fix what I have done is checked in something similar the suggestion change to the CMakeLists.txt:

https://github.com/openscenegraph/OpenSceneGraph/commit/f5021e8bc0d6bf5327f2b97651159f223451846c

PR #478, opened 2018-02-14
provide specialization for GetScalarValue macro for bool type 

LaurensVoerman:
to avoid compiler complaints in visual studio 2015 (14)
(and older visual studio versions)

MS> This warning is no longer generated in Visual Studio 2017. 
I still think that merging this code is useful to avoid compiler warnings on a slightly older compiler, particularly because this is in a header file. 
Regards, Laurens.

### DIFF
--- a/include/osg/ValueObject
+++ b/include/osg/ValueObject
@@ -154,8 +154,28 @@ class ValueObject : public Object
             virtual void apply(float in_value) { value = in_value; set = true; }
             virtual void apply(double in_value) { value = in_value; set = true; }
         };
+#if defined(_MSC_VER)
+        template<>
+        class GetScalarValue <bool> : public ValueObject::GetValueVisitor
+        {
+        public:
 
+            GetScalarValue() : set(false), value(0) {}
 
+            bool set;
+            bool value;
+
+            virtual void apply(bool in_value) { value = in_value; set = true; }
+            virtual void apply(char in_value) { value = in_value != 0; set = true; }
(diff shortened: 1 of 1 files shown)


LaurensVoerman:
Wrapped the specialization in "#if defined(_MSC_VER)" as it's not allowed there by C++ . I have now found that the proper order would be:
----
namespace osg {
		template<typename T>
		class GetScalarValue;//forward declaration
class ValueObject : public Object {...};

        template<typename T>
        class GetScalarValue : public ValueObject::GetValueVisitor{...};
        template<>
        class GetScalarValue <bool> : public ValueObject::GetValueVisitor {...};
}
----
This would make a osg::GetScalarValue  instead of osg::ValueObject::GetScalarValue (needed to make the specialization for bool). For symmetry I suppose the SetScalarValue has to move too. 
As these things conflict with an actual bugfix submitted earlier, I'll defer that for now. 
Feel free to reject this PR, is's getting a bit messy just to fix a compile warning.
Regards, Laurens.

LaurensVoerman:
I implemented a version without the #ifdef _MSC_VER; 
submitted as https://github.com/openscenegraph/OpenSceneGraph/pull/479
I prefer that version, but I'll leave this PR open in case your preference is different.
Regards, Laurens.

openscenegraph:
Having compiler specific code paths in headers is really not desirable, it's a really painful way to code and maintain code long term and really not something I'd like to have as an example for others , so I'll close this PR and when I can will test the proposed alternative.

PR #490, opened 2018-03-01
osgDB: Fix handling of return value of loadLibrary in Registry.

torokati44:
"Unable to load ..." was printed if the library could be loaded, and it wasn't if not.

This incorrect, inverted logic was probably a copy-paste error.


openscenegraph:
Wow, can't believe this bug hadn't been spotted sooner.  I've used -e many times in testing and hadn't twigged the erroneous message.  Thanks for the fix.

PR #491, opened 2018-03-05
ReaderWriterGDAL: fix stride when reading multi-channel images with…

mmassing:
… more than 1 byte per pixel (e.g. 16bit RGB PNGs).
The problem can be easily reproduced by reading a 16bit PNG od 16bit RGB tiff file.


openscenegraph:
Thanks, I've done a review and changes look correct so I'll be merging them.  

One thing that jumped out at my was the numBytesPerPixel variable is misnamed, it really should be numBytesPerComponent so I'll rename this after merging your change.  Thanks.

PR #495, opened 2018-03-13
fix installation of PDB files from the examples.

aitormoreno:
This PR fixes the installation of the PDB files of the examples. 
More info in the [OSG mailing list](http://forum.openscenegraph.org/viewtopic.php?t=17032)


openscenegraph:
Thanks for the changes. I've just done a review of the your changes and the wider MACRO(SETUP_EXAMPLE EXAMPLE_NAME) and MACRO(SETUP_APPLICATION APPLICATION_NAME) and they are inconsistent.  Since I OSG forum/mailing list thread didn't highlight problems with the applications I presume it's down to the differences.  

I presume at some point hey would have been more similar and have diverged - my guess is that there was a fix for the pdb issue for the applications but as the examples aren't built by default problem with them weren't spotted.  

As your changes further diverge the two implementations I don't feel they are the best solution. to minimize the differences I've checked in the following commit.  Could you please test this and let me know if it works fine.  If so I'll backport this change from OSG master to the 3.4 branch.

https://github.com/openscenegraph/OpenSceneGraph/commit/57f2b1fa27671962601385167a5ccd90b85536b7

https://github.com/openscenegraph/OpenSceneGraph/commit/57f2b1fa27671962601385167a5ccd90b85536b7

aitormoreno:
The new commit in master does not find the PDB files for the applications and for the examples. The variable  `${TARGET_TARGETNAME}` expands to `application_xxxx` or `example_xxxx` and therefore the file to be installed is `bin/application_xxxd.pdb`, `bin/example_xxxd.pdb` which do not exist.

Just replacing `${TARGET_TARGETNAME}` with `${TARGET_NAME}` fixes the installation. I can update this PR if you like.

openscenegraph:
Thanks for the testing. Amending the PR would be great, that way I will know it works.  Cheers, Robert.

aitormoreno:
I pushed the new changes and sync'd with master to solve the conflict against 57f2b1f

openscenegraph:
Thanks for the updates.

PR #499, opened 2018-03-21
Fix include directory order

AMDmi3:
Include paths to external libraries should always be added with `SYSTEM` keyword, so they are always ordered after local include directories. This particular patch fixed build of osg 3.4.1 in presence of osg 3.4.0:

```
/usr/bin/c++  -Dosgdb_ply_EXPORTS -I/ssd/portstrees/batchports-mem/graphics/osg/work/OpenSceneGraph-OpenSceneGraph-3.4.1/include -I/usr/local/include -Iinclude -O2 -pipe -march=nocona -fstack-protector -fno-strict-aliasing -Wall -Wparentheses -Wno-long-long -Wno-import -pedantic -Wreturn-type -Wmissing-braces -Wunknown-pragmas -Wunused -Wno-overloaded-virtual -fPIC -MD -MT src/osgPlugins/ply/CMakeFiles/osgdb_ply.dir/vertexData.o -MF src/osgPlugins/ply/CMakeFiles/osgdb_ply.dir/vertexData.o.d -o src/osgPlugins/ply/CMakeFiles/osgdb_ply.dir/vertexData.o -c /ssd/portstrees/batchports-mem/graphics/osg/work/OpenSceneGraph-OpenSceneGraph-3.4.1/src/osgPlugins/ply/vertexData.cpp
/ssd/portstrees/batchports-mem/graphics/osg/work/OpenSceneGraph-OpenSceneGraph-3.4.1/src/osgPlugins/ply/vertexData.cpp:567:44: error: no member named 'readImageFile' in namespace 'osgDB'
        if (textureFile && (image = osgDB::readImageFile(textureFile)) != NULL)
                                    ~~~~~~~^
1 error generated.
```

- This is common problem all over osg CMake code, should I submit fixes for other parts?
- This applies to 3.4 branch and likely all later branches, please don't forget to cherry-pick, or should I submit a PR for each branch?


openscenegraph:
This issue hasn't been reported in the many years that the OSG has been out in the wild so I'm surprised you are hitting upon this issue, suggests that there is something specific and unusual about what system you are using and what the various combinations of items that have been installed are.

Could you explain enough details so that others can reproduce the issue first hand before we embark on a solution if one is required.

AMDmi3:
There's nothing specific and unusual. Include directory for external library misses SYSTEM keyword (which would otherwise make it used after project local include directories in header file lookup) and is thus placed before local includes, so includes from older installed version of osg are picked up, and the compilation fails, it's clearly visible in the log.

openscenegraph:
Could you explain how to reproduce the issue please, as I've never seen the issue myself before or seen others reporting it before.  The extra description you provided doesn't tell anything other that you multiple versions of the OSG on the go.  How do you build and install these other versions of the OSG?  What platform are you building on?

openscenegraph:
I've been experimenting with adding in SYSTEM as per PR and do a build with VERBOSE=1 and don't see any differences.   I'm using Cmake 3.5.1 under Kubuntu 16.04.

Could it be you are using a Cmake version prior to 3.0.0?  The following change suggest that modern versions of Cmake now treat imported targets as system.

https://cmake.org/gitweb?p=cmake.git;a=commitdiff;h=a63fcbcb9f6c

openscenegraph:
Can you give me specific instructions how to recreate the problem not arm wave, I'm not omnipresent. FYI, I don't have FreeBSD and am not about to install it, but I have Linux and can happily build and install multiple versions of the OSG, but can't figure out what specific combination fails from your description, there are things that you are doing that you are not telling us about.

Please give us a step by step explanation of how you build or install the various versions of the OSG and what configurations you use.

AMDmi3:
I have no idea on how to reproduce this on Linux, as I don't have it and am not about to install it either. As I understand step by step explanation for FreeBSD won't be useful for you. Why's this needed at all? It's a simple rule that external includes must always go after local ones.

AMDmi3:
> Please explain what you are doing so I can attempt to recreate it on my system

I still do not understand how that will help, after you've stated that you are not going to install FreeBSD. Still, satisfying the formal request.

Here's how a normal FreeBSD user will run into the problem:

- take stock FreeBSD 11.1 system
- install osg 3.4.0 from ports: `cd /usr/ports/graphics/osg && make install clean`
- update the ports tree `portsnap fetch update` afeter a moment osg port was updated to 3.4.1
- try to build the newer version while 3.4.0 is still installed: `cd /usr/ports/graphics/osg && make install clean`
- experience an above error

This was reported by two users and reproduced by me.

However, if you ask how to reproduce that manually, given than there's no more 3.4.0 port and 3.4.1 was fixed by the attached patch, this may be done like this:

- take stock FreeBSD 11.1 system
- get port of osg 3.4.0: `svn co svn://svn.freebsd.org/ports/head/graphics/osg@459451 osg340`
- get port of osg 3.4.1 without the fix: `svn co svn://svn.freebsd.org/ports/head/graphics/osg@464404 osg341`
- install older osg: `make -C osg340 install clean`
- try to build newer osg: `make -C osg341`
- experience an above error

openscenegraph:
Thanks for the details, provide a better idea of what you are doing at your end.  I'm trying to recreate conditions under Linux.  

So far I've build and installed 3.4.0  to /usr/local and then built 3.4.1 from source but it worked without any conflicts.  I then built and installed 3.4.0 to /usr and then built 3.4.1 from source but again it worked without any conflicts.

I have now built and installed cmake 3.10.3 to see if this has a influence (Kubuntu 16.04 repository pulls cmake-3.5.1 that that's what I was previously testing.)  So far the build is going fine (it's already built osgDB).  When I run make VERBOSE=1 I don't get a build line that has all the includes that you build error line has, so I suspect this might be the difference.  

The error line you mentioned in your first line has a "-I/usr/local/include -Iinclude" after the -I entry for the local OSG headers, these don't occur for my build.  Could this be the reason why your builds are failing while when I do essentially the same thing I don't get any errors.

Doe these additional "-I/usr/local/include -Iinclude" change when you use the SYSTEM argument as per your PR?  Could you do a make VERBOSE=1 to see what is going on w.r.t the compile line.   When I added the SYSTEM usage I didn't see any difference in command lines being used for build (that was with 3.5.1 though.)

The reason important for me to try and establish what is going is what the consequences will be for all platforms if we do introduce the SYSTEM usage and just how many CMake files we'll need to modify.  It's not been reported that we need to use this before so adding after all these years is something that is really has to prove itself as necessary and without any consequences.

A bit unrelated, but there a reason svn is being used over pulling down the official OSG git repository?  How up to date is this svn repository?

openscenegraph:
I have done quite a bit reading and testing on my system with different build combinations and so far haven't come across any changes do to adding SYSTEM usage when include the GL headers, but then the GL headers don't even appear on my system.  I haven't yet figured out why CMake is not adding these in, perhaps it's a platform specific issue.

As I'm not seeing any issues with using SYSTEM and appears to help on FreeBSD I've checked in an changes - this are broadly similar to your ones,I just moved the INCLUDE_DIRECTORIES to after the out of source build include so make the CMake script ready more naturally as both OSG source related includes are kept next to each other.  I have merged this change with OSG master, and the 3.4 and 3.6 branches.  

I also upped the CMake required version for 3.4 branch to 2.8.0 to make sure it's safe to use SYSTEM.

The commit is:
https://github.com/openscenegraph/OpenSceneGraph/commit/68ba89c6dc2bae2df2b14e589f13db3817204ded

I will now close this PR as it shouldn't be required, whether we need to extend this style usage out further is something for other PR's, I would rather discussions of the issues on osg-users mailing list/forum so that others in the community can chip in, github comments is poor way of getting feedback from the wider community.

PR #501, opened 2018-03-22
Restore PBO for GLES2, which was removed for no reason

kornerr:
The "GLExtension fix" PR ( https://github.com/openscenegraph/OpenSceneGraph/pull/485 ) broke web builds ( https://github.com/OGStudio/openscenegraph-cross-platform-guide/tree/master/1.10.SampleWeb ).
Moreover:
* neither "GLExtension fix" PR, nor osg-users mailing list contained ANY description
* the change was INCONSISTENT because the checks before and after it were still using the removed GLES2_FEATURES

My PR restores PBO for GLES2.


openscenegraph:
When I reviewed the #485 PR I checked online for whether PBO is supported by default under GLES2. PBO support was added in GLES3.  Even if it turns out that it isn't complete, it sure looks like it's appropriate so I don't think it's appropriate to revert it.

Could you raise the topic for discussion on the osg-users mailing list/forum so we can properly discuss the issues you've come across so we can see what is appropriate to do to resolve them.

PR #504, opened 2018-03-25
Lxh dev

llduncangithub:


openscenegraph:
Absolutely crazy PR.  Is this a dumb attempt to hack the OSG?  If you have a valid modification to the OSG please just submit the key changes that you have with a full explanation of what those changes are and why you feel they are required.

PR #506, opened 2018-03-26
Open scene graph 3.6 text

valid-ptr:


valid-ptr:
Should i create the same PR for the master?

openscenegraph:
The VertexArrayState change looks sound.

The TextBase alignment looks like an issue that needs to be look into it, could you post an explanation of how to reproduce the issue on osg-users/forum and I can look into it,

The OSG_GL_FIXED_FUNCTION change is buggy.  I understand the motivation but this ain't the way to do it as it will could GL/GLES versions that don't support glEnable.glDisable with GL_TEXTURE_2D.

Could you generate separate PR's for each class of fix that make, this will allow me to properly review and merge the ones that can be merged, rather than just rejecting the all because a subset of the changes aren't appropriate.  Thanks.

W.r.t keeping master up to date, I can cherry-pick commits easily so no need to worry about this - fixes made to master I merge with the 3.6 branch and visa versa.

PR #507, opened 2018-03-26
OSG 3.6: Write support with .curl pseudo extension added

valid-ptr:


jtorresfabra:
Hey @valid-ptr and @robertosfield , 
This is a comment about the form, not the content. 

It would be nice if we start to get the git history cleaner. Having 3 commit s for a 3 lines change should be not ok. Moreover when 2 of the commits are probably not necessary. Rebasing (and fixing up or squashing commits) is pretty easy, and it would be nice to have this in mind. 

Thanks!

PR #512, opened 2018-03-28
don't lose geometry during optimizer's merge pass

aumuell:
We had problems where parts of our geometry was lost during optimization. The geometry was split into batches of ~10K elements and organized into multiple Geometry objects as children of a single Geode.

### DIFF
--- a/src/osgUtil/Optimizer.cpp
+++ b/src/osgUtil/Optimizer.cpp
@@ -1956,7 +1956,7 @@ bool Optimizer::MergeGeometryVisitor::mergeGroup(osg::Group& group)
                 ++mitr)
             {
                 DuplicateList& duplicateList = *mitr;
-                if (duplicateList.size()>1)
+                if (!duplicateList.empty())
                 {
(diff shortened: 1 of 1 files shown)


openscenegraph:
The change suggest to me that a duplicateList of size 1 is being created, but for if so then there is nothing to merge so this code block shouldn't be reached at all, suggestion a bug earlier in the code.  Merging this "fix" would just hide the actual bug.

My guess is the code looking for geometries to merge is mis-classifying the some of the geometries you have, spotting this type of subtle bug is hard to do by just looking at the code, the most helpful thing would be to have a test dataset that can illustrate the bug, then we can use this a unit test to hunt down the actual problem code and then prove that any fixes to the code fix it.

Could you post some data on the osg-users mailing list/forum so we can look into it further, or post it on the OpenSceneGrpah github Issue tracker.  Thanks.

Could you

aumuell:
Unfortunately, I did not get the ok to upload the data which caused the problems.

But thank you: 3.6.4-rc4 works ok now.

PR #513, opened 2018-03-30
Remove texture attribute member override

mp3butcher:
base issue:
StateSet::removeTextureAttribute doesn't work for tu other than 0
fix:
texture attribute member should be 0 to be retrieved correctly in StateSet methods
I'm not sure of it but there can't be several texatt on the same tu at a time..no?
Further I can't be agree on the TextureAttribute::_textureunit member as a single Tex can be shared on differnent texture unit among statesets (so having tu member of textture is really a BAD thing)...but perhaps there a pipeline conformance reason (and but it sure shouldn't be exposed as getmember)

### DIFF
--- a/include/osg/TextureAttribute
+++ b/include/osg/TextureAttribute
@@ -32,8 +32,6 @@ class TextureAttribute : public StateAttribute
 
         virtual bool isTextureAttribute() const { return true; }
 
-        virtual unsigned int getMember() const { return _textureUnit; }
-
     protected:
         virtual ~TextureAttribute() {}
 


mp3butcher:
Oups Sorry it's late in fact i think i'm right

openscenegraph:
This would break the shader pipeline support as there wouldn't be any mechanism for properly setting where the uniforms should be assigned.  Fixing one bug but introducing a whole set of new ones is not acceptable so the only sensible route is to close this PR.  

Once 3.6.0 is out the door I'm happy to look any bug with master and can work what should be done.

PR #515, opened 2018-04-01
AutoTransform must never cache its bounding sphere when auto-scaling …

hwiesmann:
…its children to screen

When auto-scaling its children to screen the bounding sphere changes by each scene modification (in the extreme case it changes for each frame). The bounding sphere is only newly calculated when it or one of its children have dirty bounds. But this actually means that mostly the cached bounding sphere does not correspond with the real bounding sphere.
This can cause problems during culling: the AutoTransform node is culled because of a wrong (not updated) bounding sphere. A demonstration of this phenomenon is shown in a modified osgautotransform.cpp file as mentioned on http://forum.openscenegraph.org/viewtopic.php?t=17173&highlight=autotransform+culling.

To solve the issue AutoTransform::computeBound always returns an invalid bounding sphere when the flag "_autoScaleToScreen" is set.

### DIFF
--- a/include/osg/AutoTransform
+++ b/include/osg/AutoTransform
@@ -125,7 +125,6 @@ class OSG_EXPORT AutoTransform : public Transform
 
         osg::Matrixd computeMatrix(const osg::NodeVisitor* nv) const;
 
-        mutable bool                    _matrixInitalized;
         mutable osg::Matrixd            _cachedMatrix;
 
         enum AxisAligned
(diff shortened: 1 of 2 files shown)


openscenegraph:
A quick update.  I've spent most of the afternoon looking into this issue and testing your changes, and haven't yet decided on what the best route forward is yet.  I have to be careful that I don't merge something that fixes one users problem but inadvertently break it for other usage cases.  I will spend some more time later today thinking about the issue.

PR #520, opened 2018-04-04
Remove duplicated and incorrect fetching of 'map_bump' attributes in obj plugin

rhabacker:
Parsing key 'map_bump' was processed in the block, where the attributes
for 'bump' are extracted and results into having parts of the key in the
extracted filename, generating an invalid filename.

The mentioned string compare could be removed without loosing any features,
because the key 'map_bump' is parsed correctly some lines below.


rhabacker:
What is the strategy to get this into OpenSceneGraph-3.6 branch ? Are you cherry-picking this commit or should I file an additional pull request ?

PR #522, opened 2018-04-05
Log level changed from INFO to DEBUG in DisplaySettings::setShaderHin…

valid-ptr:
…t function

After applying this patch
osgviewer --major-version
will return 3, not something like "SHADER_GL3\n3"

PS
OSG_NOTIFY_LEVEL=FATAL osgversion --major-version is not very friendly ;)


openscenegraph:
Changing to INFO is more appropriate so I've checked this in so will close this PR.

PR #523, opened 2018-04-05
Text (2D) alignment fix

valid-ptr:


valid-ptr:
NOTE: _alignment variable used in osg::Text::computeGlyphRepresentation impl.

openscenegraph:
Could you please provide information on what the problem is and how to reproduce it as I would like to investigate the issue properly.

valid-ptr:
I've changed osgtext3D example. Changes in TextBase fixes dynamic change of "right alignment" in 2D and 3D texts.

I've found some difference in 2d/3d visualization.
Run: **./osgtext3D --2d**
Press left-key once,
Hit Enter (return) key (and you will see the difference in the "right alignment" without TextBase patch)
Press up-key.

Positions on 3d and 2d texts are not the same...
Maybe it's ok!?

robertosfield:
Thanks for the mods to the example.  I'll have to check these changes out as branch and see what is going on, at this point it's too much for me to be able to do a simple test and know exactly what is going on.  It's unfortunate that it's so close to 3.6.0 as it be nicer to have a bit more time to be able to investigate things, get any changes checked in so that others can test.  I will have a think about how best to juggle things.

openscenegraph:
I have investigated the TextBase::setAligment() bug and using suggestions created a focused unit test that I've added to the osgtext example, so now you can run ostext --alignment

The test is in the commit:
  https://github.com/openscenegraph/OpenSceneGraph/commit/d1c54223cbe11aa21e44568155adffff065f3c19

Without fixing TextBase::setAlignment() two text string have different alignment of the two line strings, when applying the suggested fix the osgtext --alignment test then justifies the text consistently.  I've checked this in as the commit:
https://github.com/openscenegraph/OpenSceneGraph/commit/36f891c6cd7fbed5185530573818ff7e8aa17da9

openscenegraph:
With these fixes I'm going to close this PR as the key fix is now part of OSG master and the 3.6 branch.  Thanks for your help.

PR #526, opened 2018-04-10
Update View.cpp

beachwalker:
I changed the cast from
const osgViewer::GraphicsWindow* gw = dynamic_cast<const osgViewer::GraphicsWindow*>(eventState->getGraphicsContext());
to 
const osg::GraphicsContext* gw = dynamic_cast<const osg::GraphicsContext*>(eventState->getGraphicsContext());
because only GraphicsContext is needed here.

### DIFF
--- a/src/osgViewer/View.cpp
+++ b/src/osgViewer/View.cpp
@@ -887,7 +887,7 @@ bool View::containsCamera(const osg::Camera* camera) const
 const osg::Camera* View::getCameraContainingPosition(float x, float y, float& local_x, float& local_y) const
 {
     const osgGA::GUIEventAdapter* eventState = getEventQueue()->getCurrentEventState();
-    const osgViewer::GraphicsWindow* gw = dynamic_cast<const osgViewer::GraphicsWindow*>(eventState->getGraphicsContext());
+    const osg::GraphicsContext* gw = dynamic_cast<const osg::GraphicsContext*>(eventState->getGraphicsContext())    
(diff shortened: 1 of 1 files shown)


src/osgViewer/View.cpp
```diff
@@ -887,7 +887,7 @@ bool View::containsCamera(const osg::Camera* camera) const
 const osg::Camera* View::getCameraContainingPosition(float x, float y, float& local_x, float& local_y) const
 {
     const osgGA::GUIEventAdapter* eventState = getEventQueue()->getCurrentEventState();
-    const osgViewer::GraphicsWindow* gw = dynamic_cast<const osgViewer::GraphicsWindow*>(eventState->getGraphicsContext());
+    const osg::GraphicsContext* gw = dynamic_cast<const osg::GraphicsContext*>(eventState->getGraphicsContext())
```
filnet: missing ;

openscenegraph:
You really need to compile and test you PR's being you considering submitting.  This is a bare minimum.

Also using a dynamic_cast<GraphicsContext*> is. rather redundant for a pointer than is already an osg::GraphicsContext*.

I have done this in the commit:
   https://github.com/openscenegraph/OpenSceneGraph/commit/a58f6c68d6bfada2fc4bd860abe86c81a857c1a1

As I have properly implemented the intended change I'm closing this attempt at a PR.

src/osgViewer/View.cpp
beachwalker: yeah, sorry... did this on the go.

acidrainstudios:
Did you mean gc instead of gw for the variable name?

PR #527, opened 2018-04-10
Add 'dump scene to file' handler

rhabacker:
This handler is for debugging purpose.

### DIFF
--- a/applications/osgviewer/osgviewer.cpp
+++ b/applications/osgviewer/osgviewer.cpp
@@ -19,6 +19,7 @@
 
 #include <osgViewer/Viewer>
 #include <osgViewer/ViewerEventHandlers>
+#include <osgViewer/DumpSceneHandler>
 
 #include <osgGA/TrackballManipulator>
 #include <osgGA/FlightManipulator>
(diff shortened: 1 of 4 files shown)


rhabacker:
Should I better recreate a pull request to master branch ?

rhabacker:
> It's a very specific type of handler for a specific style of debugging. As an OSG example it might be fine, but I don't feel it's right for a public library like osgViewer.

This plugin has been designed similar to RecordCameraPathHandler, which save the currently recorded animation to a file with a predefined filename into the current directory and the ScreenCaptureHandler, which saves a dump of the displayed screen into a file named 'screen_shot.jpg' in the current working directory. From my opinion these plugins support also a specific style of debugging and they has been added to osgviewer. 

If the problem is caused by using the temp dir and not the working dir, I'm happy to refactor the plugin to follow that style.  Also I saw that RecordCameraPathHandler uses a different api to enable file name autoincrement  void setAutoIncrementFilename( bool autoinc = true ), which may be a good pattern to use also.

> As a general note, new features should be done against master as OSG-3.6 is now in feature freeze.
thanks for this pointer.

PR #529, opened 2018-04-16
Add 'dump scene to file' handler

rhabacker:
Refactored to make it similar to RecordCameraPathHandler, which has be added to osg viewer

### DIFF
--- a/applications/osgviewer/osgviewer.cpp
+++ b/applications/osgviewer/osgviewer.cpp
@@ -19,6 +19,7 @@
 
 #include <osgViewer/Viewer>
 #include <osgViewer/ViewerEventHandlers>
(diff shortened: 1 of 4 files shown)


openscenegraph:
Hi Ralf, I don't think this refactor of the PR addresses the fundamental issue, I feel it's just too niche a feature to warrant adding to core osgViewer library.  I realize there are already some quite niche the core libraries, and in hindsight I probably wouldn't include all of this, but I can't revert these without breaking the build for some users.  However, integrating even more such examples of niche functionality just takes us further down the path of a bloated API.

rhabacker:
I see - in the initial pr you mentioned an example to which this could be added - which example are you refering too ?

PR #531, opened 2018-04-16
Including osg/Geode to fix osgManipulator build

jasonbeverage:
Hi Robert!

Just a small change to fix the osgManipulator build.  TrackballDragger needs <osg/Geode> included since IntersectVisitor was removed.

Thanks!

Jason


openscenegraph:
Ouch, looks like I added the include<osg/Geode> to the .cpp rather than then header. I'll need to review the other heads as well.

PR #533, opened 2018-04-20
Fix typos and spelling.

aluaces:
I realize that this is a big one, so I don't expect it to be reviewed quickly :)


include/osgText/Text
```diff
@@ -73,7 +73,7 @@ public:
      * placing text against noisy backgrounds.
      * The color of the background shadow text is specified by setBackdropColor().
      * DROP_SHADOW_BOTTOM_RIGHT will draw backdrop text to the right and down of
-     * the normal text. Other DROW_SHADOW_* modes do the same for their repective directions.
+     * the normal text. Other DROW_SHADOW_* modes do the same for their respective directions.
```
filnet: DROW ?

aluaces: It was written that way before, but nevertheless I will correct it, thanks.

openscenegraph:
Wow, that's a lot.  Just run through them, all look appropriate.  Amazing thing is that no all the typos/spelling errors were mine!!! Might be a first :-)

PR #535, opened 2018-04-24
add wrapper for osg::DrawIndirectBufferObject

mp3butcher:
a quick pr to add wrapper for osg::DrawIndirectBufferObject (reject it if you think it would deserve a file for itself)


openscenegraph:
I have just done a quick review, and I think it would better as a separate file.  I have pulled in the changes into a separate file myself and merged this with OSG-3.6 and master.

PR #536, opened 2018-04-24
When GL3 build is enabled, default context requested is version 3.3, …

emminizer:
…enabling core profile by default.


emminizer:
Note that Mac versions of Graphics Window already force something similar.  This does change behavior in Windows / Linux GL3 builds of OSG such that now core profiles are selected by default.  Mac OS Graphics Window appears hard-coded to already do so.

openscenegraph:
After merging this change I had a change of heart, I don't feel that having an #ifdef spread across two different .cpp files is a good practice for long term maintenance, a better approach will be to have a single #define that both these .cpp uses for the version definition.  I will look into implementing this now.

emminizer:
Looks good, I think a CMake solution to this is definitely sufficient.  Thanks.

PR #554, opened 2018-06-01
Geometry: Add VBO for geometry arrays in dirtyGLObjects if required. …

emminizer:
… Prevents bug with late call to Array::setBinding().

### DIFF
--- a/src/osg/Geometry.cpp
+++ b/src/osg/Geometry.cpp
@@ -681,6 +681,15 @@ void Geometry::setUseVertexBufferObjects(bool flag)
 
 void Geometry::dirtyGLObjects()
 {
+    if (_colorArray.valid() && _colorArray->getBinding() == osg::Array::BIND_PER_VERTEX)
+        addVertexBufferObjectIfRequired(_colorArray.get());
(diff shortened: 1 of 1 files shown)


openscenegraph:
I have done a first pass review and this really doesn't look like the right way to fix the issue what you have.  During the work toward 3.6.1 I did resolve a bug related to VBO's and setBinding() so there is chance this this issue has already been fixed.  However, as I don't know anything specific about the problem you have seen and which version you were using there is no way for me to comment on this.

Could you please write a post to osg-users mailing list/forum outlining which version of the OSG you have seen any issue with and how we can recreate it.  if you haven't already done so please test first on 3.6.1.

Cheers,
Robert.

emminizer:
Hi Robert,

This was against OpenSceneGraph-3.6, and affected 3.6.1.  The error was impacting my clean 3.6.1 build.

I agree that I'm not happy with the solution, but it does solve the crash, and was primarily offering it as "here's something that does work".

I sent the details on the mailing list because I didn't think this was the correct forum on which to do so. 
 I'll work on demonstrating via an example.

 - Dan

PR #555, opened 2018-06-14
FLT: Missing optional attr files on textures no longer generates a co…

emminizer:
…nsole warning.

### DIFF
--- a/src/osgPlugins/OpenFlight/PaletteRecords.cpp
+++ b/src/osgPlugins/OpenFlight/PaletteRecords.cpp
@@ -344,8 +344,13 @@ protected:
         texture->setImage(image.get());
         stateset->setTextureAttributeAndModes(0, texture, osg::StateAttribute::ON);
 
-        // Read attribute file
+        // Since the .attr file is optional according to the OpenFlight spec, check to see
+        // if the file exists before reading it, to avoid printing an unnecessary warning.
         std::string attrname = filename + ".attr";
+        if (!osgDB::fileExists(attrname))
+            return stateset;
+
+        // Read optional attribute file
         osg::ref_ptr<AttrData> attr = osgDB::readRefFile<AttrData>(attrname,document.getOptions());
         if (attr.valid())
         {


emminizer:
Hi Robert,

This is the text issue I mentioned on the mailing list.

According to the OpenFlight 1.0 (beta) spec I found online, the .attr files are potentially optional in display applications.

In OSG 3.4, missing .attr files did not warn because `osgDB::readRefObjectFile()` only warned on `ReadResult.error()`.  In 3.6, it warns on `!rr.success()`.  Missing file is not a success, but it's also not an error.  Because of this minor (but good) change, I'm seeing some extra console spam.

Because the file is optional, I'm skipping the `readRefFile()` call if I know it will fail due to a missing file.

Thanks,

 - Dan

LaurensVoerman:
I think the spec you refer to is somewhat outdated (1995). The most recent
version you can find here
<https://www.presagis.com/workspace/uploads/files/openflight16-6.pdf> (OpenFLT
16.6 - 2016).
While I do not object to your changes, I do not see how you read the spec
to say missing .attr files is not an error.

" These attribute files are used by the modeling software, and may not be
necessary for the application using the database."

missing .attr files used to (~15 years ago) be rewarded with crappy texture
filtering - do we use a reasonable default now?

Regards, Laurens.

emminizer:
Hi Laurens,

Thanks for the more up to date spec link.  It sounds like you've got a long term experience with OpenFlight and the OSG plugin here, I like hearing your input.

The text you quoted is exactly why I don't read it as an error.  It specifically calls out the possibility of an absence of those .attr files and says they might not be necessary for a particular use case.

The current defaults for filtering are in `Texture.cpp`, defaulting to `LINEAR_MIPMAP_LINEAR` for `_min_filter` and `_mag_filter` set to `LINEAR`.  We have many models that don't have .attr files and haven't noticed any significant filtering issues with them, though we load models for real-time simulation more than modeling purposes, so crummy filtering wouldn't be as obvious for us.

Thanks,

 - Dan

LaurensVoerman:
Hi Dan,

Checking the code reveals that a missing .attr results in 
repeat modes (REPEAT,REPEAT,CLAMP)
min/mag filters (LINEAR_MIPMAP_LINEAR, LINEAR)
NO osg::TexEnv (this cannot be achieved with an .attr file)

Laurens.

PR #562, opened 2018-06-21
fix the bug introduced in SHA-1: 94891778c40ff08fcc626f91ded73cda67713b36 (12/08/2016 - 19:44:38)

mtola:
Robert Osfield knows the problem since few months (we exchange many mails)

### DIFF
--- a/src/osg/Geometry.cpp
+++ b/src/osg/Geometry.cpp
@@ -535,7 +535,7 @@ bool Geometry::getDrawElementsList(DrawElementsList& drawElementsList) const
 
 void Geometry::addVertexBufferObjectIfRequired(osg::Array* array)
 {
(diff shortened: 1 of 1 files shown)


openscenegraph:
This isn't a bug fix, it's hack that fixes things for you but will break things for others.  The changes that you are reverting are required to support VAO, so there is no way I'm going to break this.

Could you provide a specific link to how to recreate the problems you are seeing, the best way to do this is on the osg-users ML/forum. We can then see if it can be fixed properly for the up coming 3.6.2

mtola:
Ok, I understand.

The project is rather big and for the moment private.
First public version will be end of July or maybe end of September...

Difficult for me for the moment to isolate things... So I don't know what I can do...

PR #565, opened 2018-06-25
PointSprite: Disable the GL_POINT_SPRITE_ARB mode in GL3 -- it is alw…

emminizer:
…ays on, and the mode is deprecated (removed in core profile).

### DIFF
--- a/src/osg/PointSprite.cpp
+++ b/src/osg/PointSprite.cpp
@@ -43,7 +43,11 @@ int PointSprite::compare(const StateAttribute& sa) const
 bool PointSprite::checkValidityOfAssociatedModes(osg::State& state) const
 {
     const GLExtensions* extensions = state.get<GLExtensions>();
+#if defined( OSG_GL3_AVAILABLE )
+    bool modeValid = false;
+#else
     bool modeValid = extensions->isPointSpriteSupported;
+#endif
 
 #if defined( OSG_GLES1_AVAILABLE ) //point sprites don't exist on es 2.0
     state.setModeValidity(GL_POINT_SPRITE_OES, modeValid);


openscenegraph:
I have just had a quick look at the change.  I would have thought changing the initialization of GLExtensions::isPointSpriteSupported would be the more appropriate way to address this.

emminizer:
I can see that point of view, and it would certainly work with the code as written for GL3.  It might be slightly confusing though, because point sprite is supported; it's just forced on and can't be turned off with a mode flag.

So which solution you choose is up to your own semantics -- whether `isPointSpriteSupported` means GL_POINT_SPRITE_ARB is valid, or whether the point sprite capability is supported.

I'm good with either solution.  Do you want me to update the code, retest, and update the PR with your suggestion?

openscenegraph:
If there are two separate features that can be supported/unsuppprted then there should be two different bools in GLExtensions, I haven't looked into the different combinations yet so can't suggest exactly what should be done, but this is where I think we should be heading.  Having different places decided whether a feature is supported is not ideal, so rather than add to this we should figure out the right way to do it and implement this.  Though... haven't yet worked out exactly what this is :-)

emminizer:
No problem.  I do think it's important for 3.6.2, primarily because I don't have a solid way to work around this issue in my own code without patching OSG directly.  I know you were hoping to get 3.6.2 out the door soon though too.

There are already two booleans for point sprite in GLExtensions -- `isPointSpriteSupported` and `isPointSpriteCoordOriginSupported`.  This would be a third, hypothetically, `isPointSpriteModelSupported`.  

I don't have any strong preferences so I'm happy to defer to what you think is best.  And if this doesn't make it into 3.6.2 we can always patch OSG locally here, so I guess it's not a big deal.

Thanks for your attention on this issue.

openscenegraph:
We can get the fix in for 3.6.2.  I'd much rather than people are using a stable release for their serious application development than their own local branches where you might end up getting out of sync with the mainline.

Added GLExtensions::sPointSpriteModelSupported is reasonable, I haven't looked up the official GL extension docs for it recently so can't provide a definitive answer.  I will have a look it later.

openscenegraph:
I have added a GLExtensions::isPointSpriteModeSupported:

https://github.com/openscenegraph/OpenSceneGraph/commit/d01edcb540a4d8c87c14647067ad51c33084e4cb

It's definition is simply:

    isPointSpriteModeSupported = isPointSpriteModeSupported && !OSG_GL3_FEATURES;

Could you let me know if this works fine for you.

emminizer:
Just did a quick test.  Looks good on GL3 OSG.  No warnings out of Point Sprite and the functionality is still working.  Thanks.

PR #567, opened 2018-06-27
Remove call to nonexistent member seekpos() of std::fpos in VS 2017

LarryIII:
Hello, there.

OSG failed with below error message in VS15.8:
error C2039: 'seekpos': is not a member of 'std::fpos<_Mbstatet>'

See http://eel.is/c++draft/fpos -- to get to an offset you can convert to int; now there is no seekpos member in VS15.8.
We have fixed this issue in vcpkg (https://github.com/Microsoft/vcpkg/pull/3753).

Thanks,
Larry

### DIFF
--- a/src/osgPlugins/osga/OSGA_Archive.cpp
+++ b/src/osgPlugins/osga/OSGA_Archive.cpp
@@ -70,10 +70,12 @@ inline std::streampos STREAM_POS( const OSGA_Archive::pos_type pos )
 
 inline OSGA_Archive::pos_type ARCHIVE_POS( const std::streampos & pos )
 {
-#if defined(_CPPLIB_VER)//newer Dinkumware(eg: one included with VC++ 2003,2005)
-    fpos_t position = pos.seekpos();
+#if (defined(_CPPLIB_VER) && defined(_MSC_VER) && _MSC_VER <= 1900)//newer Dinkumware(eg: one included with VC++ 2003,2005)
+	fpos_t position = pos.seekpos();
+#elif (defined(_CPPLIB_VER) && defined(_MSC_VER))   // VC++ 2017 or latest
+	fpos_t position = pos;
 #else // older Dinkumware (eg: one included in Win Server 2003 Platform SDK )
-    fpos_t position = pos.get_fpos_t();
+	fpos_t position = pos.get_fpos_t();
 #endif
     std::streamoff offset = pos.operator std::streamoff( ) - _FPOSOFF( position );
 
--- a/src/osgPlugins/osga/OSGA_Archive.cpp
+++ b/src/osgPlugins/osga/OSGA_Archive.cpp
@@ -70,12 +70,10 @@ inline std::streampos STREAM_POS( const OSGA_Archive::pos_type pos )
 
 inline OSGA_Archive::pos_type ARCHIVE_POS( const std::streampos & pos )
 {
-#if (defined(_CPPLIB_VER) && defined(_MSC_VER) && _MSC_VER <= 1900)//newer Dinkumware(eg: one included with VC++ 2003,2005)
-	fpos_t position = pos.seekpos();
-#elif (defined(_CPPLIB_VER) && defined(_MSC_VER))   // VC++ 2017 or latest
-	fpos_t position = pos;
+#if defined(_CPPLIB_VER)//newer Dinkumware(eg: one included with VC++ 2003,2005)
+    fpos_t position = pos.seekpos();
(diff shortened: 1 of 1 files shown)


openscenegraph:
I am surprised that no other OSG users have reported this problem yet, we've done quite a few calls for testing with recent releases with active testing under Windows.

What specific compilers and build tools are you using when you see this error?

I am reluctant to merge a change that complicates already overly complicated workaround code, so really want to make sure it's required.

emminizer:
> I am surprised that no other OSG users have reported this problem yet, we've done quite a few calls for testing with recent releases with active testing under Windows.

My main machine for testing these releases is MSVC 2015.  I also test 2010, 2012, and 2013, though not always against the release candidates.  I do not currently build MSVC 2017 (15.8), so I wouldn't have seen this in any of my testing.

bjornblissing:
I have just compiled the 3.6.2-rc2 using Visual Studio 2017 version 15.7.4 and the `seekpos()` is still available in this version. So this must have been something that changed in the preview versions of Visual Studio. I have browsed the release notes for 15.8 preview 1, 2 & 3, but I fail to find any notice about this.
https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-preview-relnotes

But it might be related to Microsoft's new deal with trying to be fully standard conformant.

LaurensVoerman:
Visual Studio 2017 version 15.8 is a preview version,
I'm using 15.7.3 - that still compiles fine.
Laurens.

openscenegraph:
Thanks for the feedback.  So a preview version? Does this mean that it could be part of the next VS release like a VS 2018 or just a point release of VS2017?

If this is part of a MS effort to be more standards compliant then could we just change the code to be the same as use for other platforms rather than having all these workarounds - I'm expecting to have all the old workarounds still required, but would rather avoid yet another layer of MS VS workarounds.

bjornblissing:
This is a point release of 2017. Usually the preview version take about a month or two to propagate to the stable version. So this issue will become apparent for all Visual Studio 2017 users quite soon.

LarryIII:
Hi, sorry for the misunderstand about VS15.8, in fact, we are testing OSG in an upcoming VS 2017 release. So for this PR, you can consider it is a future PR that will unblock customer. 
I used _MSC_VER to check if customer use VS 2017 or previous VS. Moreover, I think if you merge this PR now, it will unblock anything because below code definaton: 
#if (defined(_CPPLIB_VER) && defined(_MSC_VER) && _MSC_VER <= 1900)//newer Dinkumware(eg: one included with VC++ 2003,2005)
	fpos_t position = **pos.seekpos();**
#elif (defined(_CPPLIB_VER) && defined(_MSC_VER))   // VC++ 2017 or latest
	fpos_t position = **pos;**

bjornblissing:
@LarryIII But shouldn't the `_MSC_VER` version wait to trigger until it is larger than the current working compiler version, i.e. 1914?

Such that the code should be written as:

```
#if (defined(_CPPLIB_VER) && defined(_MSC_VER) && _MSC_VER > 1914)   // VC++ 2017 version 15.8 or later
	fpos_t position = pos;
#elif (defined(_CPPLIB_VER) && defined(_MSC_VER)) // Dinkumware (eg: one included with VC++ 2003, 2005...)
	fpos_t position = pos.seekpos();
#else // older Dinkumware (eg: one included in Win Server 2003 Platform SDK )
	fpos_t position = pos.get_fpos_t();
#endif
```

openscenegraph:
Thanks @bjornblissing, your suggested code looks more natural ordering, with the latest variant coming first, then the older MS variant then the fall-back.  

@LarryIII could you modify your changes along these lines, I can then merge the changes and gets some testing of this, then I'll make a 3.6.2-rc3.  Cheers.

alexkaratarakis:
[note: Larry & I are on the C++ team in Microsoft]
`osg` is in [vcpkg](https://github.com/Microsoft/vcpkg) and we run daily builds for the entire `vcpkg` catalog with our latest compiler daily builds. 

This particular change is indeed a conformance change; the C++ spec does not specify a `seekpos()` function so it is getting removed in 15.8. This is also why it hasn't shown up for users yet.

We will amend this PR with @bjornblissing's suggestion and update to 3.6.2-rc3 in `vcpkg`. :)

LarryIII:
Closed this one! I have sent a new PR in: https://github.com/openscenegraph/OpenSceneGraph/pull/569
Please help merge. And we will update to 3.6.2-rc3 in vcpkg. Thanks all!

PR #568, opened 2018-06-27
OpenFlight: Textures now correctly export in GLCORE mode.

emminizer:
I'm about to send a mailing list post about this change and why it's needed.  In short, check out osg::Texture::getModeUsage().  Because it has `ifdef OSG_GL_FIXED_FUNCTION_AVAILABLE` around it, the `GL_TEXTURE_2D` mode never gets set `ON`, it's kept at `INHERIT`.  That means `isTextured()` always returned false.

### DIFF
--- a/src/osgPlugins/OpenFlight/expGeometryRecords.cpp
+++ b/src/osgPlugins/OpenFlight/expGeometryRecords.cpp
@@ -63,7 +63,12 @@ bool
 FltExportVisitor::isTextured( int unit, const osg::Geometry& geom ) const
 {
     const osg::StateSet* ss = getCurrentStateSet();
+#ifdef OSG_GL_FIXED_FUNCTION_AVAILABLE
     bool texOn( ss->getTextureMode( unit, GL_TEXTURE_2D ) & osg::StateAttribute::ON );
+#else
+    // In this mode, osg::Texture::getModeUsage() is undefined, so just detect if a texture is present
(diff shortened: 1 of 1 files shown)


src/osgPlugins/OpenFlight/expGeometryRecords.cpp
```diff
@@ -63,7 +63,12 @@ bool
 FltExportVisitor::isTextured( int unit, const osg::Geometry& geom ) const
 {
     const osg::StateSet* ss = getCurrentStateSet();
+#ifdef OSG_GL_FIXED_FUNCTION_AVAILABLE
     bool texOn( ss->getTextureMode( unit, GL_TEXTURE_2D ) & osg::StateAttribute::ON );
+#else
+    // In this mode, osg::Texture::getModeUsage() is undefined, so just detect if a texture is present
+    bool texOn = (ss->getTextureAttribute(0, osg::StateAttribute::TEXTURE) != NULL);
```
mp3butcher: Hi,
I just take a look at this pr and at first glance (i may be wrong) you check texture on tu 0 instead of tu unit.

Don't explain me why if you did it on purpose but perhaps it's a miss....

emminizer: I agree with you @mp3butcher .  Thanks for pointing this out.

PR #569, opened 2018-06-28
Remove call to nonexistent member seekpos() of std::fpos in VS 15.8

LarryIII:
Remove call to nonexistent member seekpos() of std::fpos in VS 2017 version 15.8. It has removed in our curent latest published release VS 15.8 preview 3(_MSC_VER=1915). So it failed with below issue：
error C2039: 'seekpos': is not a member of 'std::fpos<_Mbstatet>'

See http://eel.is/c++draft/fpos -- to get to an offset you can convert to int; now there is no seekpos member in VS15.8.
We have fixed this issue in vcpkg (Microsoft/vcpkg#3753).

The previous PR that we discussed here: https://github.com/openscenegraph/OpenSceneGraph/pull/567
I send the new one, please help approve. Thanks!


openscenegraph:
Thanks for the revision Larry :-)

PR #570, opened 2018-06-29
Adds a fix for Windows GL3 profile. 

DieSlower:
This PR adds a fix for the broken GL3 profile option on Windows. Windows does not have the outdated GL3/gl3.h library, so if we enable the GL3 profile, it needs to link against the regular windows gl/GL.h

The user will still need to manually link to and call all the ARB functions as they do now, but at least cmake will automatically configure correctly.


DieSlower:
Very well. 
The name is actually a psychological test. It means different things depending on your outlook and demeanor. If you are more negative and take a more snap judgement outlook, then yes, it means something negative to you. If you you are a more positive person and look for a deeper meaning, you  realize because of biology, that it's actually telling you to have a longer life.

PR #576, opened 2018-07-18
variables renaming to avoid name collision with size_t

mp3butcher:


mp3butcher:
I cant remember but it was dealing with memory management (perhaps i included a c header)

PR #578, opened 2018-07-27
updated osgviewerWX to remove deprecated calls in wxWidgets 3.x

nicolas-f:
Cherry-pick old PR of @Alexibu

Needed upgrade of WXWidgets. This PR make sample compatible with WXWidgets from version 3.0.1

I successfully tested it with wxWidgets 3.1

Pull request was closed

https://github.com/openscenegraph/OpenSceneGraph/pull/47

Thanks


robertosfield:
Do you know how far back WxWidget compatibility will go with these changes?  I don't want to merge something that fixes the build for some users but breaks it for others.

nicolas-f:
Yes, WxWidget 3.X is incompatible with WxWidget 2.X . So yes this sample will not work with WxWidget 2.X . But this is only sample code so this PR break nothing. 

Maybe I should specifiy wxwidgets version compatibility in sample CMake file ?

PR #579, opened 2018-07-27
MultiSwitch Name Serialization

dstephan1:
added switch names to osgSim::MultiSwitch serializer

I can remove the test code. Is there any standard place to put tests? CMake question - is it possible to add plugins as dependencies to specific examples? Rebuilding two targets was a bit awkward.


openscenegraph:
Hi David,  thanks for the changes. I've had a quick run through.  Could you separate the serializer changes and the test into separate PR's.  I don't feel the test program is required except for initial testing/debugging, so I would suggest that just sharing this testbed via osg-users ML/forum would be fine for testing.

PR #581, opened 2018-07-29
Load debug plugins when running a debug version of OSG

damiandixon:
When built for debug look for a plugin with 'd' on the name.
The fallback is to the release plugin shared library.


openscenegraph:
Something not right with the motivation for this change, the osgDB::Registry handles the addition of the  OSG_LIBRARY_POSTFIX.

The revision is wrong anyway as it's specific to only one platform so has to be rejected on this account as well.

I can only guess that something has got out of sync with your build of the OSG and that the  and OSG_LIBRARY_POSTFIX isn't set to 'd' for the debug build as it should be.

If you have issues like this then the place to raise it is on the osg-users mailig list/forum as others can then guide you on how to do things, and if there is a problem with the codebase we can work out what is wrong and what the appropriate fix would be.

PR #586, opened 2018-08-02
On cygwin, don't use pthread functions that require _GNU_SOURCE in OpenThreads

cxw42:
This fixes a compilation failure in OpenThreads on cygwin x64 on Win8.1 and Win10 with gcc 7.3.0.  This is the smallest, most targeted fix I can think of, since I didn't want to unintentionally introduce any regressions.

Problem: `pthread_{yield,getconcurrency,setconcurrency}` are not declared in `pthread.h` on cygwin with gcc 7.3.0 (latest as of writing) unless `_GNU_SOURCE` is defined.  However, cmake's [`CheckFunctionExists`](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html) doesn't test for declarations, just for linkage.  Therefore, cmake detects those functions as existing, but then OpenThreads fails to compile because they are undeclared.

Solution implemented in this PR: on cygwin, just don't use the functions that are hidden when `_GNU_SOURCE` is not defined.  For example, without `pthread_yield`, the build falls back to `sched_yield`, which compiles just fine.

Alternative solution: define `_GNU_SOURCE` while compiling OpenThreads on cygwin (as is done now for Android).  However, this changes a number of other functions, and can possibly [change behaviour](https://stackoverflow.com/a/5583764/2877364) of some functions.  Therefore, I think the above approach is less likely to have unintended consequences.  However, if you prefer the `_GNU_SOURCE` approach, I am certainly happy to give it a try and let you know what I find.

### DIFF
--- a/src/OpenThreads/pthreads/CMakeLists.txt
+++ b/src/OpenThreads/pthreads/CMakeLists.txt
@@ -20,6 +20,11 @@ SET(TARGET_SRC
     ../common/Version.cpp
     ../common/Atomic.cpp
 )
+
+IF(CYGWIN)  # define for pthread_{yield,getconcurrency,setconcurrency}
+    ADD_DEFINITIONS(-D_GNU_SOURCE)
+ENDIF()
+
 IF(ANDROID)
       ADD_DEFINITIONS(-D_GNU_SOURCE)
       SET(CMAKE_REQUIRED_DEFINITIONS "${CMAKE_REQUIRED_DEFINITIONS} -D_GNU_SOURCE")


openscenegraph:
Just below your changes is an entry:

cxw42:
Good point.  Unfortunately, as far as I can tell, `CHECK_FUNCTION_EXISTS` only determines whether symbol `_pthread_yield` exists in the system link libraries ([doc](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html), [source](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/CheckFunctionExists.c#L6)).  On my system, `CHECK_FUNCTION_EXISTS(pthread_yield HAVE_PTHREAD_YIELD)` sets `HAVE_PTHREAD_YIELD` true regardless of the setting of `_GNU_SOURCE`.  However, the resulting calls to `pthread_yield()` in OpenThreads don't compile because nothing has actually defined `_GNU_SOURCE`.

Context: I have done some work on the scripting support that I would like to eventually send to osg-submissions.  I need to rebase from 3.5.6 to master first.  Before I do that, I would like to get `mkdir build && cd build && cmake .. && make` to run to completion cleanly so that the rebased code will be easier for others to test.  

Even though cygwin isn't a widely-used platform, I appreciate that OSG supports it --- that support is one of the first things that drew me to the project.

openscenegraph:
The more we can avoid special cases populating the code and build system the easier it is to maintain support for niche platform combinations, so I'm keen to avoid special cases being added.

Is there a reason not just to add a define for _GNU_SOURCE?

cxw42:
I'll give it a try!

cxw42:
With `_GNU_SOURCE`, it compiled cleanly.  I tested it by running `osgviewer cow.osgt` and cycling the threading model with `m`.  I didn't see any glitches or errors under any of the threading models.  Looks good to me!

PR #587, opened 2018-08-02
Drawable: Explicitly release VAO on destruction to avoid resource leak.

emminizer:
This appears to help the issue identified on the mailing list.  It may not be the correct fix but it seems like a step in the right direction.

The code reads like it was heavily influenced by Display Lists.  `~Drawable()` destroys associated display lists indirectly, by calling `dirtyGLObjects()`, which gets the call eventually to the contexts' `DisplayListManager`.  VAO was not making it to its `VertexArrayStateManager` class.  It looks like the hook is there with `VertexArrayState::release()`, but it wasn't being called when the drawable closed out.  Now it is called.

I tested on Core Profile with 3.6.2 and this modification.  Afterwards, I no longer see a leak.  Before, I was leaking up to 3 megs a minutes.

One note and why I think this might not be the right solution: I added a counter to the code to count the number of `glGenVertexArrays` to `glDeleteVertexArrays` calls.  I'm missing one on exit.  It seems like the VAO created in the very last frame doesn't get cleaned up.  I tried to track this down but failed to do so.  Since a resource leak of 1 VAO is better than 1 per frame, I thought this was worth at least posting for your consideration.  Thanks.

### DIFF
--- a/src/osg/Drawable.cpp
+++ b/src/osg/Drawable.cpp
@@ -257,6 +257,11 @@ Drawable::Drawable(const Drawable& drawable,const CopyOp& copyop):
 Drawable::~Drawable()
 {
     dirtyGLObjects();
+    for (unsigned int i = 0; i<_vertexArrayStateList.size(); ++i)
+    {
+        VertexArrayState* vas = _vertexArrayStateList[i].get();
+        if (vas) vas->release();
+    }
 }
 
 osg::MatrixList Drawable::getWorldMatrices(const osg::Node* haltTraversalAtNode) const


openscenegraph:
After a quick look, my inclination would be that the ~Drawable() destructor should probably call releaseGLObjects() rather than dirtyGLObjects(), though have no clue whether this will help in this instance as releaseGLObjects() is currently calling dirtyGLObjects()...

mp3butcher:
@emminizer 
since you seams to have an mod osg that tell you a bit more about a code change impact, could you test it
replace in all Drawablles::createVertexArrayState
` VertexArrayState* vos = new osg::VertexArrayState(renderInfo.getState());`
  with
   VertexArrayState* vos = _vertexArrayStateList[renderInfo.getContextID()] = new osg::VertexArrayState(renderInfo.getState());

mp3butcher:
releaseGLObjects seams a bad idea:
releaseGLObjects require a proper State to be called so can't be invoked in destructor
Further if by some magic you find a state out of your hat, it would destroy GL resources potentially used by other Drawable.

in this pr by calling vas->release(); the state used to release GL is the weak pointer given at vas creation, so it's working but perhaps there's a more elegant way

openscenegraph:
I have investigated the issue and the way to resolve it was to call the Geometry::releaseGLObjects() from the Geometry destructor, and the Drawable::releaseGLObjects() from the Drawable destructor, and to refactor the releaseGLObjects() so that it properly handle both display lists and release of the VertexArrayState objects.

With the this Dan's leaking_vao test program no longer increases in memory as the goes on, instead stabilizes right away.

The fix has been checked into master and the 3.6 branch:
https://github.com/openscenegraph/OpenSceneGraph/commit/eee5d5482e41290e7c210273f8c6bae489cedb76

PR #588, opened 2018-08-02
Replace REGISTER_WINDOWINGSYSTEMINTERFACE for X11 using OSGVIEWER_EXPORT

cxw42:
This fixes a linker error in example_osgslice on cygwin x64, Win10, gcc 7.3.0.

Problem: when compiling on cygwin for dynamic linking (using #586), osgslice can't link because `graphicswindow_X11()` is not found.

Solution: Use the same `REGISTER_WINDOWINGSYSTEMINTERFACE` replacement for X11 as for Win32 (0bca415d5af4647da868888c1588b9617dbe3eb8).

Explanation: On cygwin, X11 is the default windowing system.  Since `osgViewer` is compiling to a Win32 DLL, `X11WindowingSystem` needs `OSGVIEWER_EXPORT` just like `Win32WindowingSystem` does.  On non-Windows platforms, `OSGVIEWER_EXPORT` is a no-op, so this should not affect any other platform.

### DIFF
--- a/src/osgViewer/GraphicsWindowX11.cpp
+++ b/src/osgViewer/GraphicsWindowX11.cpp
@@ -2277,7 +2277,9 @@ public:
 
 };
 
-REGISTER_WINDOWINGSYSTEMINTERFACE(X11, X11WindowingSystemInterface)
+// Replaces REGISTER_WINDOWINGSYSTEMINTERFACE using OSGVIEWER_EXPORT
+extern "C" OSGVIEWER_EXPORT void graphicswindow_X11(void) {}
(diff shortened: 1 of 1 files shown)


openscenegraph:
Replace a neat a tidy macro with a messy hack for just one niche build combination.  No that's not an acceptable change.

cxw42:
@openscenegraph Fixed per your comment.  I was also able to remove the hack from src/osg/GraphicsWindowWin32.cpp.  Would you be willing to consider [the new version](https://github.com/openscenegraph/OpenSceneGraph/compare/master...cxw42:cygx11?expand=1)?  It's not showing up here (I think because the PR is closed).  Much appreciated!

(I also added a few lines of doxygen - happy to remove them if you prefer.)

PR #590, opened 2018-08-09
Prevent culling of Drawables with culling disabled, but which have valid bounds

AnyOldName3:
If the scene graph contains any Drawables with culling disabled, but which have valid bounds, and the CullVisitor is set to compute the near and far planes, it's possible for the Drawables to be culled.

Such a situation could occur if, for example, the drawable's vertices were to be repositioned by the vertex shader, but a programmer hadn't set the bounds to an invalid value as there was never any need when the application was initially written as it used fixed near and far planes.

I've spent the last six days trying to stop shadows making animated meshes disappear in OpenMW, and would rather no one else has to go through anything similar. This fix may not actually be necessary if there was big writing in lots of places warning that this could happen and saying how to avoid it, but I'm pretty sure it's a bug if a node with culling disabled still gets culled, so it's probably better off merged.

### DIFF
--- a/src/osgUtil/CullVisitor.cpp
+++ b/src/osgUtil/CullVisitor.cpp
@@ -924,7 +924,7 @@ void CullVisitor::apply(osg::Drawable& drawable)
 
     if (_computeNearFar && bb.valid())
     {
-        if (!updateCalculatedNearFar(matrix,drawable,false)) return;
+        if (!updateCalculatedNearFar(matrix,drawable,false) && drawable.isCullingActive()) return;
     }
 
     // need to track how push/pops there are, so we can unravel the stack correctly.


mp3butcher:
you will see that with Robert but i have the feel this patch will not pass:
Can you reproduce the behavior in a simple example and post it on the forum before submitting a patch (it would spare you some avoidable rejections)
If I understand your issue the following will perhaps help:
For hardware animated mesh the classic trick is to setInitialBound with respect to animations pose extrema.
I have the code for this i will merge it to animationviewer

openscenegraph:
Thanks for the looking into this.  I don't this it's the most efficient way to resolve this issue.  Technically compute near/far is different from culling so switching off compute near/far isn't ideal.  Having an invalid bounding box is the most efficient way of disabling both.

If we are to using CullingActive to disabling compute near/far then the way to do it would be to restructure the code so that the if (isCullingActive()..) line above, i.e. pseudo code follows:

if (isCullingActive) 
{
    if (isCulled) return;
    if (!updateComputeNearFar) return;
}

--

As a general comment, guys, your github names come across as childish, the type of things my children would come up with when they were 10.  Your github account is part of your public profile to the world, potential employers, other members of the community.  So might think it "cool" but plenty like myself will find it more difficult to take you and your work seriously.
}

AnyOldName3:
In the example where this actually was a problem in real code, there was a decision to not mark the bounds of an animated Geometry as dirty after skinning because it was unnecessary work - the parent node had bounds computed based on the skeleton, so the Geometry would only be reached when it actually needed drawing anyway. Therefore the Geometry had the same bounds as before skinning, despite actually being somewhere else. Ideally, its bounds should have either been set to something invalid or a computed value, but it had been working for years until I started changing other things.

The check to see if culling is active for the Drawable ended up after the near/far plane computation because conceivably there could be a Drawable with dodgy bounds but which produced sensible results when the near/far planes were computed based on primitives rather than bounding volumes. If you want to change things so that only nodes with culling active actually contribute to plane computation, that's your decision, but it sounds to me like the kind of thing that needs mentioning in some documentation.

The main purpose of this PR was to illustrate that its possible for Drawables with culling disabled to still get culled, which is certainly not intuitive. Now other people are aware of the situation, the specific implementation of a fix isn't important to me.

Regarding names, I've used this name online for over a decade and at this point, using anything else would be weird. Excluding the OSG mailing list and OSG GitHub, though, literally no one anywhere cares that it's not a real name. Robert's crusade to call everyone childish is far more unprofessional in my opinion - this is my third time coming up against it and it's not getting any more pleasant.

openscenegraph:
It sounds like the right fix for the problem you have is to disable the computation of the bounding box using a ComputeBoundCallback.  It's not perfect, but is the most appropriate thing to do as t will directly fix the issue, rather than hide it as this suggested fix is.  

I'm going to close this PR as it doesn't resolve any issue, it only hacks around a problem elsewhere.

As for names, well if go by a handle that I think is childish I will call you out on it, saying that you've go by it for years just makes it worse, to me it just seems like you've been childish for years and haven't yet grown up.  This is impression you are conveying to others by using these silly handles.  

I have to deal with hundreds of contributors over many years, the silly handle shell game only hinders the ability to know who was has discussed what which whom, who actually contributed.  It obfuscation.  There is a good reason why we don't go by made up handles in real life, it's not respectful to others and hinders communication.  Would you go by this handle for a job interview?  At work? When writing an academic paper? Down the pub with friends?

If you really want to stick with a silly handle for you github at least use your own name, while @mp3butcher doesn't have a git handle conveys professional, at least it's possible to figure out who he actually is and to enable us to work out a paper trail for discussions and submissions so we have chance of joining up the dots when we need to.

PR #592, opened 2018-08-11
replace unsized InternalFormat GL_RGBA to GL_RGBA16 for ImageLess Textures

mp3butcher:
…and set sourceFormat to GL_RGBA
correct a bug introduced by the use of immutable texture storage
According to several users (http://forum.openscenegraph.org/viewtopic.php?t=17437)
the bug is involve in others examples
It may be GL_RGBA8 will mod it if required

### DIFF
--- a/examples/osgSSBO/osgSSBO.cpp
+++ b/examples/osgSSBO/osgSSBO.cpp
@@ -799,7 +799,8 @@ osg::Node* createPrerenderSubgraph(osg::Node* subgraph, const osg::Vec4& clearCo
 
     osg::Texture2D* texture = new osg::Texture2D;
     texture->setTextureSize(tex_width, tex_height);
-    texture->setInternalFormat(GL_RGBA);
+    texture->setInternalFormat(GL_RGBA16);
+    texture->setSourceFormat(GL_RGBA);
     texture->setFilter(osg::Texture2D::MIN_FILTER, osg::Texture2D::LINEAR);
     texture->setFilter(osg::Texture2D::MAG_FILTER, osg::Texture2D::LINEAR);
 
--- a/examples/osgoit/DepthPeeling.cpp
+++ b/examples/osgoit/DepthPeeling.cpp
@@ -307,7 +307,8 @@ void DepthPeeling::createPeeling()
         colorTexture->setFilter(osg::Texture::MAG_FILTER, osg::Texture::NEAREST);
         colorTexture->setWrap(osg::Texture::WRAP_S, osg::Texture::CLAMP_TO_BORDER);
         colorTexture->setWrap(osg::Texture::WRAP_T, osg::Texture::CLAMP_TO_BORDER);
-        colorTexture->setInternalFormat(GL_RGBA);
+        colorTexture->setInternalFormat(GL_RGBA16);
+        colorTexture->setSourceFormat(GL_RGBA);
 
         _colorTextures[i] = colorTexture;
     }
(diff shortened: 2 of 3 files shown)


openscenegraph:
I have just had a quick look over the changes and the thread.  Having to change examples to fix an issue that was introduced by changes to the core is a bit of red flag.  

Occasionally it is the case that user code is buggy and changes to the core reveal this, and in these case it necessary to push the fix onto user code.  In this instance, I'm not at point where I think this is clear cut example of this.

My first instinct is to see if we can tighten up he fall-back for the internal format for cases where it's not defined.  I haven't had a chance to look into the issue yet, so can't provide any answers. 

Looking at Texture.cpp and the _sourceFormat and _internalFormat, it's all a bit convoluted now, originally it was all quite simple but the various contributions over the years have pushed/pulled the implementation in various directions.   How to resolve it isn't yet clear.

Perhaps have some fallback computation of the _internalFormat for when it's not set, there isn't an osg::Image assigned to the texture to use as a guide of the format, and then the _sourceFormat is set.

This isn't to say that this PR shouldn't be merged, only that if we could get things working without this PR by having a fallback in the core so that end user code works fine as it did before.  Then possibly we could merge this PR as it provides an example of a good practice - i.e. explicitly defining the the format rather than relying upon a fallback.

mp3butcher:
According to glTexImage2D documentation, the support for unsized internal format is quite oldish... and in my sense should not have been supported by osg in the first place (it's quite a miracle that calls to glTexImage2D find a way to work with both internalformat and format = GL_RGBA) ...

Concerning this correction I think there's nothing wrong in "hacking"(there's no real source) sourceFormat  as osgoit already use this kind of trick to create imageless depthtexture
in examples/osgoit/DepthPeeling.cpp l.281:
```
        _depthTextures[i]->setInternalFormat(GL_DEPTH24_STENCIL8_EXT);
        _depthTextures[i]->setSourceFormat(GL_DEPTH_STENCIL_EXT);
        _depthTextures[i]->setSourceType(GL_UNSIGNED_INT_24_8_EXT);

```

I don't understand whatumean by 'some fallback computation of the _internalFormat' in that case
we only can derive format from sized internalformat, internalformat concern GPUside representation and is mandatory to be specified by user (even with an attached image)

openscenegraph:
Please remember that we are trying to maintain backwards compatibility in the OSG, we aren't trying to break compatibility.  The introduction of gTextureStorage is breaking backwards compatibility due to it's dependent on the internalFormat being set. whereas the previous incarnation wasn't.  It's you PR that has broken other user code here.  What I'm looking for is a resolution to this without forcing other users to amend their code.

mp3butcher:
it wasn't professional at the first place to use unsized internal format.
I just put some sense and good practice in your dear buggy library

openscenegraph:
Simple question.  What platform portability are you expecting with force RGBA16 onto this examples?  

The osgoit example is a modern usage case and we can expect it only to work on modern hardware, the osgprerender example is over 15 years old so has been happily working across a range of hardware and OS's through that period.  It doesn't make any sense to add restrictions such as 16bit formats.

One of the advantages of having RGBA vs sized types is that it enables the GL driver itself to decide what the hardware is capable of and provide the appropriate type.  For many applications that exact type isn't critical so have better portability that this driver mapping provides is advantageous.  For those application where the data type is critical then we'd want the ability to set the data explicitly, but it's not something we should enforce unless as an extra constraint over what OpenGL has traditionally provided.  BTW, this ability of the GL driver to work out what will work natively is a big part of why GL applications can work over decades, it's one of GL key strengths.

mp3butcher:
I use GL_RGBA16 because its the worse case : not present in Texture::sizedInternalFormats

PR #595, opened 2018-08-15
Remove duplication of REGISTER_WINDOWINGSYSTEMINTERFACE code in GraphicsWindowWin32

cxw42:
Refactor so `src/osgViewer/GraphicsWindowWin32.cpp` doesn't duplicate code in `include/osg/GraphicsContext`.  New macro `REGISTER_WINDOWINGSYSTEMINTERFACE2` can accept `OSGVIEWER_EXPORT`, introduced into Win32 in May as part of 0bca415d5af4647da868888c1588b9617dbe3eb8.

Also fixes the Cygwin build problem identified in #588, and still present in master (e008784ab67be96e3e079aecfd2fc115b7b7885b).  Uses a neat and tidy macro, per Robert's feedback on #588.

Also tested and works with osgQt on my Cygwin setup (0bca415d5af4647da868888c1588b9617dbe3eb8 refers to osgQt, so I tried it out).

### DIFF
--- a/include/osg/GraphicsContext
+++ b/include/osg/GraphicsContext
@@ -597,10 +597,15 @@ struct WindowingSystemInterfaceProxy
     osg::ref_ptr<T> _wsi;
 };
 
-#define REGISTER_WINDOWINGSYSTEMINTERFACE(ext, classname) \
-    extern "C" void graphicswindow_##ext(void) {} \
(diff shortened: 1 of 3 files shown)


openscenegraph:
This looks more sensible, thanks.

PR #596, opened 2018-08-15
add a bool parameter to allow forced reindexation of a mesh

mp3butcher:
IndexMeshVisitor don't reindex indexed
http://forum.openscenegraph.org/viewtopic.php?t=17418

### DIFF
--- a/include/osgUtil/MeshOptimizers
+++ b/include/osgUtil/MeshOptimizers
@@ -46,7 +46,7 @@ class OSGUTIL_EXPORT IndexMeshVisitor : public GeometryCollector
 {
 public:
     IndexMeshVisitor(Optimizer* optimizer = 0)
-        : GeometryCollector(optimizer, Optimizer::INDEX_MESH), _isForcedReindexationEnable(false)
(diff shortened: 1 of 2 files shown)


openscenegraph:
I have done a first pass review but won't merge right away as the naming of the feature is awkward, I don't know what it should be called yet but I'm hopeful there is a more natural way to name it.  As a general note, puttingeverytingaslowercasewithinasinglewordisaboutasundreadableasyourcanmakeavariable.

mp3butcher:
I agree with lower case naming and will fix it but
bool _isForcedReindexationEnable seams explicit enough...
what would you propose?

openscenegraph:
How about something like set/getGenerateNewIndicesOnAllGeometries(bool)?

mp3butcher:
ok done... I let the _isForcedReindexationEnable variable 
I would think it emphase semantic

openscenegraph:
Thanks for the change, we are getting close now.  

I had to wince when looking at the two different names - this is just dreadful example of how to do things, it's worse than just having the original name.  What will coders following in our footsteps make of this code?  They'll think the original coder has a multiple personality disorder...

I will see if I can fix it and then get it checked in.

PR #598, opened 2018-08-16
Replace option noTriStripPolygons with triStripPolygons (inverting …

mp3butcher:
…default behavior too).
As a remark I would say that a readNode plugin shouldn't optimize without user consent


openscenegraph:
This change will break user code that uses the original option string so isn't acceptable as is.

In general I would agree that ideally you'd not do optimization in plugins and leave this up to users to do, but often the data loaded in it's raw form is dreadful for real-time performance that it's pretty unusable without optimization.   

The obj plugin right now produces deadful scene graphs so needs a rewrite on how it builds the scene graph, we are seeing even bigger performance issues with how the obj plugin creates scene graph now that the OSG is assigning VBO's which is required to support VAO.  So now more than ever we need to optimize the obj loading.

mp3butcher:
IndexMeshVisitor would give better performance than tristripping (less primitive set count)

PR #600, opened 2018-08-17
use selectSizedInternalFormat to fix image less glTexStorage paths

mp3butcher:
..leaving image less Cube map path to be impelmented

### DIFF
--- a/src/osg/Texture1D.cpp
+++ b/src/osg/Texture1D.cpp
@@ -236,10 +236,11 @@ void Texture1D::apply(State& state) const
 
         applyTexParameters(GL_TEXTURE_1D,state);
         GLExtensions * extensions = state.get<GLExtensions>();
-        bool useTexStorrage = extensions->isTextureStorageEnabled;
+        bool useTexStorage = extensions->isTextureStorageEnabled && extensions->glTexStorage1D!=0 && (_borderWidth==0);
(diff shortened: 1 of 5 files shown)


mp3butcher:
will think of a way to remove this path divergence

openscenegraph:
You beat me to checking these changes in.  I made the same changes as in your PR, but one competition of this spotted that the textureObject = generateAndAssignTextureObject(..) code above each of these glTexStorage entries was using the _internalFormat as the format to match, totally ignoring the fact that the glTexStorage code uses sizedInternalForamt and the glTexImage code uses the _sourceFormat in it's logic, so... there is potential for attempting to reuse texture objects of a different format.

To resolve this bug we have to pass in the actual internal format being used and easiest way to resolve this looks to be duplicating the generate & bind code for each of the two branches.

There may be a way of avoiding all this branching and code duplication, but rather than keep guessing what this might be I'll get my changes checked in:

https://github.com/openscenegraph/OpenSceneGraph/commit/d37b826a745b4638666b0d53c613033a17df4397

Could you test this out and if I've missed anything fill in the gaps.  I'll close this PR as I think I have covered this all in the above commit, plus extra work.   I suspect we'll need to revisit this code again, and if you have ideas on how to simplify/unify it then please post your ideas.

Cheers,
Robert.

mp3butcher:
good catch with the generateAndAssignTextureObject.. Haven't seen it

PR #601, opened 2018-08-17
TexStorage code cleaning + add feature on Texture2DArray

mp3butcher:
…g representation)

### DIFF
--- a/src/osg/Texture.cpp
+++ b/src/osg/Texture.cpp
@@ -2318,20 +2318,19 @@ void Texture::applyTexImage2D_load(State& state, GLenum target, const Image* ima
             int width  = inwidth;
             int height = inheight;
 
-            bool useTexStorage = extensions->isTextureStorageEnabled && extensions->isTexStorage2DSupported() && (_borderWidth==0);
-            GLenum sizedInternalFormat = useTexStorage ? selectSizedInternalFormat(image) : 0;
-            if (useTexStorage && sizedInternalFormat!=0)
+            GLenum texStorageSizedInternalFormat = extensions->isTexStorage2DSupported() && (_borderWidth==0) ? selectSizedInternalFormat(image) : 0;
+            if (texStorageSizedInternalFormat!=0)
             {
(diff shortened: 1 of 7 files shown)


openscenegraph:
Thanks.  I have merged 

https://github.com/openscenegraph/OpenSceneGraph/pull/601/commits/d23f1dd56dc667c8b26ae0eecdc8cb631d548568

But haven't yet merged 

https://github.com/openscenegraph/OpenSceneGraph/pull/601/commits/45620991a116bef88a1b8f66f8557b938ec603a4

As I think if we are to conflate usage together to remove the redundant comparison then we should also renamed the variable to make it clear that this role has been merged i.e.

bool useTexStorage ....
GLeunm sizedInternalFormat....

To be merged to become something like:

GLenum useTexStorageSizeInternalFormat ...

A bit long winded but makes the if () make a bit more sense, one could also rewrite the bool useTexStorage line so it does the check against the GLExtension & _borderWidth on the same line as call to the selectSizedInternalFormat(image).

This way we can reduce the number of vars, checks and lines of code whilst improving the clarity of what role the variable(s) have.

mp3butcher:
done

mp3butcher:
done

mp3butcher:
done
I add missing implementation to Texture2DArray too

mp3butcher:
ARG..damned me.. Dont take the last commit...

robertosfield:
Merging with 3.6.x is a possibility as I don't think it affects the API. However, it's a bit risky and push up how much testing will need to be done in the community,  We could make a branch off 3.6 merge the TexStorage get this tested first then merge into 3.6 if it looks safe enough.

openscenegraph:
I have made a OpenSceneGraph-3.6-TexStorage branch so that experiments with merging the glTexStorage changes into the 3.6 branch can be safely made:

https://github.com/openscenegraph/OpenSceneGraph/tree/OpenSceneGraph-3.6-TexStorage

PR #607, opened 2018-08-20
GL_RGBA16F and GL_RGBA32F not defined on Windows

plevy:
GL_RGBA16F and GL_RGBA32F not defined on Windows so use _ARB versions.  I assume since Travis passed, this is a Windows issue


airscrat:
Yes, windows default gl has not defined GL_RGBA16F and GL_RGBA32F.

PR #609, opened 2018-08-26
give more priority to ARB suffixed

mp3butcher:
in order to fix a recurrent malfunction of osgbindlesstext under linux nvidia proprietary driver


openscenegraph:
The core version should always take precedent over the extension.  The way that GL is developed is features go from proprietary extension like NV then get promoted to a ARB then if it's good enough to get in the core it gets support directly in a major GL release.

For the ARB version to be working but the non core version to fail strongly looks like a driver bug.  Driver bugs come and go, they can easily morph - if we merged this then it could break with the next drive release.  It's just not practical to go chasing driver bugs - the right way to fix them is contact the driver developer, report the problem and get it fixed that way.

PR #614, opened 2018-09-04
Protect the _programSet in Shader with a mutex.

jasonbeverage:
Hi Robert!

Here is a change to Shader that protects the _programSet with a mutex.  Crashes can occur if you reuse Shader objects between multiple Programs b/c the addProgramRef and removeProgramRef functions aren't thread safe.  The VirtualProgram framework in osgEarth can generate Programs as needed using shared Shader objects, so we would see crashes when running with multiple windows using CullThreadPerCameraDrawThreadPerContext mode.

I've attached a simple example that shows the issue, it just creates and destroys programs in background threads that share a global Shader object.

Thanks!

Jason
[sharedshader.txt](https://github.com/openscenegraph/OpenSceneGraph/files/2349377/sharedshader.txt)

### DIFF
--- a/include/osg/Shader
+++ b/include/osg/Shader
@@ -347,6 +347,7 @@ class OSG_EXPORT Shader : public osg::Object
         /** osg::Programs that this osg::Shader is attached to */
         typedef std::set< osg::Program* > ProgramSet;
         ProgramSet                      _programSet;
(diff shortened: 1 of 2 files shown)


openscenegraph:
Thanks Jason, I've merged with master 3.6 branch so it'll be in the upcoming 3.6.3.

PR #616, opened 2018-09-05
ParticleSystem: Added support for GLES2 and upper (GL_QUADS -> GL_TRI…

valid-ptr:
…ANGLES)

There is no GL_QUADS on GLES2 and upper, so I've just changed GL_QUADS to GL_TRIANGLES, like I've done this some time ago for Geometry::createTexturedQuadGeometry()

KOS


openscenegraph:
This PR will introduce a slow down for all other platforms so isn't a good solution.  

For GLES I would recommend not using osgParticle and instead rolling your own particle systems.

valid-ptr:
OK, I will correct the PR soon...

valid-ptr:
Hello, Robert!

Please, Checkout!

PR #621, opened 2018-09-08
fix examples context creation for X11

mp3butcher:
(when DISPLAY not :0.0)
It will avoid a lot of complains related to the new Ubuntu (gnome3) default policy (DISPLAY=:1) 
so merging it on release branch can be good idea too


openscenegraph:
Thanks Julien, good effort scooping up all the various Traits usage, didn't realize there was so many!

PR #622, opened 2018-09-10
osgText: fixed thread-safety issues in Glyph and Font

gwaldron:
Patch to address a thread-safety issue when creating osgText objects from multiple threads simultaneously.


openscenegraph:
I merged this change and now any time text is use I get a hang. This is a very serious regression that I'm now investigating,  Basically we have a the same thread acquiring the same mutex and then hanging as a consequence.  Was this PR only tested under Windows?

gwaldron:
Yes - only tested under Windows. Glad to see it works with the reentrant mutex.

PR #638, opened 2018-10-09
fix precision numeric for stl writer

alavenant:
We want to be more accurate for stl files.


mp3butcher:
syntax is ok but you call it on each triangle..:/ perhaps there's a better place to set this (osgDB::OutputStream?)

PR #644, opened 2018-10-10
add indexed accesor to event handler collection

mp3butcher:
minor


openscenegraph:
Yuk, hacky... For master I think it would be easier to just change the EventHandlerList to be a to std::vector<> the the access would be straight forward.

PR #646, opened 2018-10-12
Use options when loading images in Collada loader

flashk:
The Collada loader uses the global registry options when loading images, instead of the options passed into the loader. This patch modifies the loader to use the specified options to load images.

### DIFF
--- a/src/osgPlugins/dae/ReaderWriterDAE.cpp
+++ b/src/osgPlugins/dae/ReaderWriterDAE.cpp
@@ -144,7 +144,7 @@ ReaderWriterDAE::readNode(const std::string& fname,
     std::string fileName( osgDB::findDataFile( fname, options ) );
     if( fileName.empty() ) return ReadResult::FILE_NOT_FOUND;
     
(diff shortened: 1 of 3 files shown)


src/osgPlugins/dae/ReaderWriterDAE.cpp
```diff
@@ -143,6 +143,9 @@ ReaderWriterDAE::readNode(const std::string& fname,
 
     std::string fileName( osgDB::findDataFile( fname, options ) );
     if( fileName.empty() ) return ReadResult::FILE_NOT_FOUND;
+    
+    pluginOptions.options = options ? static_cast<Options*>(options->clone(osg::CopyOp::SHALLOW_COPY)) : new Options;
```
robertosfield: In recent OSG release you can use the osg::clone(options) template function instead of static_cast<Options*>(options->clone(osg::CopyOp::SHALLOW_COPY)).  The clone template function is defined in the include/osg/Object header.

flashk: Thanks Robert. I've updated the code to use the osg::clone() function.

PR #648, opened 2018-10-22
fix compilation with OSG_USE_FLOAT_PLANE

LaurensVoerman:
Hi Robert,
master branch compilation fails with OSG_USE_FLOAT_PLANE defined. Not sure if there is still a valid case for doing this, I did not see a speedup on x64 at all. 

Regards, Laurens.


openscenegraph:
I have checked in a simpler solution

https://github.com/openscenegraph/OpenSceneGraph/commit/9a21a1d1cf57395800b534a9d8e407967670d3a3

PR #650, opened 2018-10-25
remove apply(Node&) overload on BoneMapVisitor

mp3butcher:
to prevent visitor to exit before complete BoneMap has been collected


mp3butcher:
to test the diff in behavior, It require to hack a little the osganmiationhardware example:
http://forum.openscenegraph.org/viewtopic.php?p=75126#75126


bug affect also master

PR #656, opened 2018-11-06
Fix OSG_GL_CONTEXT_VERSION default values

plevy:
On my Windows build, line 532 does not properly set the context version but this is a workaround to set it properly to 3.3 when GL3 is available

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -519,7 +519,12 @@ ELSE()
     OPTION(OSG_CPP_EXCEPTIONS_AVAILABLE "Set to OFF to disable compile of OSG components that use C++ exceptions." ON)
 ENDIF()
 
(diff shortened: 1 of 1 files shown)


robertosfield:
I had a quick look at the change and I don't like duplicating a long line just to workaround for a later cmake set() not working as intended.  It would be better to get to the bottom of why this is not working correctly and amend it properly without duplicating a long statement.

plevy:
Agreed, it is just a workaround for some other issue.  Thanks for looking at it.

openscenegraph:
I tried out several different combination of cmake usage to get the OSG_GL_CONTEXT_VERSION to use the appropriate default and ended up with this commit to resolve the issue:

   https://github.com/openscenegraph/OpenSceneGraph/commit/8b4359b77e9eb1436ff106e1262375a7a6620bbe

PR #659, opened 2018-11-13
Fix duplicate geometries in FBX export.

dquam:
The fix to allow the export of orphaned Geometry nodes accidentally
duplicated those geometry nodes in 'normal' scene trees. This commit
refactors the code to rely on the visitor to traverse the Geometry as
opposed to traversing the Geometry in the Geode visit and the Geometry
visit.

### DIFF
--- a/src/osgPlugins/fbx/WriterNodeVisitor.cpp
+++ b/src/osgPlugins/fbx/WriterNodeVisitor.cpp
@@ -598,6 +598,14 @@ void WriterNodeVisitor::buildFaces(const std::string& name,
         mesh->EndPolygon();
     }
     setControlPointAndNormalsAndUV(geometryList, index_vert, texcoords, mesh);
+
+	_geometryList.clear();
+	_listTriangles.clear();
+	_texcoords = false;
+	_drawableNum = 0;
+	for (MaterialMap::iterator it = _materialMap.begin(); it != _materialMap.end(); ++it)
+		it->second.setIndex(-1);
+	_lastMaterialIndex = 0;
 }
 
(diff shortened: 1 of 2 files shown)


dquam:
This pull request addresses an issue introduced in #411 which added the apply(osg::Geometry) method but didn't remove the traversal from the apply(osg::Geode) method resulting in duplicate geometry export for most scenegraphs.

openscenegraph:
I have done a first pass review and it looks to me like the new code will call buildFaces(..) on every Group in a hierarchy as WriteNodeVisitor returns back up the scene graph.   So if you have a Gruop that contains a Group you'll end up with two calls to buildFaces on the same data.

The approach you've taken might be more workable if the clears are done after the buildFaces() is called.

The fallback check in the WriterNodeVisitor::apply(Geometry&) that ensures buildFaces() is called if the Geometry is a root node (the getNumParernts()==0) isn't robust as you could have a geometry in the scene graph but call the writeNodeFile() and hence the WriteNodeVisitor() on a geometry leaf rather then whole scene graph.  In this case the faces would never be written out as no Group would be traversed and the Geometry has more than one parent. 

The fallback would probably be better handled by checking the NodeVisitor::getNodePath() to see what nodes have been traversed so far.

As things stand these two issues will need resolve before the changes can be merged.

dquam:
Thanks for the feedback. I attempted to address the issues you presented. I moved the clears inside of buildFaces and changed the root node check to use getNodePath().size() == 1. Please let me know if I misunderstood your requests or there are any further issues.

Issue #658, opened 2018-11-13
CMake installation could use some modernization


mxgrey:
First of all, thanks for developing and maintaining this incredibly useful open source software. For years now, it's been my favorite scene graph rendering software to work with, bar none.

I have a suggestion that I hope won't come across as too frivolous, because it has a significant impact on my whole workflow as a software library developer: OSG's integration with cmake uses some outdated practices with known weaknesses, and it would be great to modernize the build system, if only a bit.

To add some context, I'm developing on Ubuntu 18.04 which conveniently provides [osg-3.2](https://packages.ubuntu.com/bionic/libopenscenegraph-dev) and [osg-3.4](https://packages.ubuntu.com/bionic/libopenscenegraph-3.4-dev) packages. However, there were some recent developments in OSG that I'd like to take advantage of, but they're not available until after `osg-3.4`.

I'm able to build the master branch of OSG with no problem, so it's easy to get the latest version :+1: , but actually *using* the latest version in a downstream dependent cmake project poses some issues. Ideally I should be able to use `find_package(OpenSceneGraph <version>)` and cmake should automagically find me a version of OSG that matches my version requirement. However, the default way that cmake will do this is to run [this find-module](https://gitlab.kitware.com/cmake/cmake/blob/5e12fad8704e4797ecda3906f3b558775b70539f/Modules/FindOpenSceneGraph.cmake) which will ultimately lead to [this find_library(~)](https://gitlab.kitware.com/cmake/cmake/blob/5e12fad8704e4797ecda3906f3b558775b70539f/Modules/Findosg_functions.cmake#L45) call.

The problem here is that the newest version of OSG will install to `lib64`, but these system-installed find-modules will only look at [lib directories](https://gitlab.kitware.com/cmake/cmake/blob/5e12fad8704e4797ecda3906f3b558775b70539f/Modules/Findosg_functions.cmake#L54), without the `64` suffix. I've noticed that the OSG codebase [provides a find-module](https://github.com/openscenegraph/OpenSceneGraph/blob/85504b45f6229c2695bc109d771045fcb10fc1e0/CMakeModules/FindOSG.cmake#L32) which can handle the `lib64` case, but the only way for a downstream project to use that find-module is to copy/paste it into their own codebase and set up their `CMAKE_MODULE_PATH` to look at this custom find-module instead of using the system-installed one.

While that may be a sufficient solution in some cases, it has some significant shortcomings.

[...]

PR #660, opened 2018-11-14
Create a config-file and export library targets for downstream users

mxgrey:
I'm marking this as a [WIP] because I still need to test it out. **If anyone would be kind enough to test this on Windows, I would greatly appreciate it**.

This will only affect users with `cmake-2.8.8` or later. It should export targets of all the OSG libraries and create a config-file that can import those targets into consumer projects.

Again, I haven't had a chance to test this yet, so there may be errors, but I wanted to open this PR in case anyone has immediate concerns or feedback.

This addresses issue #658.


mxgrey:
While testing I've noticed that there are currently some issues with the imported targets. Since some of the component libraries depend on each other, it's important for each component library target to import any targets it depends on before it imports itself, and cmake doesn't do this automatically.

I'll be pushing some changes to fix that issue.

mxgrey:
I've removed `[WIP]` because I believe this is PR complete, and I've successfully tested it on my machine.

If others can test this out on their own projects, it would be much appreciated, **especially on Windows machines, because those have some special cases**.

Testing might be a bit tricky, because by default cmake will choose the find-module, which will skip over using this new config-file. To force your project to search for the config-file, you can add the `CONFIG` argument to `find_package(~)`. Here's an example:

```
set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS TRUE)
find_package(OpenSceneGraph REQUIRED COMPONENTS osgUtil osgDB osgText osgGA osgFX osgSim osgViewer CONFIG)
```

The `set_property(~)` call may be necessary for 64-bit builds, because cmake will not search in `lib64` by default. It may also be necessary to use `cmake-3.12.4`, because it seems that earlier versions might not search `lib64` no matter what.

### pkg-config

One thing about this PR that I expect might be controversial is that I've removed the litany of `openscenegraph-<component>.pc.in` files and replaced them with a generic template and I'm having them automatically generated. Besides improving maintainability, this change revealed that two .pc files were missing: `openscenegraph-osgUI.pc` and `openscenegraph-osgPresentation.pc`. It also uncovered that `openscenegraph-osgDB.pc` was missing its dependency on `osgUtils`.

My motivation for this change was having a way to automatically generate the `openscenegraph-<component>-import-targets.cmake` scripts, and to ensure that there is a single source of truth for the dependencies of each component. Having manually-written `pkg-config` files represents a second source of truth and a likely vector for build errors.

mxgrey:
Also, the config mode should work for providing the old fashioned `XXX_[LIBRARIES|INCLUDE_DIRS]` variables, as well as new imported targets that you can link against.

The imported targets looks like `osg3::<component>`, for example `osg3::osg` for the core library and then `osg3::osgDB`, `osg3::osgUtil`, `osg3::osgGA` etc.

All of the targets will also be wrapped up in `OpenSceneGraph_TARGETS` for convenience. You can use `target_link_libraries(tgt [PUBLIC|PRIVATE] ${OpenSceneGraph_TARGETS})` to link your target against all of them in one stroke.

Note that linking against the imported targets will also provide your target with the installed include directory information, so you no longer need to use `OPENSCENEGRAPH_INCLUDE_DIRS` or any variation of it.

mxgrey:
> Configuring on Windows I get configure errors.

Those errors make it sound like the configuration script isn't making it to [this line](https://github.com/openscenegraph/OpenSceneGraph/pull/660/files#diff-4722b1eb26539146620766ffe15197d1R27) on Windows (from there, it should either export the target by calling [this line](https://github.com/openscenegraph/OpenSceneGraph/pull/660/files#diff-4722b1eb26539146620766ffe15197d1R33) or [this line](https://github.com/openscenegraph/OpenSceneGraph/pull/660/files#diff-4722b1eb26539146620766ffe15197d1R161).

Does the configuration/build of OpenThreads get skipped on Windows? If so, we'll need to make it so that OpenThreads is not added to `TARGET_LIBRARIES` [like this](https://github.com/openscenegraph/OpenSceneGraph/pull/660/files#diff-c8d25a1523dcf06db77245c1f6037a6cR47) when configuring on Windows (or when configuring in any other circumstance where OpenThreads is not needed).

mxgrey:
I realized that the OpenThreads configuration starts forking [here](https://github.com/mxgrey/OpenSceneGraph/blob/23f680ef4a693862271d5adf777e6edb2c8057f5/src/OpenThreads/CMakeLists.txt#L87). I didn't notice it because my IDE hides files that aren't touched during configuration on my machine (for better or worse).

In the latest commit I've added target exporting for WIN32. But I'm still not testing anything on Windows yet, so there may be more errors remaining.

sbrkopac:
The changes worked for me. I added a modern cmake example and opened https://github.com/mxgrey/OpenSceneGraph/pull/1. I elected to write a new one as to keep the old one intact since it will still need have potential support.

openscenegraph:
Hi @mxgrey, I've just tested out your cmake config changes, impressive effort :-)

When doing an install on my Kubuntu 18.04 64bit system everything compiles and installs to the place I point CMAKE_INSTALL_PREFIX to, and if I build the osgCMakeExampleModern I have to specify OpenSceneGraph_DIR to point to the lib64/cmake/OpenSceneGraph directory, I can't just leave CMake to find it from specify the the same directory that I set the CMAKE_INSTALL_PREFIX too.  

Even if I install the OSG to default /usr/local the find(OpenSceneGraph ... CONFIG)  still won't find the configuation script without me specifying the lib64/cmake/OpenScneGraph directory.

With the VSG project I'm just using CMake defaults for lib directories so it builds and install the libs and the cmake config directory into /usr/local/lib/cmake directory or the ${CMAKE_INSTALL_PREFIX}/lib/cmake directory,  When I then use a Cmake find I don't need to speciying the VSG_DIR install path when installed to /usr/local, or just need to specify the CMAKE_INSTALL_PREFIX directory otherwise.  This makes it easier than is currently possible with the new OpenSceneGraph config.

I'm guessing this is an issue with the find script not working with defaults is done to the OSG installing to lib64 rather than lib.  I haven't looked into this yet.  FYI, the OSG's CMake script was written by other contributors, so I haven't need to dive into the final details of every aspect of it.  I don't recall the logic behind the lib64 install and will have to look up the CMake install scripts to see where this is being set and for what purpose.  It could be that we should just be defaulting to installing in lib rather lib64.

I thought I'd raise this to get back what your impressions are. It would be nice to have things work more simply as I've been able to do with the VSG, and am open to changing default install paths as long as there isn't other fallout from it, which there could well be, so it may be just a case of having to put up with there extra hurdle just to retain backwards compatibility.

Thoughts?
Robert.

mxgrey:
> I'm guessing this is an issue with the find script not working with defaults is done to the OSG installing to lib64 rather than lib.

That's exactly right. Since it's being installed in `lib64`, there are two conditions that need to be met for cmake to find it like normal:

1. You need to be using cmake 3.12 or higher
2. You need to put the line `set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS TRUE)` some time before your call to `find_package(OpenSceneGraph ...)`.

Obviously this is not desirable, but it's a direct consequence of installing the library into `lib64` instead of `lib` or `lib/<arch>`.

It does make sense to install the library into an architecture-specific directory for multi-architecture platforms (and most PC platforms *are* multi-architecture), but cmake isn't especially well-suited for handling the case where the architecture is embedded in the library directory name. Older versions of cmake are directly compatible with a pattern like `lib/<arch>` instead of `lib<arch>`. See the line [`<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/`](https://cmake.org/cmake/help/v3.0/command/find_package.html).

If you're open to changing the install location, we should be able to make it so that

1. cmake's usual rules find the library with no extra effort, and
2. the old find-module that gets distributed with cmake can work again for the latest master, while
3. the library still goes to an architecture-specific directory when installed system-wide to `/usr/`

The built-in standard way of doing this with cmake is to use [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). If you're okay with making the install location more standard, I'm happy to make this change.

openscenegraph:
Thanks for the info.  I'm keen to get things working better out of the box.  It would probably be best to two PR's one for config support - I'll merge this one right now.  Then we can sort out the lib64/install path next, and welcome help with this.

Issue #655, opened 2018-11-05
openmw SEGV with OSG 3.6.3


mimi1vx:
openmw issue - > https://gitlab.com/OpenMW/openmw/issues/4692

openmw engine crash with signal 11 if is used latest stable OSG.

crashlog : https://pastebin.com/vVmuHFdp

openmw community bisected commit which caused this problem to: https://github.com/openscenegraph/OpenSceneGraph/commit/eee5d5482e41290e7c210273f8c6bae489cedb76

robertosfield:
I suspect the OSG commit just reveals an underlying bug elsewhere rather than directly causing the cash. Where the actual problem might be will only be resolvable by having a small test program that can reproduce the issue.

Could this be a threading issue?  I'm not familiar with OpenMW, is it being run multi-threaeded?  What happens when the the osgViewer part uses SingleThreaded threading model.  If it is running multi-threaded then it would be worth trying a SingleThreaded to see what happens.  If it resolves the crash then it's a hint that it could be a threading issue.

At this point there isn't anything I can do to help resolve the issue as the crash is too open ended.

akortunov:
> If it resolves the crash then it's a hint that it could be a threading issue.

I tried to use the SingleThreaded model, and the game still crashes. The only difference is the stack trace (crash happens in the main thread).

robertosfield:
Perhaps this is some form of dangling pointer issue then.  A stack trace
for each of the different types of crashes (single threaded vs
multi-threaded, might help show a pattern.

akortunov:
Here the logs:
[openmw-crash.multi.log](https://github.com/openscenegraph/OpenSceneGraph/files/2552635/openmw-crash.multi.log)
[openmw-crash.single.log](https://github.com/openscenegraph/OpenSceneGraph/files/2552636/openmw-crash.single.log)

Note: the game can crash on different lines of the osg::GLBufferObject::compileBuffer() function.

Also I get some such warnings, which I do not have with older OSG versions:
```
Warning: detected OpenGL error 'invalid operation' after applying attribute Viewport
```
but it may be a completely separate issue.

mp3butcher:
stack trace would be more relevant to debug

akortunov:
OSG_VERTEX_BUFFER_HINT=VAO does not fix the crash.

Here are logs with OSG_NOTIFY_LEVEL=INFO and OSG_NOTIFY_LEVEL=DEBUG, from game start to moment of crash:
[...]

PR #668, opened 2018-12-01
Fix an issue with VAS

mp3butcher:
solving https://github.com/openscenegraph/OpenSceneGraph/issues/655
a test could be removed if we consider negligible the cost of calling glBindVertexArray(0) but i'm not so sure about that

### DIFF
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -55,7 +55,7 @@
 #endif
 
 
-// #define INLINE_DRAWABLE_DRAW
+#define INLINE_DRAWABLE_DRAW
 
 namespace osg {
 
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -566,7 +566,7 @@ inline void Drawable::draw(RenderInfo& renderInfo) const
 
         State::SetCurrentVertexArrayStateProxy setVASProxy(state, vas);
 
-        state.bindVertexArrayObject(vas);
+        state.bindVertexArrayObject();
 
         drawInner(renderInfo);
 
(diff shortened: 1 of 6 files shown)


akortunov:
Unfortunately, in its current state this PR does not ifx the issue - I still get crashes with the same stack trace.

mp3butcher:
I don't understand what you're doing.
When I reproduced your bug,  the game didn't even launch the intro....With this patch I played MW all the night in debug whitout any problem

mp3butcher:
No I don't understand why it fails with you... I just check all my stuff and there's no mistake ...
Please try again

robertosfield:
Thanks for looking into this @mp3butcher, I have done a first pass review of the changes and my initial response is that it looks like there is a bug lurking somewhere, it could be in the OSG, or it could be in 3rd party code.  From the changes I couldn't easily spot what the "bug" on the OSG that the changes are attempting to address.  If you've found a bug on the OSG could you point out the failure mechanism, or point me in the direction of an example that illustrates the bug.

robertosfield:
I can't just merge random changes that may or may not fix and issue that
may or may not be a real problem in the OSG code base.  I need to be
understand what the actual bug is and how the changes address it,  This
particular set of changes doesn't feel coherent, it feels like a collection
of different hacks that somehow do something.  If you submitted these
changes before I rejected them it was probably for this reason.

Each time I have merged stuff with making sure it right is has come back to
bite us with bugs being introduced.  So it might fix things for the issue
you've seen, but it might just be masking a problem elsewhere.

Once I understand what is actually going wrong I'll be able to judge the
validity of fix, or come up with the most appropriate fix myself.

mp3butcher:
These aren't random changes
If I haven't follow you on with some changes on VAS, there's a reason...I just can't remember.

mp3butcher:
oups rolling back I can't reproduce the crash...!?
I don't know what happen I close the pr until i understand something

robertosfield:
Thanks for perceiving with this issue.  Fingers crossed 3.6.3 is just fine.

Issue #674, opened 2018-12-13
delete of shared GLObjects when Drawable is released


mp3butcher:
Introduced when trying to fix VAO leak  https://github.com/openscenegraph/OpenSceneGraph/pull/587#issuecomment-412280987
commit engaged:
https://github.com/openscenegraph/OpenSceneGraph/commit/eee5d5482e41290e7c210273f8c6bae489cedb76
To reproduce
```
osgViewer::Viewer viewer;
if(wantmorebug) //rise another crash at program end with OSG_VERTEX_BUFFER_HINT=VAO
    viewer.setThreadingModel(osgViewer::Viewer::DrawThreadPerContext);
osg::Geometry * geom= osg::createTexturedQuadGeometry(osg::Vec3f(-0.5,-0.5,-0.5),osg::Vec3f(1,0,0),osg::Vec3f(0,0,1));
geom->setUseVertexBufferObjects(true);
osg::Geometry *  geom2=(osg::Geometry*)geom->clone(osg::CopyOp::SHALLOW_COPY);
   osg::Group *gr=new osg::Group;
   gr->addChild(geom);
   gr->addChild(geom2);

   viewer.setSceneData(gr);
   viewer.realize();
   viewer.frame();//for the fun as GLObjectVisitor already have done its stuff

    //assume contextid==0
    int contextid=   0;
    OSG_WARN<<"BufferObject uploaded suppose ok at frame 0:"<<
             ( geom->getVertexArray()->getVertexBufferObject()->getGLBufferObject(contextid) ?
                   "OK" : "BAD!!!!!!!!!" )<<std::endl;
    //remove  geom2 (no more ref to it)
    gr->removeChild(1);
    OSG_WARN<<"BufferObject uploaded suppose ok at  preframe 1:"<<
              ( geom->getVertexArray()->getVertexBufferObject()->getGLBufferObject(contextid) ?
                    "OK" : "BAD!!!!!!!!!" )<<std::endl;
    viewer.frame();
    OSG_WARN<<"BufferObject uploaded suppose ok at postframe 1:"<<
              ( geom->getVertexArray()->getVertexBufferObject()->getGLBufferObject(contextid) ?
                    "OK" : "BAD!!!!!!!!!" )<<std::endl;
viewer.run();
```

mp3butcher:
commenting out Geometry::releasGLObject in ~Geoemtry seams enough but ...I'd like to be sure

mp3butcher:
[...]

PR #675, opened 2018-12-13
Prevent geometry to release shared GLObject when destroyed

mp3butcher:
fix for https://github.com/openscenegraph/OpenSceneGraph/issues/674

### DIFF
--- a/src/osg/Geometry.cpp
+++ b/src/osg/Geometry.cpp
@@ -83,7 +83,6 @@ Geometry::Geometry(const Geometry& geometry,const CopyOp& copyop):
 Geometry::~Geometry()
 {
     _stateset = 0;
-    Geometry::releaseGLObjects();
 }
 
 #define ARRAY_NOT_EMPTY(array) (array!=0 && array->getNumElements()!=0)


emminizer:
For reference:

Some of the original discussion is on https://github.com/openscenegraph/OpenSceneGraph/pull/587

The original discussion on the leak was covered in the thread https://www.mail-archive.com/osg-users@lists.openscenegraph.org/msg77211.html "VAO Resource Leak in OSG 3.6.2" starting 2 Aug 2018.

mp3butcher:
> This PR would be best described as a workaround with unintended side effects....
@robertosfield 
 The side effects come from the design not from this fix. 
It was your choice to do as it is, you've unintentionnally added this bug I think as nothing motive Geometry::releaseGLobject
Please watch you commit https://github.com/openscenegraph/OpenSceneGraph/commit/eee5d5482e41290e7c210273f8c6bae489cedb76

vsg-dev:
A subtle issue with C++ is that you can't call subclass implementations from the destructor of base class, so you have to do this calls in the destructor of the subclass.  This is why the releaseGLObjects() is called from the Geometry destructor.  

Issues with handling sharing of arrays might be a bug, might be a design limitation.  

All this extra complexity comes about due to the need to support VAO under GL3 core profile.  I think we are mostly the there for good support, but the OSG can be used in so many different ways it can be hard to get it right first time.

mp3butcher:
mmh I have a coredump on last vao with 3.6 will dig

vsg-dev:
C++ lesson.... 

The destructor of a subclass gets called first.  This destructor has access to all member variable and member functions that subclass, including virtual functions of that subclass that are provided by that subclasses vtable.

The destructor of the parent class gets called next.  This destructor has no access to the the subclass members or methods, or virtual functions as the references to vtable for the subclass are no longer strictly available (implementation wise they may be but it's not something you should ever rely upon.)

--

So... in this case to call Geometry::releaseGLObjects() in the destructor it must be called by the Geometry destructor, we can't rely upon the Drawable destructor to do it.

mp3butcher:
Again more confusion for me:
1/We don't  geometry to release GLObjects as they can be in used by other
2/Drawable::VAS are an exception so we call **not polymorphic** Drawable::releaseGLObject to release them (it release displaylist to but that's not a problem)
3/if I understand what you just write: you are afraid this don't work because releaseGLObject is virtual, do you agree to make them add concrete Drawable::notvirutalreleaseGLObjects but it's seems paranoia to me..:/with gcc at least explicit call to Drawable::releaseGLObject in drawable destructor don't pose problem
The fix seems to work (all VAO released at the end) but there is a crash at program ends with the  thread model DRAWTHREADPERCONTEXT

mp3butcher:
ok i think i see your point 
Drawable:::releaseGLObject may call deleted polymorphics (it's possible)

PR #676, opened 2018-12-14
 Add an interface of osg::Vec3d*

Dyuzz:
In some cases, vertex coordinates will exceed 6-digit valid digits, such as UTM coordinate system, need to use double type,so I add osg:: vec3d* interface, and use requiring overload operator() (vec3d x, vec3d y, vec3d z, bool raw). 
And it solves a bug: when STL is generated, the system crashes if vec3d type data is output

### DIFF
--- a/include/osg/TriangleFunctor
+++ b/include/osg/TriangleFunctor
@@ -46,14 +46,9 @@ public:
         _vertexArraySize=0;
         _vertexArrayPtr=0;
         _vertexArrayPtr_3d = nullptr;
-        _modeCache=0;
-        _treatVertexDataAsTemporary=false;
     }
 
     virtual ~TriangleFunctor() {}
-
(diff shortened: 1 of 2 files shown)


openscenegraph:
I don't feel it's appropriate to roll out Vec3dArray support across the OSG, it'll double the complexity of much of the code for a niche usage case that may well be inappropriate to start with.  

Most graphics hardware are design around float maths and are brought to the knees when you pass them double arrays, and may even downcast to float on each new frame to hide the lack of hardware support and this then can result in appalling performance.

The correct way to manage UTM systems in scenegraphs/OpenGL is to use the OSG's support for double matrices in the Camera and Transform nodes so that it accumulates the modelview matrix in doubles and only cast to float matrix when passing to OpenGL.  These scheme allows one to manage whole earth database with good precision.   This is how VirtualPlanetBuild/osgEarth all provide good large database support and can seamlessly go from space to ground level without precision issues.

The best place to discuss these approaches it to raise the topic on osg-users mailing list/forum.  Search for the topic too as we've discussed the techniques many times.

PR #679, opened 2018-12-19
setup camera Read/Draw Buffer

mp3butcher:
fix flickering shadows in SingleThread


openscenegraph:
I also have merge this PR with master.

PR #681, opened 2018-12-19
make technique overridable

mp3butcher:


vsg-dev:
Ideally I'd like to keep ABI compatibility for the 3.6.  Changing private to protected can change ABI on some compilers so we have to think twice about doing this on the 3.6 branch.

It may be that for bug fixes we have to change the ABI anyway, and if this is the case then I don't have any issue with this PR.  This PR will be fine for master without any delay though as there isn't any need for ABI maintenance.

Issue #664, opened 2018-11-21
Read GL_R32F texture failed


0871087123:
https://github.com/openscenegraph/OpenSceneGraph/blob/85504b45f6229c2695bc109d771045fcb10fc1e0/src/osg/Image.cpp#L425

I failed on try to read a 32bit float texture which has only one channel,  this is my code:
```
            state->applyAttribute(_texture);
            osg::ref_ptr<osg::Image> img = new osg::Image;
            img->readImageFromCurrentTexture(ri.getContextID(), true, GL_FLOAT);
```
I found a invalid enumerant error report after call glGetTexImage
I checked the code
glGetTexImage should have pixel format GL_RED as one of its' input param, but actually is GL_R32F， and it's an invalid enumerant
I found it should be computed here, from GL_R32F to GL_RED

Also , the desgin of readImageFromCurrentTexture confused me. It implemented as check every texture binding in order, it leads to if i has a cube texture binding, and currently applied a 2d texture, it will still read from cube texture. so i use this interface like this:
```
            glBindTexture(GL_TEXTURE_CUBE_MAP, 0);
            state->applyAttribute(_texture);
            osg::ref_ptr<osg::Image> img = new osg::Image;
            img->readImageFromCurrentTexture(ri.getContextID(), true, GL_FLOAT);
```
But it's not a good idea unbind cube texture when reading a 2d texture, i just use this method for test.

openscenegraph:
I've tried to piece together what might be going on at your end, best guess at this point is that you are reading from a texture that as an internal format that computePixelFormat() isn't yet supporting so is falling back to just returning the input formatted, and the later GL call isn't handling this particular internal format.  

I'm just guessing though as I don't have a test program, any knowledge of your wider application usage, what platform you are working on, what OSG version, what hardware....

Could you create a small example that illustrates your usage case that is failing.  Taking an existing OSG example and modifying would be a good place to start.  Once we have the usage cause that is failing we can work out whether it's a usage error or an internal bug, if the later we can then fix it and use the program to confirm the issue is fixed.

Thanks.

0871087123:
> I've tried to piece together what might be going on at your end, best guess at this point is that you are reading from a texture that as an internal format that computePixelFormat() isn't yet supporting so is falling back to just returning the input formatted, and the later GL call isn't handling this particular internal format.
> 
[...]

PR #683, opened 2019-01-04
Readtexture bug fix

0871087123:
This is my fix for issue #664  
I tested texture format of GL_Rxxx such as GL_R16F, GL_R32F, GL_R8I and some else.
Also tested texture format of GL_RGxxx such as GL_RG16F, GL_RG32F, GL_RG16I, etc.
The test program is in examples/osgrtt , if texture is readed, then will no error code reported , and also can read the pixel value when model show  left bottom corner of view.
Test program is template from osgdeffered, and i left the title comment unchanged.


src/osg/Image.cpp
```diff
@@ -535,9 +579,30 @@ unsigned int Image::computeNumComponents(GLenum pixelFormat)
         case(GL_ALPHA32UI_EXT): return 1;
         case(GL_ALPHA16F_ARB): return 1;
         case(GL_ALPHA32F_ARB): return 1;
+        case(GL_R16F): return 1;
         case(GL_R32F): return 1;
-        case(GL_RG): return 2;
```
mp3butcher: any reason for this removal?!

openscenegraph:
Thanks for the changes.  Could you break this PR into two parts, one for the changes to Image.cpp and another for the new example.  This will make it clearer for others to track, and easier for me to cherry-pick the fixes into the 3.6 branch so it can be part of the upcoming 3.6.4 release.

0871087123:
> Thanks for the changes. Could you break this PR into two parts, one for the changes to Image.cpp and another for the new example. This will make it clearer for others to track, and easier for me to cherry-pick the fixes into the 3.6 branch so it can be part of the upcoming 3.6.4 release.

@openscenegraph  I changed this PR changing Image.cpp only, and another branch contains 
 one example testing this change still stay in my repo. Should i make a PR on example too?

openscenegraph:
Thanks Kent.  I will go ahead and merge the Image.cpp changes to master, then have a look at if the new additions require additionally #define for older gl.h headers, then merge this with the 3.6 branch so it can be part of 3.6.4.

PR #685, opened 2019-01-09
Fixed Memory Leak and some warnnings.

AlexiaChen:


0871087123:
And , it may be not suitable to comment here. I'm from Kunming, too. Can i have your QQ number or some  else? It's interesting meeting people from the same place and have same interesting in OSG.

src/osgPlugins/3ds/ReaderWriter3DS.cpp
```diff
@@ -1338,6 +1338,7 @@ ReaderWriter3DS::StateSetInfo ReaderWriter3DS::ReaderObject::createStateSet(Lib3
             unsigned char* orig_img_data = (unsigned char*)osg_image->getDataPointer();
             int n = osg_image->s()*osg_image->t()*4;
             unsigned char* img_data = new unsigned char[n];
+            memset(img_data, 0, n);
```
0871087123: img_data seted in for loop below, is it necessary setting to zero here?

src/osgPlugins/gdal/ReaderWriterGDAL.cpp
```diff
@@ -557,14 +557,20 @@ class ReaderWriterGDAL : public osgDB::ReaderWriter
                     (unsigned char *)imageData,
                     osg::Image::USE_NEW_DELETE);
 
-                if (texCoordRange) image->setUserData(texCoordRange);
+                if (texCoordRange) 
+                {
+                    image->setUserData(texCoordRange);
+                    delete texCoordRange;
```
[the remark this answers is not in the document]
openscenegraph: This change introduces a dangling pointer and with it a bug, it's very much a step backwards.  As Kent suggests the OSG has reference counting that will ensure texCoordRange will be deleted automatically when the image gets deleted. 

What tool are you using to "detect" leaks?  It may well be that it's reporting false positives, or perhaps you are just interpreting things incorrectly as the changed code path doesn't have a leak.

As general tightening up of the code the appropriate thing to do is use a ref_ptr<<osgDB::ImageOptions::TexCoordRange> rather than C pointer.  I have also added a protected destructor for the TexCoordRange to prevent users for incorrectly attempting to delete a reference counted object (most other classes in the OSG already do this.)

src/osgPlugins/mdl/VVDReader.cpp
```diff
@@ -69,6 +69,9 @@ bool VVDReader::readFile(const std::string & file)
     if (header.magic_number != VVD_MAGIC_NUMBER)
     {
         OSG_NOTICE << "Vertex data file not valid" << std::endl;
+        // Close the file
```
openscenegraph: I've just done a code review of the surround code and it looks like the original author created a ifstream on the heap inappropriately, and should have just created it on the stack. It looks like the original author misunderstood how new and ifstream constructor work too.  Amazing such old code hadn't been picked up as being sloppy before, I can only guess that almost no one uses this niche plugin.

I have change the code to use a stack created ifstream and checked this in to master.  I'll also merge this with the 3.6 branch.

AlexiaChen:
@0871087123 QQ: 442499771

src/osgPlugins/mdl/VVDReader.cpp
AlexiaChen: :+1:

AlexiaChen:
I'm not interesting in OSG, Frankly. I just test a bunch of static code analysis tool  : )

AlexiaChen:
@0871087123  I'm not interesting in OSG, Frankly. I just test a bunch of static code analysis tool : )

src/osgPlugins/mdl/VVDReader.cpp
AlexiaChen: Huh.......   I just find you are using Coverity  to find potential bugs in this project.   And, Coverity did not find this file resource leak? Coverity may have more powerful feature compared to cppcheck.

PR #687, opened 2019-01-11
remove const on callback operation

mp3butcher:


mp3butcher:
@robertosfield 
could you consider adding modification to vascallback and a destroyVAScallback as i've done here
https://github.com/openscenegraph/OpenSceneGraph/pull/677/files
sharing vas require to control both creation and destructyion....

robertosfield:
There is no way to look at that PR to be able to work out the minimal set
of changes you think are required.  I have no interest in spray of changes
that touch dozens of files just for a very niche feature that only one user
has so far requested.

openscenegraph:
The callback method is called from a const draw method so it needs to be const.  

The point of draw methods being const is the OSG is design to facilitate mulit-threaded draw traversals, for const objects this is straight forward, for non const objects you have to introduce a mutex around contended data structures.  It can be done but it's far slower than just design code to be const.

If you really want to non const sutff instead this callback then cast away constness and make sure it's thread safe.  And if you think this is an extra level of work that you don't want to do, then good, the const is there to make sure you think twice about doing something that is potentially unsafe or inefficient.

mp3butcher:
> The callback method is called from a const draw method so it needs to be const.
> 

Drawable need to be const not the callback itself...

robertosfield:
On Sat, 12 Jan 2019 at 13:10, Julien Valentin <notifications@github.com> wrote:
> The callback method is called from a const draw method so it needs to be const.
>
> Drawable need to be const not the callback itself...

The callback is member variable that is called from a const method,
this would require the callback method being called to be const.

Robert.

PR #690, opened 2019-01-14
ObjectCache expiry/DatabasePage locking when busy to prevent deletion of referenced object.

Zaretto:
This fixes a problem when the ObjectCache is in the process of expiring an object that is simultaneously has been newly referenced from the DatabasePager thread resulting in a "Deleting still referenced object".

Problem diagnosis; when the DatabasePager has loaded from the object cache after releasing the objectCacheMutex a reference is taken (ref_ptr); as the mutex is released the object cache then expires the object at the same time as the ref_ptr is incremented in the database pager thread. This happens because the object meets the expiry time condition based on the start of the frame. So there exists a condition where the object may be deleted from the cache when it has just been referenced and this will often lead ref_ptr detecting this, warning with "Deleting still referenced object" and ultimately probably a segfault.

------------
This fix uses a ReadWriteMutex (with a new method writeTryLock) to control the expiry process. The database pager is considered to be a reader, of which there can be many (tested with 1 and 20), the ObjectCache expiry method is considered to be a writer. Only when there are no readers will the object cache expiry be able to remove items from the cache. As the expiry uses a writeTryLock expiry will be locked when there are ongoing DatabasePager loads. This will defer the expiry of objects to a later frame when the DatabasePager has finished its work.

Testing this is tricky because of the small window of opportunity; however by adding some debug print statements it has been verified that the locking is working as expected and a 60.5h test (previous failed after 40h) exhibited no problems.

NOTES:

1. Possible side effects of this fix could be an unusually large number of items being expired when the mutex is released if held for an extended period (probably never going to be a problem).

2. Lockout causes no noticeable difference to frame rate, and no frame pauses.

2. In a 3 hour test the longest expiry lockout was 549 frames, average lockout 23.4 frames. mean 24, median 10.

(cherry picked from commit b596d59410cbc7e176751f14ee77634e0de9dc1a)

### DIFF
--- a/include/OpenThreads/ReadWriteMutex
+++ b/include/OpenThreads/ReadWriteMutex
@@ -61,6 +61,11 @@ class ReadWriteMutex
             return _readWriteMutex.lock();
         }
 
+        virtual int writeTryLock()
+        {
+            return _readWriteMutex.trylock();
+        }
+
         virtual int writeUnlock()
         {
             return _readWriteMutex.unlock();
--- a/include/osgDB/Registry
+++ b/include/osgDB/Registry
@@ -15,6 +15,7 @@
 #define OSGDB_REGISTRY 1
 
 #include <OpenThreads/ReentrantMutex>
+#include <OpenThreads/ReadWriteMutex>
 
 #include <osg/ref_ptr>
 #include <osg/ArgumentParser>
@@ -119,6 +120,7 @@ class OSGDB_EXPORT Registry : public osg::Referenced
 
         /** get a reader writer which handles specified extension.*/
         ReaderWriter* getReaderWriterForExtension(const std::string& ext);
+        OpenThreads::ReadWriteMutex *getObjectCacheExpiryMutex(){return &_objectCacheExpiryMutex;}
 
         /** gets a reader/writer that handles the extension mapped to by one of
           * the registered mime-types. */
@@ -610,6 +612,7 @@ class OSGDB_EXPORT Registry : public osg::Referenced
         osg::ref_ptr<WriteFileCallback>     _writeFileCallback;
         osg::ref_ptr<FileLocationCallback>  _fileLocationCallback;
 
+        mutable OpenThreads::ReadWriteMutex _objectCacheExpiryMutex;
         OpenThreads::ReentrantMutex _pluginMutex;
         ReaderWriterList            _rwList;
         ImageProcessorList          _ipList;
--- a/src/osgDB/DatabasePager.cpp
+++ b/src/osgDB/DatabasePager.cpp
@@ -847,6 +847,7 @@ void DatabasePager::DatabaseThread::run()
 
         if (databaseRequest.valid())
         {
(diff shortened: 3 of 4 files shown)


openscenegraph:
Thanks for the work investigating the issue.  I've done a review of the PR but it doesn't feel like a resolution of a bug, more a reduction in parallelism that may co-incidentally reduce the likely-hood of failure.  I don't like having the extra mutex in the code without definite idea of the actual bug and that it warrants the extra over-head.

I have just read a post @LaurensVoerman on the osg-users mailing list/forum suggesting that simgear wasn't using the getRefFromObjectCache() methods that is should be.  I fully agree with Lauren's observation and this is most likely the cause of the crash in FlightGear.  I would recommend changing all FlightGear/Simgear code to use the Ref versions of the various read methods rather than use the C* version. The new Ref version all came about to address thread safety as just passing around C pointers has been proven to be unsafe for multi-thread apps.  The changes I suggest are along the lines:

diff --git a/simgear/scene/model/ModelRegistry.cxx b/simgear/scene/model/ModelRegistry.cxx
index f788162a..8e739a83 100644
--- a/simgear/scene/model/ModelRegistry.cxx
+++ b/simgear/scene/model/ModelRegistry.cxx
@@ -280,13 +280,17 @@ ModelRegistry::readImage(const string& fileName,
 }
 
 
-osg::Node* DefaultCachePolicy::find(const string& fileName,
+osg::ref_ptr<osg::Node> DefaultCachePolicy::find(const string& fileName,
                                     const Options* opt)
 {
-    Registry* registry = Registry::instance();
-    osg::Node* cached
-        = dynamic_cast<Node*>(registry->getFromObjectCache(fileName));
-    if (cached)
+#if OSG_VERSION_LESS_THAN(3,4,0)
+    osg::ref_ptr<osg::Object> cachedObject = registry->getFromObjectCache(fileName);
+#else
+    osg::ref_ptr<osg::Object> cachedObject = registry->getRefFromObjectCache(fileName);
+#endif
+
+    ref_ptr<osg::Node> cachedNode = dynamic_cast<osg::Node*>(cachedObject.get());
+    if (cached.valid())
         SG_LOG(SG_IO, SG_BULK, "Got cached model \""
                << fileName << "\"");

I wasn't able to compile simgear on my quick attempt to apply this change so will defer to the FlightGear team to follow this up with a general review of this part of FlightGear/Simgear.

I'm going to close this PR as i don't think it's the appropriate fix.  This issue should be resolveable by just changing the FlightGear/simgrear code and will work for OSG-3.4.x, OSG-3.6.x and master as they all have the required Ref methods.  I will add documentation to the OSG's ObjectCache methods to specify that C* method is deprecated and users should use the Ref version instead.

openscenegraph:
@Zaretto, I have checked in the follow docs changes to help point users in the right direction:
https://github.com/openscenegraph/OpenSceneGraph/commit/f223fec92dd0af7173a9782ac1234ce59b2703bc

Issue #694, opened 2019-01-19
crash with new VAO path


mp3butcher:
The following code exhibits crash with new VAO path  with my test model.
Please confirm it with yours
```
#include <osgUtil/MeshOptimizers>
#include <osgGA/TrackballManipulator>
#include <osgGA/FirstPersonManipulator>

#include <osgViewer/Viewer>
#include <osgViewer/ViewerEventHandlers>

#include <osgDB/ReadFile>
#include <osgDB/WriteFile>


class GeomLoaderCB : public osg::Camera::DrawCallback//  osg::NodeCallback
{
public:
    int _thresremoval;int _nbaddedatatime;
    GeomLoaderCB(int thresremoval=1,int nbaddedatatime=1):_nbaddedatatime(nbaddedatatime),_thresremoval(thresremoval) {}
    mutable std::list<osg::ref_ptr<osg::Geometry> > _geoms;
    void setGeometryList(osgUtil::GeometryCollector::GeometryList c) {
        for(auto f : c)
            _geoms.push_back(f);
    }
    virtual void operator () (const osg::Camera&  camera ) const {
        osg::Node * node=const_cast<osg::Camera*>(&camera)->getChild(0);

        if(_geoms.empty())return;
        osg::ref_ptr<osg::Group>  gr = node->asGroup();

        if(gr->getNumChildren()>_thresremoval)
        {
            osg::ref_ptr<osg::Geometry> geom = gr->getChild(0)->asGeometry();
            OSG_WARN<<"removing "<< geom.get()<<std::endl;
            gr->removeChildren(0,1);
            /// not changing anything geom->releaseGLObjects(); 
///geom vas is already null before removal!!!
            return;
        }

        std::list<osg::ref_ptr<osg::Geometry> > ::iterator it= _geoms.begin();
        int cpt=0;
        while(it!=_geoms.end()&&cpt++<_nbaddedatatime ) {
            osg::Geometry *newdr=(osg::Geometry*) (*it)->clone(osg::CopyOp::DEEP_COPY_ALL);
            gr->addChild(newdr);
            OSG_WARN<<"add "<< newdr ->getVertexArray()->getNumElements()<<std::endl;
            it=_geoms.erase(it);
        }
        return;
    }
};

/// This demo reproduce a bug with OSG_VERTEX_BUFFER_HINT=VAO
/// it collecte drawables given in arg then add and remove them at runtime
int main(int argc, char **argv)
{
[...]

PR #695, opened 2019-01-20
fix lazy reuse of vao with the same name

mp3butcher:
https://github.com/openscenegraph/OpenSceneGraph/issues/694
I think i have a lead on that bug:
VA0 with name (says 1) is deleted then another is regenerated with the same  name (1)  but active check of vao in state.bindVAS  don't see they are different, don't do anything and crashes....
I fixed it by forcing state.setGlobalVAS and state.currentVAO on VAO deletion

### DIFF
--- a/src/osg/VertexArrayState.cpp
+++ b/src/osg/VertexArrayState.cpp
@@ -541,6 +541,9 @@ void VertexArrayState::deleteVertexArrayObject()
     {
         VAS_NOTICE<<"  VertexArrayState::deleteVertexArrayObject() "<<_vertexArrayObject<<std::endl;
 
+        _state->setCurrentToGlobalVertexArrayState();
+        _state->setCurrentVertexArrayObject(_state->getCurrentVertexArrayState()->getVertexArrayObject());
+        
         _ext->glDeleteVertexArrays(1, &_vertexArrayObject);
         _vertexArrayObject = 0;
     }


0871087123:
So it means state.bindVAS will not rebind VAO regenerated with same name ? then crashes in the following codes.

0871087123:
Check the code and understand it should be this, bind vao to global after deletion.

openscenegraph:
Making a VSA "current" and associated VBO in the osg::State before it get deleted will just break state tracking.  

The originally code looks correct. The PR introduces bugs for sure, there is no way it's acceptable.

mp3butcher:
@openscenegraph 
Do you see the problem i describe at least?

openscenegraph:
I have a fix I'm testing for the dangling VAS pointer bug.

You example and dataset don't crash either way for me, so can't confirm that this is fixed, but at least the dangling VAS pointer will be fixed.

As for close issue, well if the Issue is poorly put together enough for it itself has bugs that require fixing then it's what am I to do?  Go fixing bugs in your code or push back and get you to submit stuff that is ready for testing.  Point out errors multiple times and see them still popping up is not good.

BTW, the BIGCITY.ive is a poorly constructed scene w.r.t performance, thee number of StateSet is way excessive and a flat group contains all the nodes isn't efficient.  A scene has a far higher draw overhead than it should for such a simple scene.   Avoiding lots of duplicate State and creating a quad tree rather than a flat group as first things to do to improve performance with datasets like this.  This is an important thing to do before you start looking at any other ways of getting better performance as you can get totally misleading results when testing against a poorly constructed scene graph.

PR #696, opened 2019-01-21
force VAO to 0 at VAS::deleteVAO to avoid reuse of the same name

mp3butcher:
**When a vao is deleted and a new one is regenerate, they may have the same name and so pass as the same for the active check in State::bindVertexArrayObject and crash**

To reproduce the bug use material provided in this issue https://github.com/openscenegraph/OpenSceneGraph/issues/694

### DIFF
--- a/src/osg/VertexArrayState.cpp
+++ b/src/osg/VertexArrayState.cpp
@@ -540,7 +540,7 @@ void VertexArrayState::deleteVertexArrayObject()
     if (_vertexArrayObject)
     {
         VAS_NOTICE<<"  VertexArrayState::deleteVertexArrayObject() "<<_vertexArrayObject<<std::endl;
-
+        _state->setCurrentVertexArrayObject(0);
         _ext->glDeleteVertexArrays(1, &_vertexArrayObject);
         _vertexArrayObject = 0;
     }


mp3butcher:
I'm not sure it's the most fitted way to fix this issue, but please let this pr open until problem's fixed

openscenegraph:
If there is a problem this "fix" certainly isn't the problem.  It really looks like you are randomly hacking without any clear understanding of what may be a problem.  My guest guess is that the State::CurrentVertexArrayState() is being invalidated by scene graph deletion.  The proposed "fix" is idiotic - it only makes sense to reset the State value is the VAS being cleaned is the one that State has a handle to.  I can't believe you'd generate a PR for such poorly thought "fix".  This is just wasting all our time.

The ssue #694 is a good example of how not to use the OSG.  You shouldn't be adding/removing scene graphs in a camera draw callback.  You shouldn't be adding/remving objects from the scene graph when a swtich or node mask works far better.  You should never use an unguraged as*().  This is only a starter for the problems in the Issue #694.

It's really hard to take any of this seriously.  The Issue is bad coding, the PR is yet another hacky submission that has no place in the OSG.

Issue #697, opened 2019-01-21
[Fermi/NV390.87 only?] newly generated VAO having same GL name as a deleted one can lead to crash


mp3butcher:
When a vao is deleted and a new one is regenerated, they may have the same name and so pass as the same for the active check in State::bindVertexArrayObject and crashes
**Driver policy dependant: with nv390.87 vao name deleted is reused right away on next glGenVertexArrays**
_On Ubuntu NVIDIA 390.87/GT440 [GT440 apitrace](https://drive.google.com/file/d/16sgWUPlRAYUKcGPZtJ0JZUyEq8IyDZA6/view?usp=sharing)
Crash doesn't happens with Mesa 18.0.5/Intel Chipset ([trace](https://drive.google.com/file/d/1kDaQ5pFnGc23wVBk1Vwi3PTz2aJLL1Vh/view?usp=sharing) : names doesnt seems to be reused right away)

To reproduce use the following code (if you need a sample dataset use my [test sample]( https://drive.google.com/file/d/0BxIH-jcsgYDdTG5ha21HZE1jX1E/view?usp=sharing) )
and run it with such a command line :
`OSG_VERTEX_BUFFER_HINT=VAO exename BIGCITY.ive --remove 2 --add 1 `

NB: with OSG_VERTEX_BUFFER_HINT=VBO it doesn't lead to crash

```
#include <osgUtil/MeshOptimizers>
#include <osgGA/TrackballManipulator>
#include <osgGA/FirstPersonManipulator>

#include <osgViewer/Viewer>
#include <osgViewer/ViewerEventHandlers>

#include <osgDB/ReadFile>

class GeomLoaderCB : public  osg::NodeCallback
{
    int _thresremoval; int _nbaddedatatime;
    std::list<osg::ref_ptr<osg::Geometry> > _geoms;
public:
    GeomLoaderCB(int thresremoval=1, int nbaddedatatime=1):_nbaddedatatime(nbaddedatatime), _thresremoval(thresremoval) {}

    void setGeometryList(const osgUtil::GeometryCollector::GeometryList& geomlist) {
        for(auto geom : geomlist)  _geoms.push_back(geom);
    }

    virtual void operator()(osg::Node* node, osg::NodeVisitor* nv) {
        osg::Group*  gr = node->asGroup();
        if(!gr || _geoms.empty() ) return;

        if(gr->getNumChildren()>_thresremoval)
        {
[...]

PR #699, opened 2019-01-22
fix bug with Fermi NV390.87 VAO name reutilization policy

mp3butcher:
fix proposal for https://github.com/openscenegraph/OpenSceneGraph/issues/697
state.resetCurrentVertexArrayStateOnMatch(vas) can be added in different places:

-    in createVAS as in the pr
-    in Drawable::draw and compileGLObjects (after createVAS call)
    What solution would you find the more acceptable?

### DIFF
--- a/include/osgParticle/ParticleSystem
+++ b/include/osgParticle/ParticleSystem
@@ -265,7 +265,7 @@ namespace osgParticle
           * for all graphics contexts. */
         virtual void releaseGLObjects(osg::State* state=0) const;
 
-        virtual osg::VertexArrayState* createVertexArrayStateImplemenation(osg::RenderInfo& renderInfo) const;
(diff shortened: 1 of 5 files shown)


openscenegraph:
Resetting the current VAS/VAO values in State after generating a VAO is nonsensical.  If there is a bug it'll be another place than the code you are modifying.

The code now resets the VAS/VAO values in State on destruction of the VAS's that hold the GL object name for the VAO.  If this isn't resolving the problem then there must be something else going amiss.

Could it be that your driver is binding the VAO at the same time as it's generated?  The two steps should be separate.

I have checked in a fix for the typo.  The right way to do PR's like this is to one for each change.  The typo change is fine.  The other changes are dubious.

PR #701, opened 2019-01-23
fix bug with NV390.87 VAO name reutilization policy

mp3butcher:
citing @openscenegraph 
> If the VAS is always destructed after the call to deleteVertexArrayObject() then it will be safe to just remove this _vertexArrayObject = 0;


openscenegraph:
heh, I made the same change and checked it into master and OpenSceneGraph-3.6 a couple of minutes before opening up the browser...  I'll close this PR as I it does the same as my commit (mine just commented the line out rather than deleting it).

If this fix address the crash you are seeing then could you close the Issue and I'll tag the next 3.6.4 rc. Thanks.

mp3butcher:
good for me 
before tagging would aa9173828aeed4c0c536942473bcc1e1d74770f8 and 38cb6b3c491d470aeda279f89a04799bf2eca867 could be added
Thank you

PR #703, opened 2019-01-25
Changes to OcclusionQueryNode

dan-t:
Hi Robert,

we had a few issues in our application with the OcclusionQueryNode which the first two commit fixed. The last one is more cosmetic one.

I'm not that happy with the API change of 'QueryGeometry::getNumPixels', but I didn't wanted
to create another method like 'QueryGeometry::isQueryValid', because it would mean
to take the same locks as in 'getNumPixels' once more, and on the other side user should
be more aware about the possible invalid return value of 'getNumPixels', and last, I didn't want to break the backward compatibility of the API.

Greetings,
Daniel

### DIFF
--- a/include/osg/OcclusionQueryNode
+++ b/include/osg/OcclusionQueryNode
@@ -72,7 +72,9 @@ public:
 
     virtual void drawImplementation( osg::RenderInfo& renderInfo ) const;
 
-    unsigned int getNumPixels( const osg::Camera* cam );
+    // Returns the num visible pixels of the query geometry. The returned
+    // value is only correct if 'valid == true'.
+    unsigned int getNumPixels( const osg::Camera* cam, bool* valid=NULL );
 
     virtual void releaseGLObjects( osg::State* state = 0 ) const;
 
--- a/include/osg/OcclusionQueryNode
+++ b/include/osg/OcclusionQueryNode
@@ -125,6 +125,11 @@ public:
     void setQueryFrameCount( unsigned int frames ) { _queryFrameCount = frames; }
     unsigned int getQueryFrameCount() const { return _queryFrameCount; }
 
+    // Resets the queries. The next frame will issue a new query.
+    // This is useful for big view changes, if it shouldn't be waited for
+    // '_queryFrameCount' till the frame contents change.
+    void resetQueries();
+
     // Indicate whether or not the bounding box used in the occlusion query test
     //   should be rendered. Handy for debugging and development.
     // Should only be called outside of cull/draw. No thread issues.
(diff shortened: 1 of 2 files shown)


openscenegraph:
I have just had a quick look.  I think it would be best to create separate PR's for each set of changes as each set of changes needs to be reviewed/accepted or amended individually.

The change to getNumPixels() doesn't look like the best way to provide the functionality you are after.  It looks to be the getNumPixels() method should remain the same but for a getTestResult(const Camera*) method would be more flexible and functional.

dan-t:
> The change to getNumPixels() doesn't look like the best way to provide the functionality you are after. It looks to be the getNumPixels() method should remain the same but for a getTestResult(const Camera*) method would be more flexible and functional.
> 
I thought about this change, but didn‘t like the exposing of the query internals.

But if you prefer it that way, I‘m gonna prepare a corresponding change.

dan-t:
> The change to getNumPixels() doesn't look like the best way to provide the functionality you are after. It looks to be the getNumPixels() method should remain the same but for a getTestResult(const Camera*) method would be more flexible and functional.
> 
Returning a TestResult doesn‘t seem to be safe, without taking a lock on _mapMutex.

openscenegraph:
Hi Dan,

I have done another review and decided to introduce a QueryGeometry::QueryResult struct that is essential a pair of { bool valid, unsigned int numPixels} and introduce a QueryGeometry::getQueueResult(..) that can be used in place of the original getNumPixels(..). This can be seen in: 02d7d980a7c6774175a5f2c9e0f3fdc72a7e4343

I have merged the ttwo other commits that were part of your PR, and made a fix to a scope bug that the final commit of yours introduced.  0f2e9e6557489ec2aff3c86e5be85a7252e5de0a

I have merged all these commits with the 3.6 banch and master.  Could you test these out to make sure everything is working.  When testing with osgocclusionquery example I noticed the the framerate jumps beteween when the result is in and the subgraph gets culled and when the result isn't ready and the subgraph is traversed, where previously it would have culled the subgraph.  Not culling a subgraph is safer than culling so the new behaviour seems better.  I do wonder if the query should wait though, but haven't thought about the mechanism that would need to be introduced to make this happen.  I just want to get 3,6,4 out so I won't go chasing this. 

Robert.

dan-t:
Hi Robert,

sorry for the '_passed' bug, I don't know how I could miss that one.

I've tested the 3.6 branch and it looks fine.

I'm still having one flickering issue in our application which I've to
further investigate if it's an issue in our application or in OcclusionQueryNode.

One thing that most likely should be fixed in OcclusionQueryNode is,
that the box for the query geometry has to be at least >0.01 in all
three dimensions, otherwise the occlusion query yields wrong results.

At least for the box size I would prepare a patch in the next few days.

Greetings,
Daniel

PR #705, opened 2019-01-29
Allow Texture subclasses to override isSafeToUnrefImageData

calumr:
If a texture is not intended to be used in all the contexts, it should be possible to unref the image data.

As we don't know if the texture is intended to be shared between contexts or not, this seems the easiest way to allow applications to override the behaviour. For example, in our application we have a main view that contains large textures we wish to free the image data for, and a smaller auxiliary view that doesn't contain any textures. The auxiliary view is preventing the texture image data from being unref'ed.

I considered adding a Texture::_forceUnrefImageDataAfterApply or some other extra state, but this way seemed cleaner & more generally useful?

### DIFF
--- a/include/osg/Texture
+++ b/include/osg/Texture
@@ -831,7 +831,7 @@ class OSG_EXPORT Texture : public osg::TextureAttribute
         bool isHardwareMipmapGenerationEnabled(const State& state) const;
 
         /** Returns true if the associated Image should be released and it's safe to do so. */
-        bool isSafeToUnrefImageData(const State& state) const {
+        virtual bool isSafeToUnrefImageData(const State& state) const {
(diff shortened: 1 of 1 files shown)


openscenegraph:
This change forces applications to subclass from the various Texture classes to enable the non standard behaviour you want.  This is doable for textures you create yourself but ones loaded from plugins, including native formats would not be supported.   This inconsistency isn't something I feel is good to introduce.

I think what might be better is look into a compile traversal for the graphics context you are interested then look to explicitly unref the image.  I haven't looked into the implementation details, but this is the route I'd take.  I'm going to close this PR as it's only a partial solution to a niche usage case that all users will pay a small penalty for, while a full solution should be possible using another approach without impacting performance for everyone.

PR #710, opened 2019-02-04
Add search location for textures in FBX plugin

bjornblissing:
The FBX plugin may fail texture lookup even if the texture is located in the same folder as the model file. 
Currently the plugin checks the following locations:

1.  The absolute filename appended to the current directory _(Works well when the absolute filename is a pure filename, but fails when the absolute filename is a complete path)_

2. The absolute filename only. _(Works well assuming the texture can be found at exactly this complete path)._

3. The relative filename appended to the current directory _(requires that the texture is placed at exactly in the same relative location to the model as the file was saved in)._

This change adds a fourth option:

4. The pure filename part of the absolute filename (i.e. without any paths) appended to the current directory. This will look in the same folder as the model is stored in.

### DIFF
--- a/src/osgPlugins/fbx/fbxMaterialToOsgStateSet.cpp
+++ b/src/osgPlugins/fbx/fbxMaterialToOsgStateSet.cpp
@@ -184,14 +184,18 @@ osg::ref_ptr<osg::Texture2D> FbxMaterialToOsgStateSet::fbxTextureToOsgTexture(co
 	{
 		filename = osgDB::concatPaths(_dir, fbx->GetFileName());
 	} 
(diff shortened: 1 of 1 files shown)


bjornblissing:
The first case makes little sense (at least to me). I cannot imagine a situation where concatenating the current model directory and an absolute path to a texture would make any sense. 

My current pull request do not break any existing applications at least. But unless this request gets merged into the current release I would probably re-submit this pull request and eliminate the first case. So the search order would be:

1. Try the absolute path
2. Else try the current directory + the relative path
3. Else try the current directory + the simple filename.

openscenegraph:
Thanks, I've merged with master and the 3.6 branch.

PR #711, opened 2019-02-07
Cleanups and support for macOS 10.14

uhertlein:
cmake cleanups and support for macOS 10.14


openscenegraph:
Thanks @uhertlein, I've merged with master.  What is the status with the OpenSceneGraph-3.6 branch and 10.14?  Do we need to apply this patch?  Will it likely work fine if I do apply it?

PR #713, opened 2019-02-21
Ensure valid occlusion query geometry

dan-t:
Regarding a race condition setting _validQueryGeometry, it's already protected by the _computeBoundMutex, which is already used for updating the query geometry.

The _computeBoundMutex isn't locked when the query geometry is accessed in the rest of the code, so I suspect that accessing _validQueryGeometry should also be fine, otherwise there currently already is a race condition.

About the validity of the query geometry, what I found in our application is, that the
box has to be at least >=0.01 in all three dimensions, otherwise the occlusion query
doesn't yield correct results.

If you like I could also add such a test to OcclusionQueryNode.

### DIFF
--- a/include/osg/OcclusionQueryNode
+++ b/include/osg/OcclusionQueryNode
@@ -198,6 +198,9 @@ protected:
 
     bool _enabled;
 
+    // If the box of the query geometry is valid.
+    mutable bool _validQueryGeometry;
+
     // Tracks the last frame number that we performed a query.
     // User can set how many times  (See setQueryFrameCount).
     typedef std::map< const osg::Camera*, unsigned int > FrameCountMap;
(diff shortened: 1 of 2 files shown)


openscenegraph:
I'll merge the change, though I have to admit as I'm not the original author of this functionality I have to sit and read the code to work out the intent, it's overall design is rather awkward.  If I came to implement the functionality from scratch I would probably go a different route.  Exactly what they might be I can't say.  I'm afraid I have enough other work on my plate so I can't dive into too deeply on this stuff.

dan-t:
Hi Robert,

are my latest changes regarding the OcclusionQueryNode going into the OpenSceneGraph 3.6.4 release?

Greetings,
Daniel

mp3butcher:
In general mutable are percontext variable...
Would'nt it be more reliable to use buffered_value<int>?...

Issue #715, opened 2019-02-25
Geometry shader parameter OpenGL error


AnyOldName3:
On some systems (the user who reported this to me is using Mesa with a Radeon Fury), OpenGL errors are being triggered here: https://github.com/openscenegraph/OpenSceneGraph/blob/OpenSceneGraph-3.6.4-rc3/src/osg/Program.cpp#L714

Investigation has lead me to believe that it's because of the differences between the OpenGL extensions exposing geometry shaders and the core feature of GL 3.2 that does the same thing.

In the extensions (e.g. `GL_EXT_geometry_shader4` and `GL_ARB_geometry_shader4`) parameters such as input and output type are set as program parameters, just as OSG is doing on the linked lines. However, in the core feature, they're set in the shader source itself. On systems which have the core feature, but not the older extensions, the `glProgramParameteri` calls are erroneous, and are causing errors to be reported.

I'm not sure how OSG handles the two different implementations of geometry shaders (if at all), but a short-term fix would be to change the final call on this line https://github.com/openscenegraph/OpenSceneGraph/blob/OpenSceneGraph-3.6/src/osg/GLExtensions.cpp#L458 from `isGLExtensionOrVersionSupported` to `isGLExtensionSupported`. It's definitely not correct to have it as `isGLExtensionOrVersionSupported` given that the extension and core feature are different, so the presence of the extension isn't implied by the GL version.

AnyOldName3:
In addition to that, it looks like the `osg::isGLExtensionSupported(contextID,"GL_OES_geometry_shader")` call is bad, too. `GL_OES_geometry_shader` exposes geometry shaders in the same way as the core OpenGL 3.2 feature, so that extension being present doesn't indicate that the `glProgramParameteri` calls should be happening, either.

mp3butcher:
use Program::setParameter

AnyOldName3:
That won't help with this - the parameters simply don't exist in core OpenGL 3.2 or `GL_OES_geometry_shader` (as they're part of the GS source instead), so they shouldn't be set at all.

AnyOldName3:
Just to clarify in case it wasn't clear - the error is triggered whether or not the application actually uses any geometry shaders. Whenever OSG links a per-context program, it sets these parameters just in case a geometry shader is in use.

emminizer:
[...]

PR #716, opened 2019-02-26
Resolve OpenGL error caused by confusion between geometry shader implementations

AnyOldName3:
There are two types of Geometry Shaders in OpenGL.

One is provided by `GL_EXT_geometry_shader4` and `GL_ARB_geometry_shader4` and requires that certain parameters are set with calls to `glProgramParameteri` before the program is linked. Without these extensions, the calls are erroneous.

The other is provided by OpenGL 3.2 as a core feature and by the GLES extensions `GL_EXT_geometry_shader` and `GL_OES_geometry_shader` and requires these parameters to be set in the GLSL source itself.

The value of `isGeometryShader4Supported` is being used [here](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Program.cpp#L709-L714) to determine if the `glProgramParameteri` calls should occur, so it should only be `true` when one of the extensions requiring them is present.

Resolves https://github.com/openscenegraph/OpenSceneGraph/issues/715

### DIFF
--- a/src/osg/GLExtensions.cpp
+++ b/src/osg/GLExtensions.cpp
@@ -457,7 +457,7 @@ GLExtensions::GLExtensions(unsigned int in_contextID):
     isVertexShaderSupported = validContext && (shadersBuiltIn || osg::isGLExtensionSupported(contextID,"GL_ARB_vertex_shader"));
     isFragmentShaderSupported = validContext && (shadersBuiltIn || osg::isGLExtensionSupported(contextID,"GL_ARB_fragment_shader"));
     isLanguage100Supported = validContext && (shadersBuiltIn || osg::isGLExtensionSupported(contextID,"GL_ARB_shading_language_100"));
-    isGeometryShader4Supported = validContext && (osg::isGLExtensionSupported(contextID,"GL_EXT_geometry_shader4") || osg::isGLExtensionSupported(contextID,"GL_OES_geometry_shader") || osg::isGLExtensionOrVersionSupported(contextID,"GL_ARB_geometry_shader4", 3.2f));
+    isGeometryShader4Supported = validContext && (osg::isGLExtensionSupported(contextID,"GL_EXT_geometry_shader4") || osg::isGLExtensionSupported(contextID,"GL_ARB_geometry_shader4"));
     isGpuShader4Supported = validContext && osg::isGLExtensionOrVersionSupported(contextID,"GL_EXT_gpu_shader4", 3.0f);
     areTessellationShadersSupported = validContext && (osg::isGLExtensionOrVersionSupported(contextID, "GL_ARB_tessellation_shader", 4.0f) || osg::isGLExtensionSupported(contextID,"GL_OES_tessellation_shader"));
     isUniformBufferObjectSupported = validContext && osg::isGLExtensionOrVersionSupported(contextID,"GL_ARB_uniform_buffer_object", 3.1f);


openscenegraph:
I have just done a first pass review and it looks like a case of throw the baby out with the bathwater.  It may well "fix" the "problem" for your application and driver combination but it looks to be me likely it'll over zealously disable a feature that is supported, just not all the features supported.  If it's indeed the later than we need a finergrained solution stripping out the extension support so that different level of features are used on different drivers.

mp3butcher:
I wonder what the purpose of  the block https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Program.cpp#L709-L714 
user could do that by himself with Program::setParameter...Am'nt I right?

mp3butcher:
is reverting https://github.com/openscenegraph/OpenSceneGraph/commit/ebf653369a5eb565f50f5f1b8ef63b98f261c9bc#diff-d06a18d49766bda1560022081d9e57e3
and add a guard on _geometryVerticesOut==0 to prevent setProgamParameter calls could be a proper fix?

AnyOldName3:
> I have just done a first pass review and it looks like a case of throw the baby out with the bathwater. It may well "fix" the "problem" for your application and driver combination but it looks to be me likely it'll over zealously disable a feature that is supported, just not all the features supported. If it's indeed the later than we need a finergrained solution stripping out the extension support so that different level of features are used on different drivers.

Have you read the discussion in the issue report about there being two different types of geometry shaders, the one provided by the extensions with `geometry_shader4`  in the name, and the one provided by OpenGL 3.2 and the extensions *without* `4` in the name? The variable is called `isGeometryShader4Supported`, so it would be sensible to assume it's only supposed to be true when the with-`4` extensions are available.

It might be reasonable to add a second bool called something like `isGeometryShaderButTheVersionFromCoreSupported` (obviously this is a facetious name, but at least it's descriptive) that checks against OpenGL 3.2, `GL_OES_geometry_shader` and `GL_EXT_geometry_shader` so there's a way to query if the other implementation of geometry shaders is available, but given that different GLSL source code is need for each implementation, conflating the two into one bool isn't helpful to anyone.

The only place OSG itself queries `isGeometryShader4Supported` is the location linked in the PR description, so there will be no negative effects to anything that lets OSG handle this. If another application is querying `isGeometryShader4Supported`, it's likely going to already have issues on some GPU driver, as it's doing something dependent on whether either of two incompatible features is available.

mp3butcher:
is reverting https://github.com/openscenegraph/OpenSceneGraph/commit/ebf653369a5eb565f50f5f1b8ef63b98f261c9bc#diff-d06a18d49766bda1560022081d9e57e3
and add a guard on _geometryVerticesOut==0 to prevent glProgamParameter calls could be a proper fix?
https://github.com/openscenegraph/OpenSceneGraph/compare/OpenSceneGraph-3.6...mp3butcher:patch-25

AnyOldName3:
It wouldn't solve the issue of the two different geometry shader implementations being conflated into one check, so while it would prevent the errors on Mesa, it wouldn't stop applications using geometry shaders doing something wrong.

mp3butcher:
@openscenegraph I checked with @JDGBOLT with mesa and core goemetry shader feature doesn't require ProgramParameter to be setted ( settings them in the shader is sufficient) so this pr seams ok

mp3butcher:
OES/EXT/ARB_geometry_shader seams to be related to the presence of the core feature and so on the removal of EXT/ARB_geometry_shader4 so this pr seams ok

AnyOldName3:
There actually isn't an `ARB_geometry_shader` without the `4` - you're expected to check the GLES extensions or rely on the OpenGL version being above 3.2.

I'm going to wait for Robert's feedback before adding an `isCoreGeometryShaderSupported` bool, mostly because I want to be sure that's the best name for it.

openscenegraph:
Thanks for the patience, I've done a review of the code the PR, and the affected GL_GEOMETRY_VERTICES_OUT_EXT etc. settings passed to glProgramPramaters in Program.cpp and looked at GL4 docs and the spec and it's clear that GL_GEOMETRY_VERTICES_OUT_EXT etc. should only ever be passed for geometry_shader4 extension, it doesn't carry over to the geometry shader support in the core GL.  Given this the PR looks appropriate and I don't believe any further changes are required.

For end users it does leave open how they should handle that geometry_shader4 extension can be assumed and the associated setting to be functional.  Perhaps the best thing would be to add a comment into the Program header about the scope of these settings.

PR #718, opened 2019-03-05
Fix osgAnimation transform not updating in some cases

flashk:
The osgAnimation::StackedTransform class will not update the matrix if the state is going from a non-identity to identity value. For example, if you have a stepped animation that is cycling the scale of an object between (1,1,1) and (2,2,2), the object will scale to (2,2,2) and remain at that value, since (1,1,1) is considered identity and will not dirty the matrix.

I've modified the StackedTransform::update class so that it will update the matrix, regardless of the element identity state. It will still check if the element is identity when apply the value to the new matrix.

### DIFF
--- a/src/osgAnimation/StackedTransform.cpp
+++ b/src/osgAnimation/StackedTransform.cpp
@@ -41,9 +41,7 @@ void StackedTransform::update(float t)
             continue;
         // update and check if there are changes
         element->update(t);
-        if (!dirty && !element->isIdentity()){
-            dirty = true;
-        }
+        dirty = true;
     }
 
     if (!dirty)


mp3butcher:
it seams right but can you join an simple example to reproduce the bug (as osgt )

flashk:
I've attached a sample .osgt file with a stepped animation that will toggle the scale of the box between (1,1,1) and (2,2,2). Without the fix, the box will scale to (2,2,2) and remain there, instead of going back to (1,1,1).

[StepScaleAnimation.zip](https://github.com/openscenegraph/OpenSceneGraph/files/2932442/StepScaleAnimation.zip)

mp3butcher:
great :) Thank you

openscenegraph:
Thanks for the testing model, I tried it and confirmed the bug before and that it's fixed with applying the patch.  The explanation of the bug and fix also makes sense.  Cheers.

PR #720, opened 2019-03-05
finish removal of ShadowVolume feature began in b920d482c74d2bc75636b…

mp3butcher:
…7c805c624c2c2d983dd
https://github.com/openscenegraph/OpenSceneGraph/commit/b920d482c74d2bc75636b7c805c624c2c2d983dd

### DIFF
--- a/include/osg/CullingSet
+++ b/include/osg/CullingSet
@@ -14,6 +14,8 @@
 #ifndef OSG_CullingSet
 #define OSG_CullingSet 1
 
+//#define COMPILE_WITH_SHADOW_OCCLUSION_CULLING
+
 #include <osg/Polytope>
 #include <osg/ShadowVolumeOccluder>
 #include <osg/Viewport>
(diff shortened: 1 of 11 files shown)


include/osg/CullingSet
```diff
@@ -14,6 +14,8 @@
 #ifndef OSG_CullingSet
 #define OSG_CullingSet 1
 
+//#define COMPILE_WITH_SHADOW_OCCLUSION_CULLING
```
Capostrophic: This appears to have been undef'd by mistake.

openscenegraph:
You look to be conflating ShadowVolume with the convex planar occluder culling, these are two completely different features.

Also having lots of #ifdef's across the codebase is just really bad programming and something one should avoid as much as possible,  The more optional compile paths you have in a codebase the less testing the different code paths undergo the higher the chance bugs will be missed in testing and the lower the software quality will be over time.

For these two reasons this PR is not appropriate for merging.

PR #721, opened 2019-03-08
fix texture output bug of 3ds plugin

0871087123:
I found i can't output a texture correct 3ds file when i have multiple node referenced same texture image. 
And found in this LOC, it used a unique file name instead of correct file name.
Is it should have unique file for every material in 3ds file ? 
I guess node name should use unique name but not sure about texture file name. So i deleted it and it works.
Now i have texture coorect 3ds file, verified with assimp_viewer.

### DIFF
--- a/src/osgPlugins/3ds/WriterNodeVisitor.cpp
+++ b/src/osgPlugins/3ds/WriterNodeVisitor.cpp
@@ -528,7 +528,6 @@ void WriterNodeVisitor::writeMaterials()
                         path = osgDB::getPathRelative(_srcDirectory, mat.image->getFileName());
                     }
                     path = convertExt(path, _extendedFilePaths);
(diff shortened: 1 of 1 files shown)


openscenegraph:
I'm not the original author of the code, but looking at it the author clearly had the intent of making sure that the filenames were unique, this makes sense if we are writing out a new complete model and associated image files.  

When making the file unique the other references with the output .3ds needs to reference the new unique file name, so if this isn't working then that's where the bug is.  Just removing this line might fix things for your particular usage case, but it's really just hiding a bug that other could hit if the lack of unique filename could result in an inappropriate image being loaded.

The right thing to do is provide a bug report via Issues with an explanation of how to reproduce the issue, and if you have any insights then provide these.  Then others can look into the actual bug rather than just hiding it for one particular usage case.

PR #723, opened 2019-03-13
Add OSG_USE_FLOAT_QUAT build option to use single precision quaternions

Capostrophic:
Some OSG-based projects benefit (or will benefit) from the loss of precision in quaternions when there isn't a focus on complex calculations and simulations. So this makes use of Quat::value_type so that single precision quaternions are available through a build-time option, which is off by default, very alike to the one available for matrixes.

I'm not very familiar with OSG codebase and plugins, but I think I shouldn't have missed anything.


Capostrophic:
Oh, cool. I somewhat wonder if this and/or my other miniscule PR are non-intrusive enough that they could be merged into 3.6 branch as well, but if they're not, it's OK.

psi29a:
If necessary, I can get this downstreamed into Debian/Ubuntu.

PR #729, opened 2019-03-18
Use GNUInstallDirs for library installation path

jslee02:
This PR is based on the [comment](https://github.com/openscenegraph/OpenSceneGraph/pull/660#issuecomment-441502412) in #660.

OSG currently installs binaries in `lib<arch>` (e.g., `lib64`), which is not searched by CMake < 3.12. This PR changes the binary installation path using [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) that is the built-in standard way.


openscenegraph:
I have done a first pass review and am concerned that users will find a different behaviour than they are used to with previous versions of the OSG.  What differences are you aware of and on what platform combinations?

I looked into the GUIInstallDirs history and it looks like it requires CMake 2.8.5, while the OSG CMake build system currently only requires 2.8.0 so either ther use of GUIInstallDir will need to be guarded and a fallback provided or for us to set 2.8.5 as a minimum.  Again I have concern that just upping the required version to 2.8.5 will break the build for some users on older systems.

I flag these issues because we could easily break the build for users so we need to establish first what these might be.

jslee02:
> I have done a first pass review and am concerned that users will find a different behaviour than they are used to with previous versions of the OSG.

This change shouldn't affect the users as long as they use the `find_library()` to find the libraries as `FindOpenSceneGraph.cmake` and `FindOSG.cmake` do.

> What differences are you aware of and on what platform combinations?

I'm on Ubuntu 16.04. If OSG is built and installed from source, the binaries are installed in `/usr/lib64` that is not searched by `find_library()` unless:
* CMake >= 3.12
* set `LIB_POSTFIX` explicitly as `/<arch>` when run cmake for OSG (e.g., `cmake .. -DLIB_POSTFIX="/${DEB_HOST_MULTIARCH}" )
* put the line `set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS TRUE)` some time before your call to `find_package(OpenSceneGraph ...)` in the downstream

> I looked into the GUIInstallDirs history and it looks like it requires CMake 2.8.5, while the OSG CMake build system currently only requires 2.8.0 so either ther use of GUIInstallDir will need to be guarded and a fallback provided or for us to set 2.8.5 as a minimum. Again I have concern that just upping the required version to 2.8.5 will break the build for some users on older systems.

In that case, I agree with that it should gaurd the use of `GUIInstallDir` for CMake < 2.8.5 and fallback to the original behavior something like:
```cmake
IF(OSG_VERSION VERSION_LESS 2.8.5)
  IF(UNIX AND NOT WIN32 AND NOT APPLE)
    IF(CMAKE_SIZEOF_VOID_P MATCHES "8")
        SET(LIB_POSTFIX "64" CACHE STRING "suffix for 32/64 dir placement")
        MARK_AS_ADVANCED(LIB_POSTFIX)
    ENDIF()
  ENDIF()
  IF(NOT DEFINED LIB_POSTFIX)
    SET(LIB_POSTFIX "")
  ENDIF()
  set(OSG_INSTALL_LIBDIR lib${LIB_POSTFIX})
ELSE()
  include(GNUInstallDirs)
  set(OSG_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR})
ENDIF()
```
then use `OSG_INSTALL_LIBDIR` instead of `CMAKE_INSTALL_LIBDIR`.

So, with the guard, this change shouldn't break anything but allows the users to find OSG using `find_package(OpenSceneGraph ...)` without any additional settings.

openscenegraph:
@jslee02  Could you add the Cmake version guard, this will help tighten up the behaviour.  Thanks.

jslee02:
Updated the logic: It only uses `GNUInstallDirs` for the library installation directory when `LIB_POSTFIX` is not specified **and** CMake >= 2.8.5. Otherwise, it should be the same as before.

PR #731, opened 2019-03-19
Check for empty DrawElements to avoid undefined behaviour.

AnyOldName3:
In `DrawElementsUByte/Short/Int`, if the vector is empty when `draw` is called, `glDrawElements` is still called with `&front()` as one of the arguments. Calling `front()` on an empty vector is undefined behaviour, which is bad (and also throws an exception on Debug builds on Windows).

This avoids the issue by returning from `draw` early when the vector is empty.

### DIFF
--- a/src/osg/PrimitiveSet.cpp
+++ b/src/osg/PrimitiveSet.cpp
@@ -187,6 +187,9 @@ DrawElementsUByte::~DrawElementsUByte()
 
 void DrawElementsUByte::draw(State& state, bool useVertexBufferObjects) const
 {
+    if (empty())
+        return;
+    
     GLenum mode = _mode;
     #if defined(OSG_GLES1_AVAILABLE) || defined(OSG_GLES2_AVAILABLE) || defined(OSG_GLES3_AVAILABLE)
         if (mode==GL_POLYGON) mode = GL_TRIANGLE_FAN;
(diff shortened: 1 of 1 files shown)


openscenegraph:
Add a bool check to a an operation done very often to just catch buggy data is penalty that all users applications would have to pay just to avoid buggy data not crashing one users application.

The correct fix for this type of bug is to catch the buggy data well before it gets added to the scene graph, or to have a visitor that can go through the scene graph and remove these buggy elements.

AnyOldName3:
If empty geometries and primitive sets are classed as buggy and cause undefined behaviour, it's probably a good idea to document that somewhere. It looks like http://forum.openscenegraph.org/viewtopic.php?t=4505 is the only mention of the issue that existed until I stumbled upon the same problem, and it required a fairly specific Google query to find, so I doubt anyone's going to stumble upon it especially easily.

openscenegraph:
It's an obscure issue that almost no users have reported issues with - the post you mention is from a decade ago. Do we need to document every edge case issue of the OSG not catching erroneous data that has occurred in the lifetime of the project?

mp3butcher:
perhaps adding a WARN<<"Beware: Upload of an empty BufferData"<<std::endl;  in GLBufferObject would be a good way to deal with this issue

mp3butcher:
perhaps adding a WARN<<"Beware: Upload of an empty BufferData"<<std::endl;  in GLBufferObject would be a good way to deal with this issue

PR #733, opened 2019-03-23
Collada (DAE) support recognized for VisualStudio 2017

eknabe:
The 3rd party package contain Collada library, but it was not found for Visual Studio 2017 (vc141). Minor modification in CMakeModules/FindCOLLADA.cmake solves the problem.


CMakeModules/FindCOLLADA.cmake
```diff
@@ -29,9 +29,12 @@ IF(APPLE)
 ELSEIF(MINGW)
     SET(COLLADA_BUILDNAME "mingw")
     SET(COLLADA_BOOST_BUILDNAME ${COLLADA_BUILDNAME})
-ELSEIF((MSVC_VERSION GREATER 1900) OR (MSVC_VERSION EQUAL 1900))
+ELSEIF(MSVC_VERSION GREATER_EQUAL 1910)
```
jslee02: You may want to keep the old style for version comparison as:
```suggestion
ELSEIF((MSVC_VERSION GREATER 1910) OR (MSVC_VERSION EQUAL 1910))
```
or
```suggestion
ELSEIF(NOT (MSVC_VERSION LESS 1910))
```
because [`GREATER_EQUAL` was introduced in CMake 3.7](https://cmake.org/cmake/help/v3.7/release/3.7.html#commands) while [OpenSceneGraph supports CMake 2.8.0](https://github.com/openscenegraph/OpenSceneGraph/blob/master/CMakeLists.txt#L16).

openscenegraph: We need to retain support for the older CMake version so need to aboid useding GREATER_EQUAL.  Could you revise the submission, the easiest fix would be to simply bump the version number from 1900 to 1910 as you did in you "old style" suggestion.  

I'm not a Windows users. Why is there a need to bump the version from 1900 to 1910.  What is the problem that occurs without it?

eknabe: Ok. I'm not a Visual Studio expert, but I'll try to explain. Using Visual Studio 2017 (vs 2015 / vc141) the DAE/Collada plugin was not "found". I think since the current statement is:
ELSEIF((MSVC_VERSION GREATER 1900) OR (MSVC_VERSION EQUAL 1900))
    SET(COLLADA_BUILDNAME "vc14")
    SET(COLLADA_BOOST_BUILDNAME "vc140")

So, it falls back to vs14/vc140 which corresponds to Visual Studio 2015. I just tried to bump to vs15/vc141/VisualStudio2017 by changing into (also applying jslee02 excellent mod):

ELSEIF((MSVC_VERSION GREATER 1910) OR (MSVC_VERSION EQUAL 1910))
    SET(COLLADA_BUILDNAME "vc14")
    SET(COLLADA_BOOST_BUILDNAME "vc141")
ELSEIF(MSVC_VERSION EQUAL 1900)
    SET(COLLADA_BUILDNAME "vc14")
    SET(COLLADA_BOOST_BUILDNAME "vc140")	

It is confusing with all versions numbers. Here's a table: https://cmake.org/cmake/help/v3.12/variable/MSVC_VERSION.html

Anyway, the bottomline is: Without this fix VisualStudio2017 users will not get DAE support, even though the lib exists in the 3rdParty package. With the fix, it will.

eknabe: BTW, @jslee02: I had no clue about this. Thanks for reviewing and suggest a more robust solution!

eknabe: I you agree with the modified solution. Please help me, what is the next step I should take? Can I just accept by "commit suggestion"?

jslee02: @eknabe Yeah, you could either apply the suggested change by clicking the "commit suggestion" button or modify the code locally and push it (as the normal process). The code suggestion feature of GitHub is just to provide an easy way to collaborate within pull requests.

PR #735, opened 2019-03-26
fix fbx redundant node issue

Tlsirius:
Saving fbx function creates redundant fbx nodes, including one empty parent node for the whole scene and one empty node for each Geometry. Every time you load the model and save to file, these redundant nodes would accumulate, so after repeating load&save for some times the hierarchy become a mess. This commit fixes the problem and ensures fbx file remains the same after each load&save operation.


AnyOldName3:
You've got some tabs.

Tlsirius:
@AnyOldName3  Thanks. I've fixed it.

PR #736, opened 2019-03-26
Android ES 1 Maintenance

ccochran:
During my quest to build OSG for modern Android ABIs and Open GL ES 1 (which is probably not so popular) I ran into a few problems.  

First, the PNG plug-in would not build for Android ES 1, this was because it erroneously required exceptions to be present (checking the png plugin source, it guards against lack of cpp exceptions) so I removed its requirement from CMake. This solved the issue.

Finally the Android Macro cmake file needed some updating.  After building my 3rd party dependencies (png, jpeg, freetype), I noticed they were not being found bc this cmake file assumes we have an Android.mk (CMake is very popular now) file present to locate the header files, so I changed this to find the main header file of each dependency (this is actually how the GDAL portion is written as well).  I also modified it to only install the current ABI library instead of all versions of the ABI. This update assumes 3rd party dep file structure matches this:

```
<third party dependency> /
   include /
   lib /
      <abi 1> /
      <abi 2> /
       ...
```

instead of:

```
build /
   <third party dependency> /  
<third party dependency> /
   Android.mk
```


src/osgText/Text.cpp
```diff
@@ -49,6 +49,7 @@ Text::Text():
 {
     _supportsVertexBufferObjects = true;
 
+#ifndef OSG_GLES1_AVAILABLE
```
vsg-dev: I don't understand the need for this particular change. DisplaySettings::_textShaderTechnique is an empty string by default, so NO_TEXT_SHADER should be selected by default.  

I have strong preference for not putting platform specific hacks all over the OSG, we should only add them if we really really need them as every optional compile path is another build combination that needs to be maintained and could potentially break unseen for a long time.

ccochran: I agree adding the hack isn't a good solution, I'll take a closer look at why shaders are being added.  I haven't changed my client Text code since osg 3.2.1 so something with the Android build may have changed.

ccochran: Okay I see the problem: `NO_TEXT_SHADER` is never set on `_shaderTechnique` if the `_textShaderTechnique` is empty so it remains the default value, `GREYSCALE`. I'll explicitly set `NO_TEXT_SHADER` if the string is empty:

```
Text::Text():
    _shaderTechnique(GREYSCALE),
    _enableDepthWrites(true),
    _backdropType(NONE),
    _backdropHorizontalOffset(0.07f),
    _backdropVerticalOffset(0.07f),
    _backdropColor(0.0f, 0.0f, 0.0f, 1.0f),
    _colorGradientMode(SOLID),
    _colorGradientTopLeft(1.0f, 0.0f, 0.0f, 1.0f),
    _colorGradientBottomLeft(0.0f, 1.0f, 0.0f, 1.0f),
    _colorGradientBottomRight(0.0f, 0.0f, 1.0f, 1.0f),
    _colorGradientTopRight(1.0f, 1.0f, 1.0f, 1.0f)
{
    _supportsVertexBufferObjects = true;

    const std::string& str = osg::DisplaySettings::instance()->getTextShaderTechnique();
    if (!str.empty())
    {
        if (str=="ALL_FEATURES" || str=="ALL") _shaderTechnique = ALL_FEATURES;
        else if (str=="GREYSCALE") _shaderTechnique = GREYSCALE;
        else if (str=="SIGNED_DISTANCE_FIELD" || str=="SDF") _shaderTechnique = SIGNED_DISTANCE_FIELD;
        else if (str=="NO_TEXT_SHADER" || str=="NONE") _shaderTechnique = NO_TEXT_SHADER;
    }
    else
    {
        _shaderTechnique = NO_TEXT_SHADER;
    }

    assignStateSet();
}
```

src/osgText/Text.cpp
```diff
@@ -57,6 +57,10 @@ Text::Text():
         else if (str=="SIGNED_DISTANCE_FIELD" || str=="SDF") _shaderTechnique = SIGNED_DISTANCE_FIELD;
         else if (str=="NO_TEXT_SHADER" || str=="NONE") _shaderTechnique = NO_TEXT_SHADER;
     }
+    else
```
vsg-dev: This fix now disables use of GREYSCALE for all places where osg::DisplaySettings::instance()->getTextShaderTechnique() is not set, which is the default.  The default is GREYSCALE as this is best default implementation.   

If you want to locally disable this then one should use the osg::DisplaySettings::instance()->getTextShaderTechnique();  or if the OS doesn't support a particular options such as shaders not being available then this is determination that should be done at runtime once the graphics context is tested.  I haven't looked into this to know how practical it would be.  Right now my preference would be if you want to select something that isn't default then _shaderTechnique default should be set via the osg::DisplaySettings::instance()->getTextShaderTechnique(); value or have the Text objects manually set the technique,

ccochran: Okay, it sounded like in your response that you expected the default value to be NO_SHADER.  

I'll try the DisplaySettings approach, the ctor itself crashes so setting the technique on the Text obj after it is created isn't possible.

ccochran: I call `setTextShaderTechnique` as you suggested from my client code which works for me. I removed commits from this PR associated with text shaders / ES1.

vsg-dev:
On Wed, 27 Mar 2019 at 18:40, Colin <notifications@github.com> wrote:

> *@ccochran* commented on this pull request.
> ------------------------------
>
> In src/osgText/Text.cpp
> <https://github.com/openscenegraph/OpenSceneGraph/pull/736#discussion_r269714388>
> :
>
> > @@ -57,6 +57,10 @@ Text::Text():
>          else if (str=="SIGNED_DISTANCE_FIELD" || str=="SDF") _shaderTechnique = SIGNED_DISTANCE_FIELD;
>          else if (str=="NO_TEXT_SHADER" || str=="NONE") _shaderTechnique = NO_TEXT_SHADER;
>      }
> +    else
>
> I call setTextShaderTechnique as you suggested from my client code which
> works for me. I removed commits from this PR associated with text shaders /
> ES1
>
For this PR I think this is the best route forward.

There may still need to be some amendment to osgText to stop the crash when
this TextShaderTechnique isn't set, I don't have the time to look into this
right away so I'll have come back to this the next opportunity that I have
for some longer support.

PR #744, opened 2019-04-11
macOS support for glVertexAttribDivisor with legacy GL

zakalawe:
In compatibility mode (OpenGL 2.1), glVertexAttribDivisor is available
via glVertexAttribDivisorARB on macOS.


zakalawe:
I will also submit patches against the stable versions if that's okay? Or is there a different procedure there?

PR #747, opened 2019-04-18
Intersections with Text3D fixed + drawImplementation optimized

valid-ptr:


openscenegraph:
This breaks the rendering of 3D when the _characterSizeMode!=OBJECT_COORDS || _autoRotateToScreen, as computeMatrix() needs to be called in the drawImplementation() in this case - this why there is the complicated games in the drawImplementation().

The intersections sounds like a bug.  Is there a way to reproduce it easily?

valid-ptr:
Hello, Robert!
You are right about "_characterSizeMode!=OBJECT_COORDS || _autoRotateToScreen" while rendering.

It's easy to reproduce an intersection bug:
Just put any Text3D in MatrixTransform and translate it somewhere, 
use osgpick example for picking etc.

The BUG comes from
`void Text3D::accept(osg::PrimitiveFunctor& pf) const`
where _coords->front() are not multiplied by the **_matrix**.

openscenegraph:
I am about to start looking into the bug this PR attempts to address, first step is to create an example that reproduces the bug then move on to fixing the bug, which likely will require some mods to intersection API.  If you have an example that reproduces this issue could you provide this as it'll save me some time.

valid-ptr:
Hello, Robert and welcome back to OSG from [VSG](https://github.com/vsg-dev/VulkanSceneGraph) :)

I don't have an example, sorry!

openscenegraph:
I have added a picking test into osgtext3D example and this reproduces the intersection problems:

    https://github.com/openscenegraph/OpenSceneGraph/commit/59f417f1e01d9667ffd6f7c9481b599023856d1c

To resolve these problems I've merged the matrix accounting code from Text.cpp:

   https://github.com/openscenegraph/OpenSceneGraph/commit/02d91a3eea16dc010051f428bce994d9cc7751d8

PR #752, opened 2019-04-30
fix osgconv for readerWriters that don't implement readObject

LaurensVoerman:
Hi Robert,

I was surprised by
  C:\Users\laurens>osgconv cow.osg.0,180,0.rot
  Error reading file cow.osg.0,180,0.rot: not implemented

Tracing back the problem I found it was caused by your commit on 20/11/2018
* Added support for reading and writing images

this caused all plugins that don't implement readObject to fail with osgconv.
I implemented this fix by trying readRefNodeFile if readRefObjectFile fails 
(changed readObjectFile to readRefObjectFile to keep things working if OSG_PROVIDE_READFILE is not defined)
Flipped the logic to try readRefNodeFile first to avoid 
  Error reading file cow.osg.0,180,0.rot: not implemented
for nearly all 3d models, but now the waring is given for all images (both versions work fine, despite the warning).

I guess this could be fixed by changing the default impementation of readObject (ReaderWriter:239) to call readNode, 
or implementing readObject in all readerWriters. I'm not totaly sure this will have no side effects, so for now I propose this fix.

Regards, Laurens.

### DIFF
--- a/applications/osgconv/osgconv.cpp
+++ b/applications/osgconv/osgconv.cpp
@@ -792,12 +792,18 @@ int main( int argc, char **argv )
         itr != fileNames.end();
         ++itr)
     {
-        osg::ref_ptr<osg::Object> object = osgDB::readObjectFile(*itr);
-        if (object.valid())
-        {
-            if (object->asNode()) nodes.push_back(object->asNode());
-            else if (object->asImage()) images.push_back(object->asImage());
-            else objects.push_back(object);
+        // a few plugins don't implement readObjectFile
+        osg::ref_ptr<osg::Node> node = osgDB::readRefNodeFile(*itr);
+        if (node.valid()) {
+            nodes.push_back(node);
+        } else {
+            osg::ref_ptr<osg::Object> object = osgDB::readRefObjectFile(*itr);
+            if (object.valid())
+            {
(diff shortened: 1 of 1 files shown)


openscenegraph:
I've done a review of the change and I understand the motivation but it only handles the case where a ReaderWriter only implements the readNodeFile() and not the readObjectFile(), it doesn't handle the case where a plugin implements a readImageFile() and not readObjectFile() etc. 

I believe this approach is really just kicking the can down the road a bit, it isn't actually solving the problem.  I think the proper solution is to go through the individual ReaderWriter's and implement the readObjectFile() to map to the appropriate readNode/Image/HeightField/Shader etc.   While tedious I don't think this would take too much work, as as it doesn't change the public API should be fine for merging with the 3.6 branch and master.

LaurensVoerman:
I agree, I'll try to create a PR for that.
Laurens.

LaurensVoerman:
I've created the PR adding the readObject functions:
 for 3.6 branch: https://github.com/openscenegraph/OpenSceneGraph/pull/760
 for master branch: https://github.com/openscenegraph/OpenSceneGraph/pull/761
- so I'll close this PR.
Regards, Laurens.

PR #756, opened 2019-05-10
Update CMakeLists.txt

blaze2012:


openscenegraph:
This really doesn't look like an appropriate fix as it'll force dependencies on other users that have no need for them.  Others have used this plugin for decades without problem so there is something specific to you build of libtiff that is causing problems on the OSG side.  How best to resolve this is best done on the osg-users ML/forun.

PR #757, opened 2019-05-13
fix compile error if OSGUTIL_RENDERBACKEND_USE_REF_PTR not defined in include/osgUtil/RenderLeaf

LaurensVoerman:
Ran into a compile error after removing the #define OSGUTIL_RENDERBACKEND_USE_REF_PTR (to see if that would render faster on windows; it does not).

Fix applies to master as well.
Laurens.


openscenegraph:
Thanks, merged with 3.6 and master.

PR #758, opened 2019-05-13
fix indentation for osgt files for windows dynamic library build.

LaurensVoerman:
windows std::endl is a template with different addresses in different dll's
causing the indentation to fail for osgt files.
Fix applies to master as well.
Laurens.


openscenegraph:

PR #759, opened 2019-05-13
Various TGA improvements

AnyOldName3:
This PR makes various improvements to the TGA loader, including adding support for type 3, 9 and 11 TGA images. While it's fairly large, each individual commit is intended to be easily reviewable on its own.

It's been tested with this suite of TGA files: https://github.com/ftrvxmtrx/tga/tree/master/testdata. This is comprised of Truevision's official TGA 2.0 Utilities Package plus some extras, and seems to be the same test suite that lots of other TGA loaders are using.

### DIFF
--- a/src/osgPlugins/tga/ReaderWriterTGA.cpp
+++ b/src/osgPlugins/tga/ReaderWriterTGA.cpp
@@ -76,6 +76,15 @@ simage_tga_error(char * buffer, int buflen)
     return tgaerror;
 }
 
+enum AttributeType
+{
+    NO_ALPHA = 0,
+    ALPHA_UNDEFINED_IGNORE = 1,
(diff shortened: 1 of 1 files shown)


psi29a:
This is brilliant, thank you.

AnyOldName3:
I've had a thought about loading R5G5B5 images, too - this reader just shifts the bits left so `ABCDE` becomes `ABCDE000`. I've seen other readers either repeat the bit pattern (`ABCDEACB`) or multiply by the new maximum then divide by the old maximum (`(ABCDE * 255) / 31`) to try and distribute colours more evenly across the available gamut. I'd like Robert's opinion on such a change before I commit it.

psi29a:
Sweet! :)

PR #762, opened 2019-05-15
fix typo in ReaderWriterFBX: writObject -> writeObject

LaurensVoerman:
Hi Robert,
I noticed a type in the function name for fbx writeObject, and while I'm not sure if there is any code path that might call this function I suppose it's better to fix this anyway.
Laurens.


openscenegraph:
Thanks, fix merged with master and 3.6 branch

Issue #439, opened 2018-01-11
VBO vs VAO again... On Linux


valid-ptr:
Hello!

I've written an example for Linux, which demonstrates an issue like in #433 (but on Linux).
Please comment/uncomment  TEST{1,2}_ENABLE defines and run with
OSG_VERTEX_BUFFER_HINT=VBO/VAO to reproduce. There is something wrong with attibute locations using VBO mode in the current OSG master after 5afd32b commit.

NOTE:
I've used OpenSceneGraph 3.5.10 "default build". So OSG_GL_FIXED_FUNCTION_AVAILABLE should be true, so
```
_useModelViewAndProjectionUniforms = false;
_useVertexAttributeAliasing = false;
```
in the State.cpp by default.

**EXAMPLE**:

```
// Try to run with VAO and without
// Good result with VAO: RGB-colored (NORMALS) cessna
// Bad results without VAO : RED cessna

#include <osg/Program>
#include <osgDB/ReadFile>
#include <osgViewer/Viewer>


// NOTE: try to comment/uncomment and see the difference
#define TEST1_ENABLE
//#define TEST2_ENABLE	/* Will help for VBO */


using namespace osg;

// OpenGL 2.1 (compat profile)
static const char* vert21 =
    "#version 110                                                \n"
    "                                                            \n"
    "attribute vec4 osg_Vertex;                                  \n"
    "attribute vec3 osg_Normal;                                  \n"
    "                                                            \n"
    "uniform mat4 osg_ModelViewProjectionMatrix;                 \n"
    "                                                            \n"
    "varying vec3 Vertex_Normal;                                 \n"
    "                                                            \n"
    "                                                            \n"
    "void main()                                                 \n"
    "{                                                           \n"
    "    Vertex_Normal = normalize(osg_Normal);                  \n"
    "    gl_Position = osg_ModelViewProjectionMatrix*osg_Vertex; \n"
    "}                                                           \n"
    "                                                            \n";

static const char* frag21 =
    "#version 110                                                \n"
    "                                                            \n"
    "varying vec3 Vertex_Normal;                                 \n"
    "                                                            \n"
    "void main()                                                 \n"
    "{                                                           \n"
    "    gl_FragColor = vec4(0.5*Vertex_Normal.rgb, 1.0);        \n"
    "}                                                           \n"
    "                                                            \n";

struct SomeRealizeOperation : public osg::GraphicsOperation
{
	SomeRealizeOperation()
[...]

PR #763, opened 2019-05-15
VertexArrayState's lazy disabling of vertex attributes mechanism major fix

valid-ptr:
_vertexAttribArrays must be 'shared'

NOTE: Issue #439 will be really fixed ;)


valid-ptr:
NOTE: #439 will be really fixed ;)

PR #765, opened 2019-05-22
fix typo in header guard.

LaurensVoerman:
This will not change anything, but let's fix it anyway.
Laurens.


openscenegraph:
Thanks merged with master and 3.6 branch

PR #766, opened 2019-05-22
replaced "if (apm || !apm->valid())"

LaurensVoerman:
Am I missing something? Replaced this line with a check to see if any points have been added to the path.
Laurens.


openscenegraph:
Thanks merged with master and 3.6

PR #767, opened 2019-05-27
Removed line that forces to store all images in the object cache.

j-zarzoso:
Setting the objectCacheHint to always store the images in the cache can lead to problems with excessive memory usage when dealing with large models with multiple textures. By removing this line, it's up to the caller of the method to configure the options if he wants to store the images in the object cache.

### DIFF
--- a/src/osgPlugins/fbx/ReaderWriterFBX.cpp
+++ b/src/osgPlugins/fbx/ReaderWriterFBX.cpp
@@ -308,7 +308,6 @@ ReaderWriterFBX::readNode(const std::string& filenameInit,
                 localOptions = options->cloneOptions();
             else
                 localOptions = new osgDB::Options();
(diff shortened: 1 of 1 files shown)


robertosfield:
The use of thee object cache for images should reduce memory usage as it will ensure that repeated requests to load the same image will share the same image.  I feel the original default of enabling the cache is the most appropriate default.

However, I see that the code applies the default of using the object cache even when a osgDB::Options object is passed in to the read call, overriding what is passed in is not appropriate so I've restructured the code a little so that it will allow you to pass in your own osgDB::Option with the settings you want without it locally overriding it.

I will check this in the 3.6 branch and master.

https://github.com/openscenegraph/OpenSceneGraph/commit/8306ce5ca699ef4bd5dd8f5fa29c1d5d049cd5d9

openscenegraph:
I'm going to close this PR as the above commit will provide the ability to override the default behavior with affecting existing users.

PR #768, opened 2019-05-29
Backport object cache fixes

rhabacker:
We are using this fixes in http://www.openscenegraph.org/index.php/gallery/use-cases/189-fm-profil


robertosfield:
This PR changes the ABI for 3.4 so would require a bump in SO number for the 3.4 branch.

I believe all these changes are already part of 3.6 and master, are they all OK?

PR #769, opened 2019-05-29
ParticleSystem serializer: added missing osg::Node to the derivation chain

rickyviking:
missing osg::Node prevents correct serialization, for instance if a stateset is attached to the ParticleSystem


robertosfield:
Unfortunately this change will change the file format so would break all existing .osgb/t/x files that contain particle systems.  

When making changes like this we have to version them.  Have a look at the examples of UPDATE_TO_VERSION_SCOPED usage in the src/osgWrappers/serializers/**.cpp to see how this is done.  I don't recall off the top of my head, but adding an extra entry into the derivation chain requires a special variation of the adding.  I think there is another example of this in action elsewhere in the serializers but can't recall it.  Trawlling in the source will probably required to work this out.

openscenegraph:
I have just checked in the follow to address this issue in a safe way:

https://github.com/openscenegraph/OpenSceneGraph/commit/87597a7fffb2594656ae3f6792606dc482bd2e54

rickyviking:
thanks

PR #770, opened 2019-05-29
obj plugin fixes

rhabacker:
This pull request contains fixes for the obj plugin used by http://www.openscenegraph.org/index.php/gallery/use-cases/189-fm-profil

Request is based on #768


robertosfield:
As this PR includes the ObjectCache changes that change the ABI, we need to resolve the SO version bump before considering merging either PR.

openscenegraph:
i have merged the changes and bumped the SO version number on the 3.4 branch.

PR #775, opened 2019-06-23
Minor fixes to readers

Zaretto:
AC3D reader: display filename and object when primitive not accepted.

Aids locating the cause of the problem. (e.g. not enough vertices).

.dds.gz retain original filename.

The original filename is lost during the filename.gz.dds to filename.gz load process; this change stores the filename in the options name and sets the image name after load.

### DIFF
--- a/src/osgPlugins/ac/ac3d.cpp
+++ b/src/osgPlugins/ac/ac3d.cpp
@@ -43,7 +43,7 @@
 namespace ac3d {
 
 osg::Node*
-readFile(std::istream& stream, const osgDB::ReaderWriter::Options* options);
+readFile(std::istream& stream, const osgDB::ReaderWriter::Options* options, std::string filename);
 
 static std::string ac3dSrcFilename;
 }
@@ -72,6 +72,7 @@ class geodeVisitor : public osg::NodeVisitor { // collects geodes from scene sub
 class ReaderWriterAC : public osgDB::ReaderWriter
 {
     public:
+		std::string ac3dSrcFilename;
 
         ReaderWriterAC()
         {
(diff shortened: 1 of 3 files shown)


Zaretto:
If this is accepted I also plan to backport to the OSG 3.4.1 branch that we (still) currently use in FlightGear.

src/osgPlugins/ac/ac3d.cpp
```diff
@@ -45,6 +45,7 @@ namespace ac3d {
 osg::Node*
 readFile(std::istream& stream, const osgDB::ReaderWriter::Options* options);
 
+static std::string ac3dSrcFilename;
```
robertosfield: Use of static's introduce threading issues, the only way to avoid it would be to use a mutex around the reading and writing of the static variable.  Given this variable looks only there for some debug messages I believe i't really inappropriate change. 

My recommendation would be to ditch all the changes to the ac plugin.  The other change look acceptable.

robertosfield:
I've just done a first pass review and won't be merging it as is due the threading bug introduced in the ac plugin.  My recommendation is to just remove the changes to the ac plugin, once those removed I'll do a second review.

Zaretto:
Thanks for the review; I'll look again at the .ac plugin to see if I can get it to report the problems in a threadsafe manner.

Zaretto:
I've now removed the static - and further improved the error reporting.

openscenegraph:
Oh my, the latest changes are awful.  We are talking extensive changes just to add some debug info that will be useful for a small subset of users.  The problematic static is also still there.

I'm going to close this PR as changes take the code is a bad direction for cleanness and maintainability.

Zaretto:
Thank you for taking the time to review this; I understand what you're saying - it seems unnecessary to overburden the reading code - but at the same time even though it is a small subset of users that will benefit - these errors can be very difficult/tedious to track down. 

I'll have a serious think about how this can be resolved without making awful code.

Issue #779, opened 2019-07-02
Deprecated use of wxWidgets_USE_LIBS is broken with CMake 3.14.0


StefanBruens:
Since https://github.com/Kitware/CMake/commit/0813581859bd61a9d30c97aa893030bea85f134e

FIND_PACKAGE(wxWidgets) allows to specify *OPTIONAL" components. Unfortunately,
everything from wxWidgets_USE_LIBS ends up in the OPTIONAL component list, not in the REQUIRED component list.

robertosfield:
Does your commit affect older versions of CMake?  Does it gracefully fails when WxWidgetsi isn't installed.

The OSG only uses WxWdigets for a single example that illustrates integration between OSG and WxWidgets so it's important that WxWidgets remains only an optional dependency.

PR #780, opened 2019-07-02
Fix wxWidgets COMPONENTS specification, deprecated usage broken with CMake 3.14

StefanBruens:
Current FindWxWidgets.cmake allows to specify OPTIONAL components, unfortunately
this broke the (deprecated) use of wxWidgets_USE_LIBS.

Fixes: #779

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -789,9 +789,7 @@ IF(BUILD_OSG_EXAMPLES AND NOT ANDROID)
 
     FIND_PACKAGE(FLTK)
     FIND_PACKAGE(FOX)
-
-    SET(wxWidgets_USE_LIBS base core gl net)
-    FIND_PACKAGE(wxWidgets)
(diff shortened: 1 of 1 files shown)


robertosfield:
Two questions, which versions of CMake will this work for, and how gracefully does it fail when WxWdigets isn't installed?  The later is important as WxWdigets is only required for a single example program and is irrelevant to vast majority of OSG users.  Thanks.

StefanBruens:
USE_LIBS has been deprecated 12 years ago already (2.6.0):
https://github.com/Kitware/CMake/commit/bfeeb88b1f32086885ee02974bf54c28b0a4b9ab

It just reports the missing wxWidgets and continues:
> Could NOT find wxWidgets (missing: wxWidgets_LIBRARIES wxWidgets_INCLUDE_DIRS)

robertosfield:
Thanks for the explanation, fix now merged with master and the 3.6 branch

PR #781, opened 2019-07-03
make FrameBufferObject::releaseGLObjects call releaseGLObjects on it's _attachments

LaurensVoerman:
Hi Robert,
At the moment I wanted to report the 3.6.4 version on windows has been without problems for me I did a final update... and got a crash on program exit.
To problem I see is the destructor ~GraphicsObjectManager() trying to log an OSG_INFO message after the 
static NotifySingleton s_NotifySingleton
was already destroyed, causing an invalid pointer dereference. 
This happens in the osg159-osgd.dll!`dynamic atexit destructor for 's_contextIDMap''()	

A very naive fix would be to comment out the OSG_INFO message in the GraphicsObjectManager destructor (OpenSceneGraph\src\osg\GLObjects.cpp line 70) which might also be a good idea, but I did a bit of digging to get to the source of the problem, and came up with these results.

When I close the programs' window, the graphics context is destroyed first. The ContextData in the s_contextIDMap is destroyed in ContextData::decrementContextIDUsageCount as it's getNumContexts() now returns zero.
But when the viewer and scenegraph are destroyed, the scenegraph objects cause a new ContextData to be made in the s_contextIDMap in order to cue the deletion of RenderBuffer's openGL objects.

While I have managed to destroy the ContextData in the s_contextIDMap with an explicit call to
osg::ContextData::decrementContextIDUsageCount(0);
at the end of my program (after releasing all ref_ptrs to all nodes and shaders) I think the real solution is different:

the FrameBufferObject needs to call releaseGLObjects on it's attachments. For symetry I added the same calls for resizeGLObjectBuffers - as I think that's appropriate.
Regards, Laurens

### DIFF
--- a/src/osg/FrameBufferObject.cpp
+++ b/src/osg/FrameBufferObject.cpp
@@ -594,6 +594,14 @@ void FrameBufferObject::resizeGLObjectBuffers(unsigned int maxSize)
     _fboID.resize(maxSize);
     _unsupported.resize(maxSize);
     _fboID.resize(maxSize);
+    for (AttachmentMap::iterator i = _attachments.begin(); i != _attachments.end(); ++i)
+    {
+        FrameBufferAttachment &fa = i->second;
+        Texture* tex = fa.getTexture();
+        if (tex) tex->resizeGLObjectBuffers(maxSize);
+        RenderBuffer* rb = fa.getRenderBuffer();
+        if (rb) rb->resizeGLObjectBuffers(maxSize);
+    }
 }
 
 void FrameBufferObject::releaseGLObjects(osg::State* state) const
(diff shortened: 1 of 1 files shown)


openscenegraph:
Hi @LaurensVoerman, I've just done a first pass review and feel that resizeGLObjects() and releaseGLObjects() implementations should be moved into the FrameBufferAttachement class to encapsulate this functionality better.  I will have a look at doing this.

I will have a think about the other issues.  Do you have example code I could use to reproduce the crash?

openscenegraph:
This is the commit that adds FrameBufferAttachment::resizeGLObjectBuffers(..) and releaseGLObjects(..) : 

    https://github.com/openscenegraph/OpenSceneGraph/commit/079cf6c090958ef62edb21f6605159666246835b

LaurensVoerman:
Hi Robert,
the osgdefered example can demonstrate the crash on my machine (windows 10)
if I set
OSG_NOTIFY_LEVEL=INFO
As this depends on the order of destruction of static objects
( s_NotifySingleton and s_contextIDMap) I am not sure this is reproducable
on other machines.
Laurens.

openscenegraph:
I have now removed the debug messages from GraphicsObjectManager:

 https://github.com/openscenegraph/OpenSceneGraph/commit/e327fc16df396f2d6591a874224e56dfd68d405f

openscenegraph:
@LaurensVoerman I'm going to close this PR as I've checked in similar functionality that should hopefully resolve the issues you're seeing but as I don't have an example that reproduces the crash I'll have to leave this to you to test and provide feedback.

openscenegraph:
I just tried osgdefered on Linux with the latest changes and it runs and exits just fine for my under Kubuntu 18.04 with notify level set to info, I don't know if it would have failed before though.

PR #786, opened 2019-07-15
Fix 'Incorrect default location for true type fonts on openSUSE'

### LINKED ISSUE #778: Incorrect default location for true type fonts on openSUSE
At https://github.com/openscenegraph/OpenSceneGraph/blob/master/include/osgText/Font#L45 directory paths are listed for True Type fonts used on non-Windows operating systems. At least for openSUSE this does not fit, because the default location /usr/share/fonts/truetype is used.

rhabacker:
If fontconfig was found when compiling osg, osgText will first search for font names in the paths provided by fontconfig. This commit adds a new optional dependency to Fontconfig.

Fixes #778

### DIFF
--- a/src/osgText/CMakeLists.txt
+++ b/src/osgText/CMakeLists.txt
@@ -1,3 +1,4 @@
+find_package(Fontconfig MODULE)
 
 IF(DYNAMIC_OPENSCENEGRAPH)
     ADD_DEFINITIONS(-DOSGTEXT_LIBRARY)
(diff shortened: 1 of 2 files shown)


openscenegraph:
I merged and did some quick testing but didn't spot the following error:

CMake Warning at src/osgText/CMakeLists.txt:1 (find_package):
  No "FindFontconfig.cmake" found in CMAKE_MODULE_PATH.

My cmake-version is 3.10 and it doesn't have a FindFontconfig.cmake.  We need to find a more graceful way of handling this as getting errors on an optional component will make users think there is an issue that needs resolving.

rhabacker:
~/src/openscenegraph-3.6> rpm -q -f /usr/share/cmake/Modules/FindFontconfig.cmake
cmake-3.14.5-376.1.x86_64

It looks that fontconfig was added with cmake version 3.14 (see https://github.com/Kitware/CMake/commit/84e7920b3ac9241cf73a468120ff0ea83f9a7c86#diff-669d79ed90b41640c1db1c24796f1354).

A copy could be embedded to osg sources.

openscenegraph:
I've added a QUIET to the find_package and this addresses the noisy warning. Feel free to generate a PR with the addition of the FindFontconfig.cmake if think it would be useful.

PR #789, opened 2019-07-15
cmake: Add find_package module for fontconfig

rhabacker:


openscenegraph:
I've merged but need to follow up on the addition of the use the FeatureSumary feature as I haven't yet established how far back cmake has this, so far a quick google search has come across a cmake 3.0.2 reference to it, but don't if dates back to 2.8.0 yet (this is the min version that OSG now supports.)

openscenegraph:
I have just checked out cmake git repository and the v2.8.0 has the FeatureSummary.cmake in it so it looks like it'll be fine.

PR #790, opened 2019-07-16
osgUtil::IntersectionVisitor Select only the currently visible PagedLOD

longhuan2018:
support Select only the currently visible PagedLOD node

### DIFF
--- a/src/osgDB/OutputStream.cpp
+++ b/src/osgDB/OutputStream.cpp
@@ -581,6 +581,7 @@ void OutputStream::writeImage( const osg::Image* img )
         case IMAGE_INLINE_FILE:
             if ( isBinary() )
             {
+                bool bNeedStream = true;
                 std::string fullPath = osgDB::findDataFile( img->getFileName(), _options.get() );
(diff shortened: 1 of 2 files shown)


openscenegraph:
I've done a first pass review of the code and am not happy with how extensive the changes are, it's hard to see exactly what is done for what reason, it ends up being hard to know whether any of the changes are suitable or robust.  To achieve the functionality you are after I don't think this is the right way to go about it.  

If we need this functionality it's something for us to discuss on the osg-users mailing list or googlegroups forum,

longhuan2018:
For the modification of osgUtil::IntersectionVisitor, I want to implement the PagedLOD relative to the Intersect implementation in USE_EYE_POINT_FOR_LOD_LEVEL_SELECTION mode. It is usually useful for checking 3D reality modeling generated by software such as Smart3D (ContextCapture) and Pix4D, because they generate some floating objects or  holes, but we can't know the pagedlod file corresponding 
 to the data displayed on the current screen. Solved (because it always gets the highest data)

PR #791, opened 2019-07-16
Avoid that mingw cross-compilation choosing X11 as the windowing system.

aluaces:
I have moved the `PROJECT` stanza before the conditionals, because
otherwise expressions like `IF(MINGW)` always fail.


openscenegraph:
Thanks, merged with master as well.

PR #802, opened 2019-07-26
OSG_TEXT_USE_FONTCONFIG cmake option added: Set to ON to enable attem…

valid-ptr:
…pt to use FontConfig in osgText; Disabled by default for iOS and Android

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,6 +501,12 @@ ENDIF()
 
 OPTION(OSG_GL_LIBRARY_STATIC "Set to ON to statically link with OpenGL/GLES library." OFF)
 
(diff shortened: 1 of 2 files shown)


valid-ptr:
DO NOT MERGE

openscenegraph:
The changes look pretty sensible, so I presume the "DO NO MERGE" relates to it being a proposal and still having debug messages in place for testing.

I'd be happy to merge a cleaned up version.

valid-ptr:
I don't know why, but Fontconfig_FOUND was always TRUE, no matter
find_package(Fontconfig MODULE) presented or not in osgText/CMakeLists.txt

Maybe cached by another module like freetype or some....

I've a added additional check in the second commit.
Now everything looks perfect :)

PR #805, opened 2019-07-26
give the good camera to OcclusionQueryNode

mp3butcher:


openscenegraph:
??  No clue why you've made this changes. seem pretty random and with no understandable explanation why you think it's appropriate.

FYI, CullVisitor::getCurrentCamera() is implementation as:

        inline osg::Camera* getCurrentCamera()
        {
            return getCurrentRenderStage()->getCamera();
        }

Which is what you substituted.  So  i am going to close this PR safe in the knowledge that it's pointless.

mp3butcher:
oups i've made a mistake I wanted to write
getRenderStage()->getCamera()

mp3butcher:
...taking current renderstage yields weird behavior with shadowcams

mp3butcher:
...surely a bug in openmw shadow implementation...forget it

Issue #810, opened 2019-08-11
OcclusionQueryNode _validgeometry breaks backward compatibility


mp3butcher:
query geometry validity added in https://github.com/openscenegraph/OpenSceneGraph/commit/71997fa747f46e54847b2843227d003fb49bb872#diff-5c9d7ea871e2f1816d216fb56fc1a733
can break client code if querygeometry is set manually by user.
Nothing that can't be worked around but still....

openscenegraph:
Ahhhhh... different usage cases that are both fighting each other.  I don't have either usage case to review so can't provide any guidance at this point.  Simply revert the commit will break on usage case in favour on another.  

This does occasionally happen with OSG features and is typically an indication that the underlying design and implementation isn't flexible enough.  It's made more difficult for me to resolve by myself when the original functionality isn't written by me, and it's now a feature I've used another than using the osgocclusionqueury example.  This has meant I've had to rely on those who use the feature to determine what it needs to be fit for purpose.

Ideally we'd have the different usage cases as separate code paths in the OSG example or multiple examples that illustrate the different usage cases so that we can ensure that any changes don't break that particular usage case.  

Hopefully @dan-t who authored the above commit can chip in here.  

I'm open to suggestions on how we can evolve the design+implementation work well in the various usage cases that user require.  At this point I can't do anything as I simply don't have anything to go on, all I know is that one commit required to fix one usage case breaks another.

dan-t:
Ok, for the use case of an user definable query geometry the current API of
`OcclusionQueryNode` is really a bit strange.

I already considered it a bit strange that `QueryGeometry` is accessible
and writable by the user and didn't quite saw the point of it, well, now
I'm seeing it.

A small fix with minimal backward compatibility breakage might be to
remove `osg::QueryGeometry* getQueryGeometry()` and only keep
`const osg::QueryGeometry* getQueryGeometry() const`.

Now add a `void setQueryGeometry(osg::QueryGeometry* geom)` method
that in the case of `geom != nullptr` sets `_validQueryGeometry = true`.

But I don't have the time and use case to test this change by myself.

dan-t:
Thinking a bit further, I don't think that the current implementation of
`OcclusionQueryNode::computeBound` does really handle the user
definable query geometry correctly, because it just overrides the vertex
array of the query geometry, without considering if the geometry was
set by the user.

dan-t:
[...]

PR #811, opened 2019-08-12
Fix backward compatibility of occlusionquerynode

mp3butcher:
a proposal for https://github.com/openscenegraph/OpenSceneGraph/issues/810

### DIFF
--- a/include/osg/OcclusionQueryNode
+++ b/include/osg/OcclusionQueryNode
@@ -120,12 +120,10 @@ public:
 
     virtual void releaseGLObjects( osg::State* state = 0 ) const;
 
-
     // When disabled, OQN doesn't perform occlusion queries, and simply
     //   renders its children.
     void setQueriesEnabled( bool enable=true );
     bool getQueriesEnabled() const { return _enabled; }
-
 
     // Sets/gets the visibility threshold. If the test indicates that
     //   the number of visible pixels is less than the specified
@@ -148,7 +146,6 @@ public:
     void setDebugDisplay( bool enable );
     bool getDebugDisplay() const;
 
-
     // Set and get the StateSet used by the OcclusionQueryNode
     //   when rendering the query geometry. OQN creates its own by
     //   default, but if you use many OQNs you might want to use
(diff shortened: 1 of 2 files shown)


mp3butcher:
@dan-t says
> I don't quite see the point for the compute bound callback, the
> `_validQueryGeometry` flag is still needed and it IMHO just further
> complicates the design.
> 
The point of the callback is to override it easily without introducing a DummyComputeBoundCallback in custom path
> The non const `getQueryGeometry` method should be removed, otherwise
> the design is just broken and already was before the addition of
> the `_validQueryGeometry` flag.
> 
Would rather let not to break backward compatibility with client usage
> The `Custom` in `setCustomQueryGeometry` doesn't add any further
> meaningful information, if the getter of the query geometry is named
> `getQueryGeometry` then the setter should be just named `setQueryGeometry`.

promoting QueryGeometry from OQN attribute to real OQN property (complying with metamodel conventions) could break backwardcompatibility

> 
> The `setQueryGeometry` should handle a nullptr geometry and in this
> case switch back to its own internal query geometry.

I named it setcustomgeomety not take querygeometry as an actual property

mp3butcher:
@dan-t 
You could propose a better alternative as your the author of incriminated commit.
Personally I haven't anything against the current design, so my motivation for conflict around that is minimal....

mp3butcher:
i removed the callback stuff ... too much chance to break client code

include/osg/OcclusionQueryNode
```diff
@@ -158,6 +155,10 @@ public:
     osg::StateSet* getQueryStateSet();
     const osg::StateSet* getQueryStateSet() const;
 
+    // override construction of querygeometry based on child bounds
+    void setCustomQueryGeometry( osg::QueryGeometry* qg);
+    const osg::QueryGeometry* getCustomQueryGeometry() const { if (getComputeBoundingSphereCallback()) return getQueryGeometry(); return 0; }
+
```
mp3butcher: Just assume it's a custom query geometry if a computeboundcallback has been setted

Issue #812, opened 2019-08-13
OSG 3.6.4 RPM build problem


DieSlower:
It seems an error was introduced into OSG 3.6.4 when  OpenSceneGraph/CMakeModules/FindFontconfig.cmake  was added to the build. Fortunately it is an easy fix, and does not touch actual code. 
It is easy to miss, I have tested this on Windows and Linux (CentOS 7). Visual Studio and GCC build OSG just fine, but when on linux you try to package OSG for product delivery with RPMBuilder v4.11.3 (standard linux RPM building tool) you get an error. 

``
CMake Error: Error in cmake code at:
/home/<username>/rpmbuild/BUILD/OpenSceneGraph-OpenSceneGraph-3.6.4/CMakeModules/FindFontconfig.cmake:6:
Parse error. Expected "(", got newline with text "
".
``
The error comes from using uncommon block comments:
#[============[ and
#]============]

If this block comment is replaced with normal CMake # comments, everything seems to start working again. 

I can submit a PR for this, or if you want, you can just replace the comment code in the file. 

Let me know,
--MS

openscenegraph:
Please change your github handle it's still just plain obnoxious.  You might not care what other people think of you, but I care what people think of the OSG and it's community.

PR #818, opened 2019-08-15
fix comment block CMake syntax

mp3butcher:
https://github.com/openscenegraph/OpenSceneGraph/issues/812

### DIFF
--- a/CMakeModules/FindFontconfig.cmake
+++ b/CMakeModules/FindFontconfig.cmake
@@ -1,7 +1,6 @@
 # Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
 # file Copyright.txt or https://cmake.org/licensing for details.
 
-#[=======================================================================[.rst:
 FindFontconfig
 --------------
 
@@ -29,9 +28,6 @@ This will define the following variables in your project:
 ``FONTCONFIG_COMPILE_OPTIONS``
   this should be passed to target_compile_options(), if the
   target is not used for linking
-
-#]=======================================================================]
-
 
 # use pkg-config to get the directories and then use these values
 # in the FIND_PATH() and FIND_LIBRARY() calls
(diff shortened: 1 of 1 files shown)


DieSlower:
@mp3butcher  
Hi Julien,
Thanks for taking this on. 
The block comment needs to be replaced with #s
So each line between the block comments needs to have a leading # for this to work.

mp3butcher:
no ... it was just a cmake block syntax

DieSlower:
#FindFontconfig
#--------------

#Find Fontconfig headers and library.

#Imported Targets
#^^^^^^^^^^^^^^^^

#``Fontconfig::Fontconfig``
# The Fontconfig library, if found.

#Result Variables
#^^^^^^^^^^^^^^^^

#This will define the following variables in your project:

#``FONTCONFIG_FOUND``
#  true if (the requested version of) Fontconfig is available.
#``FONTCONFIG_VERSION``
#  the version of Fontconfig.
#``FONTCONFIG_LIBRARIES``
#  the libraries to link against to use Fontconfig.
#``FONTCONFIG_INCLUDE_DIRS``
#  where to find the Fontconfig headers.
#``FONTCONFIG_COMPILE_OPTIONS``
# this should be passed to target_compile_options(), if the
#target is not used for linking

DieSlower:
@mp3butcher 
Hmm...Just tested the new comments
#[[=======================================================================.rst:
Text
=======================================================================]]
with RPMBuilder and getting the same error...

mp3butcher:
@DieSlower  cmake version?

DieSlower:
Actually, looks like we also had a problem with some code on our end that exposed this bug. 
But now that it's fixed, your change works. 

Thanks Julien!

mp3butcher:
..so is this pr required?

DieSlower:
Yes, with the fix that you created.

PR #820, opened 2019-08-17
DDS plugin: add support for DDPF_PALETTEINDEXED8 pixel format reading

Capostrophic:
So I've added in support for reading this rather obscure format by reconstructing the image data into a full-sized RGBA image. Used the GIMP DDS plugin for reference, apparently the 256 color palette is put right after the header. Alpha flag is never used.

I used these images to test it.
[examples.zip](https://github.com/openscenegraph/OpenSceneGraph/files/3511525/examples.zip)

### DIFF
--- a/src/osgPlugins/dds/ReaderWriterDDS.cpp
+++ b/src/osgPlugins/dds/ReaderWriterDDS.cpp
@@ -222,6 +222,7 @@ struct DXT1TexelsBlock
 //
 #define DDPF_ALPHAPIXELS        0x00000001l
 #define DDPF_FOURCC             0x00000004l        // Compressed formats
+#define DDPF_PALETTEINDEXED8    0x00000020l
 #define DDPF_RGB                0x00000040l        // Uncompressed formats
 #define DDPF_ALPHA              0x00000002l
 #define DDPF_COMPRESSED         0x00000080l
@@ -944,6 +945,12 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
             internalFormat = GL_ALPHA;
             pixelFormat    = GL_ALPHA;
     }
+    else if (ddsd.ddpfPixelFormat.dwFlags & DDPF_PALETTEINDEXED8)
+    {
+            OSG_INFO << "ReadDDSFile info : format = PALETTEINDEXED8" << std::endl;
+            // The indexed data needs to first be loaded as a single-component image.
+            pixelFormat = GL_RED;
+    }
     else
     {
         OSG_WARN << "ReadDDSFile warning: unhandled pixel format (ddsd.ddpfPixelFormat.dwFlags"
@@ -984,6 +991,17 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
 
    OSG_INFO<<"ReadDDS, dataType = 0x"<<std::hex<<dataType<<std::endl;
 
+    unsigned char* palette = new unsigned char [1024];
+    if (ddsd.ddpfPixelFormat.dwFlags & DDPF_PALETTEINDEXED8)
+    {
+        if (!_istream.read((char*)palette, 1024))
+        {
+            delete [] palette;
+            OSG_WARN << "ReadDDSFile warning: couldn't read palette" << std::endl;
+            return NULL;
+        }
+    }
+
     unsigned char* imageData = new unsigned char [sizeWithMipmaps];
     if(!imageData)
     {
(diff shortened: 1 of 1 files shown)


src/osgPlugins/dds/ReaderWriterDDS.cpp
```diff
@@ -1010,7 +1028,27 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
...
+            convertedData[i * 4 + 0] = palette[imageData[i] * 4 + 0];
+            convertedData[i * 4 + 1] = palette[imageData[i] * 4 + 1];
+            convertedData[i * 4 + 2] = palette[imageData[i] * 4 + 2];
+            convertedData[i * 4 + 3] = palette[imageData[i] * 4 + 3];
+        }
+        for (unsigned int i = 0; i < mipmap_offsets.size(); i++)
+            mipmap_offsets[i] *= 4;
+        internalFormat = GL_RGBA;
+        pixelFormat = GL_RGBA;
+        osgImage->setImage(s,t,r, internalFormat, pixelFormat, dataType, convertedData, osg::Image::USE_NEW_DELETE, packing);
```
akortunov: What's about `imageData` array? I suppose we should free it in this scope to avoid a memory leak.

Capostrophic: Yeah, corrected. I'm too used to automatic memory allocation and deallocation. :smile:

src/osgPlugins/dds/ReaderWriterDDS.cpp
```diff
@@ -1010,7 +1027,28 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
...
+    {
+        // Now we need to substitute the indexed image data with full RGBA image data.
+        unsigned char* convertedData = new unsigned char [sizeWithMipmaps * 4];
+        for (unsigned int i = 0; i < sizeWithMipmaps; i++)
+        {
+            convertedData[i * 4 + 0] = palette[imageData[i] * 4 + 0];
+            convertedData[i * 4 + 1] = palette[imageData[i] * 4 + 1];
+            convertedData[i * 4 + 2] = palette[imageData[i] * 4 + 2];
+            convertedData[i * 4 + 3] = palette[imageData[i] * 4 + 3];
+        }
```
mp3butcher:
```
  unsigned char* convertedData = new unsigned char [sizeWithMipmaps * 4];
        unsigned char* pconvertedData = convertedData;
        unsigned char *pimageData = imageData;
        unsigned char *ppalet;
        for (unsigned int i = 0; i < sizeWithMipmaps; i++)
        {
ppalet=&palette[(*pimageData) * 4];
(*pconvertedData++) = *ppalet++;
(*pconvertedData++) = *ppalet++;
(*pconvertedData++) = *ppalet++;
++pimageData;
        }
```

Capostrophic: This seriously inhibits readability to questionable benefit. On my end, up to -O2 GCC either produces the exact same code or it's somehow even more complex with pointer arithmetics, while -O3 optimizes index-based loop even better than pointer-based loop. Not worth it.

mp3butcher: I'd never count on compiler to do what i want...most of the time it didn't...but it's your call

Capostrophic: Trying to outsmart the compiler isn't a brilliant idea.

akortunov: In most of cases compilers developers know C++ better than other developers, so usually you can prefer code readability over low-level optimizations.

mp3butcher: https://www.youtube.com/watch?v=x61H6qEtK08

src/osgPlugins/dds/ReaderWriterDDS.cpp
openscenegraph: In general I prefer readability over hypothetical code optimizations, I say hypothetical as compiler optimizer are pretty good.  It's also really important to know the roll of the code - it's it code that is on an important inner loop of rendering that could make a difference to fps, or is it an ancillary feature that rarely gets called.  

For that really performance critical it's more warrented to spend time and optimize carefully and be willing to compromise on readability for these improvements.  For the dds plugin, it's something run at start up prior to rendering, and only be a small number of users, and those time it's used during rendering is safely tucked away in a background database thread so... never a performance critical feature.

So it comes down to which version is more readable.  I'm so used to unrolling pointer management like @mp3butcher's variant that I find it just as readable as the original, if not more so.  I'm pretty old school though :-)

mp3butcher:
Capo Delta t2-t1: 5861171 nanoseconds
MY Delta t2-t1: 2610880 nanoseconds

PR #822, opened 2019-08-17
optimize indexed color DDS conversion

mp3butcher:
Before : 3.6 ms
After    : 1.3 ms
(they are means over 1000 tries with ubuntu gcc 7.4.0 )

### DIFF
--- a/src/osgPlugins/dds/ReaderWriterDDS.cpp
+++ b/src/osgPlugins/dds/ReaderWriterDDS.cpp
@@ -990,16 +990,12 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
     }
 
     OSG_INFO<<"ReadDDS, dataType = 0x"<<std::hex<<dataType<<std::endl;
-    typedef union {
-        float f32;
-        unsigned char c8[4];
-    } PaletteWord;
 
-    PaletteWord palette [256];
+    unsigned char palette [1024];
 
     if (ddsd.ddpfPixelFormat.dwFlags & DDPF_PALETTEINDEXED8)
     {
-        if (!_istream.read((char*)&palette->c8, 1024))
+        if (!_istream.read((char*)palette, 1024))
         {
             OSG_WARN << "ReadDDSFile warning: couldn't read palette" << std::endl;
             return NULL;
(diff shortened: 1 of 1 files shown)


Capostrophic:
If we're going there...
```cpp
unsigned char* convertedData = new unsigned char [sizeWithMipmaps * 4];
unsigned char* dest = convertedData;
unsigned char* index = imageData;
for (unsigned int i = 0; i < sizeWithMipmaps; i++)
{
    unsigned char* color = &palette[(*index++)*4];
    for (int j = 0; j < 4; j++)
        *dest++ = *color++;
}
```

src/osgPlugins/dds/ReaderWriterDDS.cpp
```diff
@@ -1031,12 +1031,15 @@ osg::Image* ReadDDSFile(std::istream& _istream, bool flipDDSRead)
...
+        unsigned char *pimageData = imageData;
+        unsigned char *ppalet;
         for (unsigned int i = 0; i < sizeWithMipmaps; i++)
         {
-            convertedData[i * 4 + 0] = palette[imageData[i] * 4 + 0];
-            convertedData[i * 4 + 1] = palette[imageData[i] * 4 + 1];
-            convertedData[i * 4 + 2] = palette[imageData[i] * 4 + 2];
-            convertedData[i * 4 + 3] = palette[imageData[i] * 4 + 3];
+            ppalet = &palette[(*pimageData)*4];
+            memcpy(pconvertedData, ppalet, 4);
```
mp3butcher: memcpy can be replace with 
`                   *((unsigned int *)pconvertedData)=*((unsigned int *)(ppalet));`
but it's not very pretty

Capostrophic: Then I propose [this](https://github.com/Capostrophic/OpenSceneGraph/commit/028c7edf8bf8304fe448606ca898cb6dce788ee9).

mp3butcher:
... considering adding -Wstrict-aliasing option might be a good thing...

mp3butcher:
@robertosfield the last commit is left to your consideration

openscenegraph:
This PR has got a bit silly with 7 different commits thrashing between different versions.  The last one looks like it might OK and is the most readable, has this one been profiled against the original version?  It seems like this last rev should be just submitted in it's own PR as I don't want all the intermediate commits merged as well.

mp3butcher:
> This PR has got a bit silly with 7 different commits thrashing between different versions. The last one looks like it might OK and is the most readable, has this one been profiled against the original version? 

Indeed, profiler timings are between master implementation and this pr

> It seems like this last rev should be just submitted in it's own PR as I don't want all the intermediate commits merged as well.

You said the lastone https://github.com/openscenegraph/OpenSceneGraph/pull/822/commits/81bb7f2bf0a9c4be7742efe49353943666429d84 might be ok and readable, and now you says this commit should be object of an other pr: I don't follow you.

Should remove the last commit and squash all others in a single commit?

openscenegraph:
I don't want the history of all your experiments.  The last incarnation of the code is what I want, this should be a minimal diff over the original code.

mp3butcher:
ok i clarified this pr with 2 commits:
- first one use union to take care of type aliasing
- second use memcpy instead (not sure about the aliasing union trick)

PR #823, opened 2019-08-18
Debugging: Hint to debug LoadLibrary issues

remoe:
This helps to debug loading dynamic libraries in an environment without implemented "dlopen". For example emscripten with static compiled OSG.


sbrkopac:
While I agree that there needs to be some mechanism to let the user know when that call fails (I've been burned by this before), adding an output there would bloat a log file extremely quickly. IMO it's worth considering keeping track of failed libraries and then only printing the message once - potentially only trying to load a failed library once as well.

remoe:
This will only log when one set notify level to DEBUG_INFO.

sbrkopac:
Yup. I just think we could, potentially, take a look at making this whole part of dynamic library loading a bit better.

PR #825, opened 2019-08-22
added "reserve" before adding multiple elements to vectors

Vic-Min:
small optimization to reduce memory reallocation

### DIFF
--- a/src/osg/ClipNode.cpp
+++ b/src/osg/ClipNode.cpp
@@ -55,6 +55,7 @@ void ClipNode::setReferenceFrame(ReferenceFrame rf)
 void ClipNode::createClipBox(const BoundingBox& bb,unsigned int clipPlaneNumberBase)
 {
     _planes.clear();
+    _planes.reserve(6);
     if (!_stateset.valid()) _stateset = new osg::StateSet;
 
     _planes.push_back(new ClipPlane(clipPlaneNumberBase  ,1.0,0.0,0.0,-bb.xMin()));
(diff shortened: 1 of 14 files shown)


openscenegraph:
I've gone through all the changes and they look safe.  However, I'm not convinced that it's a good programming practice to go through code like this and make changes on the off chance that it might help.  

Each change we make to the code we risk introducing bugs, and each change like this specifically leaves the code open to code changes elsewhere in the code making this changes inappropriate - for instance if a for loop changed what it does but the programmer didn't spot the reserve then we could easily end up with an inefficiency that's been introduced without them realizing.

To mitigate the risks of an error being introduced and being hard to revert without impacting things down the line it's also far better to have multiple PR's so that groups of related changes get checked separately, then later if there is an issue one set of commits can be reverted.

Rather than like a bad practice slip in I'm going to reject this PR. Could you break it up into separate commits, with groupings done on library affected then at least we'll have a better chance of making sure if we do find problems we can revert in a fine grained way.

The other thing to be wary of how justified are many of these changes?  Is it going to have any real tangible improvement to performance for the extra complexity introduced.  Evey single change we need to look at from this perspective.  Yes in theory it's a sensible change, but in practice does it make a measurable difference that justifies it's existence, of not then it makes no sense to apply it.  A number of these changes look to me high risk for getting things wrong and low value in terms of potential performance gain.  

So when you look through the changes perhaps even sub group them into ones you are confident are a real help to performance, and keep these separate from the ones that are less likely - such as for niche features that are only called once in the lifetime of an application.

PR #826, opened 2019-08-22
add custom uniform locations

mp3butcher:
enable support of uniform location not reported by glGetProgramiv(  GL_ACTIVE_UNIFORMS)
support array element operator "[]" and the structure field operator "."

should be added to serializer but would involve SO version increment

note : this pr is against 3.6 branch and is intended to be merge with master too

### DIFF
--- a/include/osg/Program
+++ b/include/osg/Program
@@ -156,8 +156,6 @@ class OSG_EXPORT Program : public osg::StateAttribute
         ShaderDefines& getShaderDefines() { return _shaderDefines; }
         const ShaderDefines& getShaderDefines() const { return _shaderDefines; }
 
-
-
         /** Simple class for wrapping up the data used in glProgramBinary and glGetProgramBinary.
           * On the first run of your application Programs should be assigned an empty ProgramBinary.
           * Before your application exits it should retrieve the program binary via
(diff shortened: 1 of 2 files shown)


openscenegraph:
Sorry for slow review, been busy with the VSG.

openscenegraph:
I've done a first pass review and kinda perplexed by what you are doing and why.  I don't know how you plan to use it and why it's needed.  Could you provide a usage example so I can see what you are attempting to do and how this change enables it.  It could be there is a better way of implementing it, and if not we really need an example because it's not at all obvious what this does and how to use i.

mp3butcher:
in osg, all Uniform names must match names reported by calling glGetProgramiv( GL_ACTIVE_UNIFORMS), if no match found the Uniform is ignored.
This is not flexible enough to embed uniform target location of Uniform whose name contains syntaxic operator like [] or .
This behavior forbid strategy such as update small part of a big uniform during drawing.

mp3butcher:
As an example, i can cite openmw lighting strategy: consisting of culling lightsource base on their distance to a given object.
LightsUniform is a uniform array of lightcount* light param size

Light culling results along scene objects traversal don't change a lot so instead of reuploading the whole uniform at each object traversal, I only reupload part of the lightuniform.

This usecase might not be enough to motive the changes, as light uniform is quite a small to upload. 
I'm not even sure it improve performance in that particular case.
However:
- when uniform become larger. this micromanagement can surely lead to bandwith saving
- giving the abitlity to the user to specify its own uniform target is a good thing...I don't even understand why osg do name matching glGetProgramiv( GL_ACTIVE_UNIFORMS) instead of simply use glGetUniformLocation with the uniform name...so I added it as additional feature

openscenegraph:
In my own testing I have found the cost of bandwidth is essentially negligible compared to the cost of OpenGL calls to do the copy, as the overhead cost swaps the former until you get large amounts of data.  If you data is kK in size I would recommend against making things complicated by come up with schemes that copy portions of objects.  

If you are getting into Mb size then it could be worth trying to copy in a more fine grained way.

As you haven't done any performance testing on it, I would suggest you go an do it.  I am not about to start merging extra complexities into the OSG code base just because someone thought that something might be a bottleneck in a niche usage case, but never actually benchmarked it.

I don't recall any glGetProgramiv vs glGetUniformLocation decision, but that the GLSL code was written a looooong time ago, and not by me.  It may be tht glGetUniformLocation wasn't in the original spec.  It could have been another reason.

As I don't have any concrete code to review the usage case in front of me I can't make any decisions on the topic - I have many other PR's and Issues to look at so will get on with these.

As a general comment, I want to avoid changing the API with the 3.6.x series.  New features should go into master.

mp3butcher:
could simply getting rid of the uniform name matching be wiser? If you don't know why it's here then none do

mp3butcher:
...It's been a while since i do this pr , but i can't remember if name matching is easily removable but i can try if you think it worse the time spent

openscenegraph:
The 3.6 branch is in *maintenance* mode.  We are not about to rip stuff out an replace stuff.

If you want to propose changes then these will need to be in master.

PR #832, opened 2019-08-30
WIP: EGL support for Windows. Allow to run OSG with Google/Angle

remoe:
Allow to run OSG with Google/Angle as a render backend. Tested with GLES3 on DirectX11 and Vulkan - Angle backend. 

Example: osgsimpleGLES3 

Current limitations: 
- GraphicContext traits only rudimentary ported.
- need to set "SingleThreaded" to viewer threading model.


remoe:
Someone has tried it :sunglasses:  
Forum link: http://forum.openscenegraph.org/viewtopic.php?t=17885

PR #834, opened 2019-09-03
getInternalTextureFormat() failed after reading a tiff file

0871087123:
As the title.
Code like this:
```
...
tex = new osg::Teture2D();
tex->setImage(osgDB::readImageFile("x.tiff"));
...
printf("%u\n", tex->getImage()->getInternalTextureFormat());
```

The value is 3.

So i checked the plugin tiff, found that using number of components instead of which it should be.
Also plugin tga has similar problem.
I try to fix plugin tiff first, then back to work, bypassing this problem manually.


openscenegraph:
Apologies for the slow rely, been busy with the VSG.

openscenegraph:
I have done a fist pass review and believe the problem with the internalFormat is down to recent drivers not supporting use of component number for internal format.  I think the best solution would be to create a static osg::Image function to compute the intnernal format based on the PixelFormat and DataType.  I will have a think about this.

PR #854, opened 2019-11-04
Fix bug in matrix transpose() and transpose3x3

timoore:
The diagonal elements were not copied from the source matrix.


dliepelt:
It’s not a bug, it’s a feature... after transposing a matrix, the diagonal elements will stay the same.

timoore:
That is only true if you transpose a matrix in place. If the source is
another matrix then the destination diagonals are not updated.

Tim

dliepelt:
true.... 👍

PR #857, opened 2019-11-25
use vertex length as sort hash to decrease undetected vertices duplicates

mp3butcher:
....during mesh indexation (IndexMeshVisitor)
Not a perfect solution (duplicate might not share same dist plage) but seems to works better than original (sort on XthenYthenZ only get Z duplicates)
I test with arbitrary meshes coming from Marching Cube of a signed distance map and index with this pr seams lot nicer (approx duplicates/20)

### DIFF
--- a/src/osgUtil/MeshOptimizers.cpp
+++ b/src/osgUtil/MeshOptimizers.cpp
@@ -52,29 +52,29 @@ typedef std::vector<unsigned int> IndexList;
 // ArrayVisitor on them.
 struct GeometryArrayGatherer
 {
-   typedef std::vector<osg::Array*> ArrayList;
+    typedef std::vector<osg::Array*> ArrayList;
 
-   GeometryArrayGatherer(osg::Geometry& geometry)
-   {
-       _vertexarray = dynamic_cast<osg::Vec3Array*>(geometry.getVertexArray());
-       if(!_vertexarray.valid()) add(geometry.getVertexArray(), osg::Array::BIND_PER_VERTEX);
-       add(geometry.getNormalArray());
-       add(geometry.getColorArray());
-       add(geometry.getSecondaryColorArray());
-       add(geometry.getFogCoordArray());
-       unsigned int i;
-       for(i=0;i<geometry.getNumTexCoordArrays();++i)
-       {
-           add(geometry.getTexCoordArray(i), osg::Array::BIND_PER_VERTEX);
-       }
-       for(i=0;i<geometry.getNumVertexAttribArrays();++i)
-       {
-           add(geometry.getVertexAttribArray(i));
-       }
-   }
+    GeometryArrayGatherer(osg::Geometry& geometry)
+    {
+        _vertexarray = dynamic_cast<osg::Vec3Array*>(geometry.getVertexArray());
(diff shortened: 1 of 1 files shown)


src/osgUtil/MeshOptimizers.cpp
```diff
@@ -98,40 +100,62 @@ struct GeometryArrayGatherer
...
     {
     }
 
+    ///sort by distance to origin 
     bool operator() (unsigned int lhs, unsigned int rhs) const
     {
-        for(ArrayList::const_iterator itr=_arrayList.begin();
-            itr!=_arrayList.end();
+        for(ArrayList::const_reverse_iterator itr=_arrayList.rbegin();
+            itr!=_arrayList.rend();
```
openscenegraph: What has a reverse iterator been introduced here?

src/osgUtil/MeshOptimizers.cpp
```diff
@@ -98,40 +100,62 @@ struct GeometryArrayGatherer
...
             ++itr)
         {
             int compare = (*itr)->compare(lhs,rhs);
             if (compare==-1) return true;
             if (compare==1) return false;
         }
+        if (_vertexarray.valid())
+        {
+            float dpos1 = _vertexarray->at(lhs).length2();
+            float dpos2 = _vertexarray->at(rhs).length2();
```
openscenegraph: Computing the length2() of each vertex for each comparison is expensive.  The efficient way if you really wanted to do something like this would be compute the control distance in one pass and store in a single cache vector.

src/osgUtil/MeshOptimizers.cpp
```diff
@@ -98,40 +100,62 @@ struct GeometryArrayGatherer
...
         return false;
     }
 
+    ///compare returning equals if vertices distance is low enough
     int compare(unsigned int lhs, unsigned int rhs)
     {
-        for(ArrayList::iterator itr=_arrayList.begin();
-            itr!=_arrayList.end();
+        for(ArrayList::reverse_iterator itr=_arrayList.rbegin();
+            itr!=_arrayList.rend();
```
openscenegraph: Again, what is the reverse iterator for?

openscenegraph:
I have spent some time reflecting on the approach of using an epsilon around vertices to force non identical vertices to be matched and feel it's the wrong approach with potential pitfalls that could make it break meshes when vertices. 

The problem is that if you can start collapsing near vertices together then the order could make a difference and when you do collapse vertices you collapse triangles as well leading to degenerate triangles. The ordering difference could lead to three vertices that are in all equally close to each other in their original positions, once you start collapsing the first two the third could end up further away and no longer collapse as it should.  If the vertices were in a different ordering originally then you'll end up with a different mesh.

The original scheme of matching identical vertices doesn't have any of these issues.

The issue you are trying to address with this PR is a different one than optimizing meshing, it fixing buggy/noisy source data, for this the you really want a different technique that handles how best to collapse adjacent vertices and then deal with the degenerate triangles.  One way to collapse vertices would be to quantize them and the post process for degenerate triangles.  There will be other techniques as well.  

None belong in a mesh optimizer designed for optimizing the meshing order/vertex order for best performance without altering the data representation.

PR #866, opened 2019-12-12
Fontconfig should be external library.

dedowsdi:
Add Fontconfig to TARGET_LIBRARIES cause osg3::osgText target looking for
openscegraph-Fontconfig-import-targets.cmake, which doesn't exists.

### DIFF
--- a/src/osgText/CMakeLists.txt
+++ b/src/osgText/CMakeLists.txt
@@ -54,7 +54,7 @@ SET(COMPONENT_PKGCONFIG_DESCRIPTION "Text rendering library for OpenSceneGraph")
 
 if(OSG_TEXT_USE_FONTCONFIG AND Fontconfig_FOUND)
     MESSAGE(STATUS "osgText will be linked with FontConfig library")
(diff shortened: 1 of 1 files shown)


src/osgText/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ SET(COMPONENT_PKGCONFIG_DESCRIPTION "Text rendering library for OpenSceneGraph")
 
 if(OSG_TEXT_USE_FONTCONFIG AND Fontconfig_FOUND)
     MESSAGE(STATUS "osgText will be linked with FontConfig library")
-    list(APPEND TARGET_LIBRARIES Fontconfig::Fontconfig)
+    set(TARGET_EXTERNAL_LIBRARIES Fontconfig::Fontconfig)
```
openscenegraph: Why not:

     list(APPEND TARGET_EXTERNAL_LIBRARIES Fontconfig::Fontconfig)

dedowsdi: I thought it's empty, so I use set.
`list(APPEND ...)` is indeed better for future changes.

dedowsdi:
Updated.

PR #869, opened 2019-12-13
Added FBO GL extensions (useful for mobile VR etc.)

valid-ptr:
(1) glFramebufferTexture2DMultisample
(2) glInvalidateFramebuffer

### DIFF
--- a/include/osg/GLExtensions
+++ b/include/osg/GLExtensions
@@ -657,6 +657,7 @@ class OSG_EXPORT GLExtensions : public osg::Referenced
         bool isFrameBufferObjectSupported;
         bool isPackedDepthStencilSupported;
         bool isMultisampledRenderToTextureSupported;
+        bool isInvalidateFramebufferSupported;
 
         bool isRenderbufferMultisampleSupported() const { return glRenderbufferStorageMultisample != 0; }
         bool isRenderbufferMultisampleCoverageSupported() const { return glRenderbufferStorageMultisampleCoverageNV != 0; }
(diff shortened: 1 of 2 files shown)


openscenegraph:
If we are to merge this change with 3.6 then we'll need to update the SO version as the change affects the ABI.

This change might be best just merged with master.

valid-ptr:
> If we are to merge this change with 3.6 then we'll need to update the SO version as the change affects the ABI.

Yes, but it does not change API, just improve it ;)

You decide :)

openscenegraph:
On Fri, 13 Dec 2019 at 17:13, Konstantin S. Matveyev <
notifications@github.com> wrote:

> If we are to merge this change with 3.6 then we'll need to update the SO
> version as the change affects the ABI.
>
> Yes, but it does not change API, just improve it ;)
>
> You decide :)
>
The SO version is tied to the ABI, which will change, so it has to change.

If we are changing the SO version between 3.6.4 and 3.6.4 anyway then it'll
be OK to merge.  I haven't double checked this yet.

valid-ptr:
**3.6.4** release => so version is **160**
**3.6.5** dev => so version is **161**

openscenegraph:
On Mon, 16 Dec 2019 at 09:29, Konstantin S. Matveyev <
notifications@github.com> wrote:

> *3.6.4* release => so version is *160*
> *3.6.5* dev => so version is *161*
>
I looked it up and I updated the SO version number to 141 with fixes to
osgText. commit : 149c04b0df37e605a94495363801eb518b58a059

 So time to merge :-)

PR #856, opened 2019-11-22
Fix clang 8 & libc++ build errors

elsid:
Replace operators for implicit type conversion by explicit data() method to access implementation pointer and subscript operator to access element by index just like in std::vector.

```
src/osgPlugins/tga/ReaderWriterTGA.cpp:455:22: error: use of overloaded operator '==' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'long')
        if (colormap == NULL)
            ~~~~~~~~ ^  ~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:525:16: error: use of overloaded operator '==' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'long')
    if (buffer == NULL || linebuf == NULL)
        ~~~~~~ ^  ~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:525:35: error: use of overloaded operator '==' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'long')
    if (buffer == NULL || linebuf == NULL)
                          ~~~~~~~ ^  ~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:548:30: error: use of overloaded operator '==' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'long')
            if (formattedMap == NULL)
                ~~~~~~~~~~~~ ^  ~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:574:40: error: use of overloaded operator '[]' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'int')
                        index = linebuf[x];
                                ~~~~~~~^~
src/osgPlugins/tga/ReaderWriterTGA.cpp:577:50: error: use of overloaded operator '+' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'int')
                        index = getInt16(linebuf + x * 2);
                                         ~~~~~~~ ^ ~~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:580:50: error: use of overloaded operator '+' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'int')
                        index = getInt24(linebuf + x * 3);
                                         ~~~~~~~ ^ ~~~~~
src/osgPlugins/tga/ReaderWriterTGA.cpp:583:50: error: use of overloaded operator '+' is ambiguous (with operand types 'SafeArray<unsigned char>' and 'int')
                        index = getInt32(linebuf + x * 4);
[...]

PR #871, opened 2019-12-16
fix debug compile error for ReaderWriterTGA

LaurensVoerman:
Hi Robert,
I get a compile error in the debug config in Visual Studio Community 15.9.18 & 16.4.1:
OpenSceneGraph\src\osgPlugins\tga\ReaderWriterTGA.cpp(725): error C2676: binary '+': 'SafeArray<unsigned char>' does not define this operator or a conversion to a type acceptable to the predefined operator
This was introduced by "Fix clang 8 & libc++ build errors" from "elsid" (22/11/2019).
This is a PR for OpenScenGraph-3.6, I will make a PR for master too.
Regards, Laurens.


remoe:
I think this could fix this https://github.com/openscenegraph/OpenSceneGraph/pull/856#issuecomment-564624070

PR #875, opened 2019-12-20
fallback for DCMTK_ROOT_INCLUDE_DIR 

LaurensVoerman:
with the removal of FindOurDCMTK.cmake (2017/7/26) the variable DCMTK_ROOT_INCLUDE_DIR is not set by cmake anymore.
I did not notice this before because I set this variable on the cmake command line in my build script, but while building osg with vcpkg I noticed that the dicom plugin failed to build.
For the master tree I suggest to remove the variable entirely, for 3.6.5 I would like to suggest this conservative approach, with a fallback in case the variable is not defined.
Regards, Laurens.

### DIFF
--- a/src/osgPlugins/dicom/CMakeLists.txt
+++ b/src/osgPlugins/dicom/CMakeLists.txt
@@ -1,7 +1,22 @@
 IF  (DCMTK_FOUND)
 
-    # note, we have to include a '/' in front of the directory string to prevent a CMake bug from ignoring the directory
-    INCLUDE_DIRECTORIES(${DCMTK_ROOT_INCLUDE_DIR})
+    IF (DCMTK_ROOT_INCLUDE_DIR)
+        # osg used to have a FindOurDCMTK.cmake until 2017/7/26 (3.5.6); it defined this DCMTK_ROOT_INCLUDE_DIR
+        # keep using this if it's provided on command line
+        
+        # note, we have to include a '/' in front of the directory string to prevent a CMake bug from ignoring the directory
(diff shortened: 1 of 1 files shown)


openscenegraph:
I can't see a reason to not just use the CMakeModules/FndDCMTK.cmake definition of DCMTK_INCLUDE_DIRS instead of DCMTK_ROOT_INCLUDE_DIR.

Have you tried this?

openscenegraph:
I have just built with the following:

diff --git a/src/osgPlugins/dicom/CMakeLists.txt b/src/osgPlugins/dicom/CMakeLists.txt
index 37f621a79..3f159a9c5 100644
--- a/src/osgPlugins/dicom/CMakeLists.txt
+++ b/src/osgPlugins/dicom/CMakeLists.txt
@@ -1,11 +1,11 @@
 IF  (DCMTK_FOUND)
 
     # note, we have to include a '/' in front of the directory string to prevent a CMake bug from ignoring the directory
-    INCLUDE_DIRECTORIES(${DCMTK_ROOT_INCLUDE_DIR})
+    INCLUDE_DIRECTORIES(${DCMTK_INCLUDE_DIRS})
 
     SET(TARGET_SRC ReaderWriterDICOM.cpp )
 
-    LINK_LIBRARIES(${DCMTK_LIBRARIES} ${ZLIB_LIBRARIES})
+    LINK_LIBRARIES(${DCMTK_LIBRARIES})
 
     ADD_DEFINITIONS(-DUSE_DCMTK)

openscenegraph:
I've merged the above change with master and the 3.6 branch.

PR #880, opened 2019-12-26
BufferStorage and Persistent DMA

mp3butcher:
as it's based on 3.6 branch it may be merged with 3.6 branch for completeness with already present ImmutableTextureBufferStorage feature


openscenegraph:
Do you have any code that illustrates the usage case?

mp3butcher:
done

> Do you have any code that illustrates the usage case?

done

mp3butcher:
some flags are excllusive to glBufferStorage and glMapBufferRange so i had to add a new bitfield

openscenegraph:
The changes look sensible now.  Could you add the required serializers with handling of the versioning so we can retain backwards compatibility.

mp3butcher:
mmh there's a conflict on CMakeList (i think it's because i'm based on 3.6 branch)....
:(
I used a hexint serialiser for _mappingbitfield to try to maximize osgt readability but perhaps a simple Glenumserializer could be best fitted...

examples/osgpersistentbufferstorage/osgpersistentbufferstorage.cpp
```diff
@@ -0,0 +1,134 @@
...
+    const float _scale;
+    const float _offset;
+};
+
+
+///////////////////////////////////////////////////////////////////////////
+
+int main( int, char** )
+{
+    osg::Camera* root=new osg::Camera;
```
openscenegraph: I'm trying to make sense of this example.  You have a in scene graph Camera that has a post draw callback that is called after the Camera's subraph is rendered, which is a bit odd, as one would normally update a buffer before you use it, not after.  

And why a Camera?  Cameras have very specific roles in the OSG, they aren't general purpose means of adding draw callbacks like you appear to be using them for.  Why isn't a geometry draw callback being used?

mp3butcher: I used a Camera::postdraw cause i had in mind to test readback feature ( too lazy to had a GPU buffer write pattern)
If it'd make you happier to convert it to a Drawable::drawcb, i'll comply

openscenegraph:
The draw callback changes to the example make things more relatable thanks.  I'm going to merge with master but not the 3.6 branch as we'd end up with a tangled web of serializer file formats otherwise, and 3.6 branch is for maintenance rather than new features.

PR #882, opened 2019-12-28
[*] Text/Text3D: VAO initialization fixed in drawImplementation: VBO …

valid-ptr:
…and EBO must be unbinded


openscenegraph:
I have implemented a TextBase::compileGLObjects() that resolves the bug for me without the need to modify the Text:drawImplementation().  This approach should avoid any unnecessary bind/unbinds happening at rendering-time.

https://github.com/openscenegraph/OpenSceneGraph/commit/605821e655d97b7ad5ab6098b3ec26e2f9ad2ff7

PR #883, opened 2020-01-03
[DXF Plugin]Fix dxf text chinese coding problem

longhuan2018:
When dxfText contains Chinese, it is displayed as garbled characters in windows. And does not support line width styles. 
I tried to modify the code to support these, here is the rendering.
before
after
Test Data:
[dxf-test.zip](https://github.com/openscenegraph/OpenSceneGraph/files/4018110/dxf-test.zip)


openscenegraph:
I have done a first pass review and as it stands it isn't ready to merge.

Could you split the submissions into font changes and the line width changes.

Could you change the indentation of your changes so it fits with the rest of the OSG i.e. 4 space indentation.  This is done to keep the code looking the same across platforms and different editors.

The addition of the GDAL dependency is odd, GDAL is a geospatial library, yet you seem to be using in some way.  If there is something missing from the OSG features then please highlight this rather than drag in dependencies where they don't belong.

longhuan2018:
ok, i will split the submissions into font changes and the line width changes.

PR #884, opened 2020-01-03
Fix write image file from relative to absolute

longhuan2018:
When using osgConv for format conversion, I found that the output path of the texture file is related to the current path of cmd instead of outputting to the destination file path. If the texture file name contains a directory, if the corresponding directory does not exist in the target directory, the texture file will fail to output.
image


mp3butcher:
I think you'd better describing your issue you have before going wild with coding (indentation issues of you pr is a clue...)

i dont have enough insight on how osg manage filesystem but it is cross plateform: so filesystem management conventions have to work both under win32 and linux filesystem.
have you tried your changes under linux?

Further, 17 files mod is quite a huge change to review, can you explain a bit how you proceed to fix the issue.

longhuan2018:
> I think you'd better describing your issue you had before going wild with coding (indentation issue is a clue)
> 
> i can't remember enough insight on how osg manage filesystem but it is cross plateform: so filesystem management conventions have to work both under win32 and linux filesystem.
> have you tried your changes under linux?
> 
> Also, i think relative filesystem conventions in osg was about making outputed file portable acrosse multiple plateform

ok, i will create an issue

openscenegraph:
I have done a first pass review and as is the PR is not appropriate for merging.

The indentation is out of step with the rest of the OSG.  Please use four spaces for indentation.

The old .osg and .ive file formats are deprecated, they've been deprecated since OpenSceneGraph-3.0 released back in June 2011. Modifying how these old formats behave is breaking with preexisting behavior in something only exist to retain that old behavior.

For modern application I strongly recommend using the modern serialized formats i.e. .osgt, .osgb and .osgx.  If you want to propose changes in behavior to these then you are welcome, it's best done as an option to enable it.  Such changes should be a separate PR.

Changes to the OBJ plugin again deserve a dedicated PR and new option to enable the new behavior.

PR #885, opened 2020-01-04
[Controversal:/] Changing stateattribute priority

mp3butcher:
There's a severe under perf in the priority of the Program, i reorganized that based on AZDO guide lines
Quoting Mathias Schott from nvidia
> Sort by renderpass (framebuffer, blending, depth/stencil..)
> then sort by pipeline (shader,tesseleation..)
> then sort by material (texture,uniforms...)
> then sort by object (buffers, matrices....)

would require feedback on my compliance with this

### DIFF
--- a/include/osg/StateAttribute
+++ b/include/osg/StateAttribute
@@ -122,19 +122,38 @@ class OSG_EXPORT StateAttribute : public Object
           * that type's value. */
         enum Type
         {
+/// Renderpass attributes
+            FRAME_BUFFER_OBJECT,
+
+/// Shader attributes
+            VERTEXPROGRAM,
+            FRAGMENTPROGRAM,
+            PROGRAM,
+
+/// Material attributes
             TEXTURE,
+            SAMPLER,
+            BINDIMAGETEXTURE,
+            MATERIAL,
+            LIGHTMODEL,
+            LIGHT,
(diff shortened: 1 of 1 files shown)


openscenegraph:
I understand the motivation.  However, I don't know if it'll make any difference to performance without sorting on contents of StateSet rather than the just the StateSet pointers as is done right now.  Sorting on contents of StateSet is slower on the CPU so could introduce it's own bottlenecks.

Have you done any performance testing on this change? What tests were they?

mp3butcher:
The ability to sort correctly statesets even if it is not used in the renderLeafs sorting is not a bad feature...It can be used in a scenegraph preprocess to reorder it for example...

openscenegraph:
For this type of sorting I think it may be better to introduce a concept of StateAttribute Category that all the same level for sorting have the same value.  What the best name for it I haven't thought about.  Using a int or uint would be fine, or even an enum.  It's value could be accessed from a virtual function so no extra data need be stored.

openscenegraph:
I am going to close this PR as I feels it's the wrong approach to the problem that is intended to be solved.  I think the right way is to have a dedicated sort value for classifying the type of state, rather than having this implicit in the ordering of an enum.

PR #888, opened 2020-01-07
Dxf support linewidth

longhuan2018:
Enable dxf plugin to support line width properties


openscenegraph:
Thanks for the splitting the PR and fixing the indentation.

I've done a code review and see opportunities for making the resulting scene graphs more efficient.  First up the use of smoothing mode and Hint should be applied as an option at the top level of the scene graph, not applied to the Geode leaves as this will result in lots of duplicate settings.

Secondly the StateSet + LineWidth combination will be assigned independently to each Geode, again this will result into lots of potential duplicate state.  The way to address this would be to have a map of line width to StateSet or LineWidth objects and enable reuse where the same line width is used across several geometries.

With these changes you'll achieve better performance.

longhuan2018:
I modified the code based on your comments. Regarding smoothness, I think it is better to let the person who calls it set it up externally. So I removed it from the code.

> Thanks for the splitting the PR and fixing the indentation.
> 
> I've done a code review and see opportunities for making the resulting scene graphs more efficient. First up the use of smoothing mode and Hint should be applied as an option at the top level of the scene graph, not applied to the Geode leaves as this will result in lots of duplicate settings.
> 
> Secondly the StateSet + LineWidth combination will be assigned independently to each Geode, again this will result into lots of potential duplicate state. The way to address this would be to have a map of line width to StateSet or LineWidth objects and enable reuse where the same line width is used across several geometries.
> 
> With these changes you'll achieve better performance.

PR #891, opened 2020-01-10
Dxf support chinese

longhuan2018:
The reason for Chinese garbled characters is that when osgText :: setText, char is passed in instead of wchar. Another reason is that the set font does not support Chinese. So if it can be controlled by options, then the user can decide when to use it, which will be more flexible.
Test data and command line:
[dxf-test.zip](https://github.com/openscenegraph/OpenSceneGraph/files/4044403/dxf-test.zip)
```
osgViewer -O "WideChar  FontFile=msyh.ttf" dxf-test.dxf
```


openscenegraph:
I have done a code review and left wondering whether we can avoid the need to the Options settings for enabling wchar and specifying the font.

osgText::String is a vector or unsigned int as well a various constructors for doing the UTF8 and other encodings to a osgText::String.  Have a look at include/osgText/String header.  It may be cleaner just to use these rather that text->setText(..).

On the font front, what does DXF format provide in terms of guidance on font?  If so would be possible to use this?

src/osgPlugins/dxf/dxfEntity.cpp
```diff
@@ -910,10 +911,19 @@ dxfText::drawScene(scene* sc)
     sc->ocs(m);
 
     ref_ptr<osgText::Text> _text = new osgText::Text;
-    _text->setText(_string);
+    if (_useWideChar && !_string.empty())
```
openscenegraph: Are DXF text strings always encding in UTF8? if so we can just use line

  _text->setText(osgText::String(_string, osgText::String::ENCODING_UTF8));

In place of the original _text->setText(_string) line.

openscenegraph:
Rather than try to iterate a solution based on my suggests I have take a first step to implement control of the encoding and font by adding these as variables to the dxfText entity:

   https://github.com/openscenegraph/OpenSceneGraph/commit/a4054572b9e1c149648d83817a51182ca5ef70db

This is checked into master and the 3.6 branch.  I have gone for this approach as the code is cleaner than equivalent in the PR. The next step would be to merge the option parsing that this PR has ontop of this commit.

openscenegraph:
I have been trying to get the dxf-test.dxf working on my Linux system and have tried UTF8, UTF16 and UTF32 and non render the Chinese characters as I'd expect.  The original PR uses the Windows CodePage functionality that isn't available on other platforms like Linux, it also strikes me as a assuming the DXF was created with the same ecoding as what is currently running on the system that saved the DXF, which doesn't seem generally safe.

I'm stuck with how best to resolve this.  I will check in what I have once I've done a code review.

openscenegraph:
I have added support for using CurrentCodePath into osgText and dxf plugin:

https://github.com/openscenegraph/OpenSceneGraph/commit/9e75926338fcb7eb4732acd54af4864b151b8d93

The use of Current Code Page is only functional under Windows so I can't test it myself.  The change is checked into the 3.6 branch and master. 

As the above commit wraps up similar functionality to this PR I'll now close to PR.  If the commits don't work as intended please create a Issue here on github,

PR #900, opened 2020-01-17
Fix particle rotation.

dedowsdi:
Particle rotation is reversed, a particle with `0 0 1` angular velocity rotate clockwise.

### DIFF
--- a/src/osgParticle/ParticleSystem.cpp
+++ b/src/osgParticle/ParticleSystem.cpp
@@ -343,16 +343,16 @@ void osgParticle::ParticleSystem::drawImplementation(osg::RenderInfo& renderInfo
 
                     if (_alignment==BILLBOARD)
                     {
-                        xAxis = osg::Matrix::transform3x3(R,scaled_aligned_xAxis);
+                        xAxis = osg::Matrix::transform3x3(scaled_aligned_xAxis, R);
                         xAxis = osg::Matrix::transform3x3(modelview,xAxis);
 
-                        yAxis = osg::Matrix::transform3x3(R,scaled_aligned_yAxis);
+                        yAxis = osg::Matrix::transform3x3(scaled_aligned_yAxis, R);
                         yAxis = osg::Matrix::transform3x3(modelview,yAxis);
                     }
                     else
                     {
-                        xAxis = osg::Matrix::transform3x3(R, scaled_aligned_xAxis);
-                        yAxis = osg::Matrix::transform3x3(R, scaled_aligned_yAxis);
+                        xAxis = osg::Matrix::transform3x3(scaled_aligned_xAxis, R);
+                        yAxis = osg::Matrix::transform3x3(scaled_aligned_yAxis, R);
                     }
                 }
 


openscenegraph:
Thanks for the PR.

Could you explain how to recreate the bug so I can test before and after results. Thanks.

dedowsdi:
Create a trivial particle system, change angular velocity range to `0,0,1 - 0,0,1`  should do:
```
#include <osgViewer/Viewer>
#include <osgParticle/ParticleSystemUpdater>
#include <osgParticle/ModularEmitter>

int main( int argc, char* argv[] )
{
    osg::Group* root = new osg::Group;

    osgParticle::ParticleSystemUpdater* psu = new osgParticle::ParticleSystemUpdater;
    root->addChild(psu);

    auto leaf = new osg::Geode;
    root->addChild(leaf);

    osgParticle::ParticleSystem* ps = new osgParticle::ParticleSystem;
    leaf->addDrawable( ps );
    psu->addParticleSystem( ps );

    osgParticle::ModularEmitter* emitter = new osgParticle::ModularEmitter;
    emitter->setParticleSystem( ps );
    root->addChild( emitter );

    osgParticle::RadialShooter* shooter =
        static_cast<osgParticle::RadialShooter*>( emitter->getShooter() );
    shooter->setInitialRotationalSpeedRange(osg::Z_AXIS, osg::Z_AXIS);
    shooter->setInitialSpeedRange(1, 1);

    osgParticle::RandomRateCounter* counter =
        static_cast<osgParticle::RandomRateCounter*>( emitter->getCounter() );
    counter->setRateRange( 1, 3 );

    osgViewer::Viewer viewer;
    viewer.setSceneData( root );
    return viewer.run();
}
```

openscenegraph:
Thanks for the example.  I'm confused by it and in particular your comment:

dedowsdi:
## About the example

The ParticleSystem in the example shoot all particles with 1 angular velocity along z axis.

`0,0,1 - 0,0,1` is a  range with the same start and end,  so all particles rotate along z axis in 1 angular velocity, it's purely for debug purpose.

`0,0,1` is the same as `osg::Z_AXIS`,  I have an abbreviation for that in `vim`,  I shouldn't use it, sorry for the sloppy and confusion.

## About the code:

The original author must think it's a bad idea to scale ParticleSystem with `modelview` matrix, he spent a  lot of code to calculate `xScale` and `yScale` to neutralize the scale in `modelview` matrix before he loop the particles. 

The particle rotation is applied here:
https://github.com/openscenegraph/OpenSceneGraph/blob/25afb086ef1012cb26cd4f3b2d1569ddb9d1739d/src/osgParticle/ParticleSystem.cpp#L336-L357

The `scaled_aligned_xAxis` is `1, 0, 0` multiplied by `xScale` to neutralize the scale int the `modelview` matrix.  If it's billboard,  `modelview` matrix is post multiplied by `xAxis` to get `model` space `xAxis`. Note that this is the first time the `scale` part of `modelview` matrix take effect, it will take effect again in the `pipeline`, it's all been taking care of by `xScale` and `yScale`.

But it doesn't make sense to post multiply `R` by `scaled_aligned_xAxis`, that will inverse  the rotation.

openscenegraph:
I have now had a chance to play around with your example and the suggested fix and agree that the original code is a bit of mess and is rotating the wrong way based on the local Z axis of the billboard pointing towards the eye point, and your fix looks correct and behaves consistent with a +ve rotation around the axis should be ant-clockwise.  

I am a bit hesitent to merge with 3.6 as existing users may have already compensated for this bug by negating the rotation.   However, I'm going to merge in the hope that visually the particle rotation direction won't be that critical in most applications so a change won't be an serious issue.

The wider code really looks like it could do with rewriting completely to be more efficient, however,, doing particle systems on the CPU is pretty daft these day so I really consider osgParticle really just there for backwards compatibility so don't see it as something we should invest time in trying to perfect and risk breaking when users really should be implementing particle systems on the GPU on modern hardware.

Issue #901, opened 2020-01-19
occlusion query  api variations across osg version


mp3butcher:
Just to mention that OcclusionQueryNode api have changed 3 times (3.4,3.6,master)
Perhaps recent changes made by @dan-t  to master should be merged with 3.6 ...

Capostrophic:
> Perhaps recent changes made by @dan-t to master should be merged with 3.6 ...

And break backward compatibility with existing 3.6.x applications? So much for bug fix releases.

openscenegraph:
Bug fixes and backwards of API is highly desirable but isn't always possible.

@mp3butcher & @dan-t, if there are changes that are in master and would be appropriate then please generate a PR for them. Thanks.

mp3butcher:
@dan-t @openscenegraph 
https://github.com/openscenegraph/OpenSceneGraph/pull/902

openscenegraph:
Merged, thanks.

PR #902, opened 2020-01-20
3.6 OQN API convergence 

mp3butcher:
following issue https://github.com/openscenegraph/OpenSceneGraph/issues/901
NB: master lacks cfd190a9faf886f3bf9b5e2c44ef0ceef2631202 and
5311d6fb6b1a1840322e12a52d58792193f81e61 if i'm right

### DIFF
--- a/include/osg/OcclusionQueryNode
+++ b/include/osg/OcclusionQueryNode
@@ -82,9 +82,9 @@ public:
     };
 
     /** return a QueryResult for specified Camera, where the QueryResult.valid is true when query results are available, and in which case the QueryResult.numPixels provides the num of pixels in the query result.*/
-    QueryResult getQueryResult( const osg::Camera* cam );
+    QueryResult getQueryResult( const osg::Camera* cam ) const;
 
-    unsigned int getNumPixels( const osg::Camera* cam );
+    unsigned int getNumPixels( const osg::Camera* cam ) const;
 
     virtual void releaseGLObjects( osg::State* state = 0 ) const;
 
@@ -158,8 +158,11 @@ public:
     osg::StateSet* getQueryStateSet();
     const osg::StateSet* getQueryStateSet() const;
 
-    // Get the QueryGeometry object used for occlusion query. Returns 0 if no QueryGeometry is created.
-    osg::QueryGeometry* getQueryGeometry();
+    // Set and get the QueryGeometry object used for the occlusion query.
+    // By default an axis aligned box is used as the query geometry.
+    // Resetting to the default query geometry is done by setting it to 0.
+    // Returns 0 if no QueryGeometry is created.
+    void setQueryGeometry( osg::QueryGeometry* geom );
     const osg::QueryGeometry* getQueryGeometry() const;
 
     // Set and get the StateSet used by the OcclusionQueryNode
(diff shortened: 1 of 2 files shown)


mp3butcher:
oups

Capostrophic:
Mr Osfield, it's not that I'm against making occlusion query node usage more sensible, but this significantly changes the API in a branch that is supposed to be in maintenance mode, and the latest stable release of OpenMW, potentially the most widely installed OSG application, will not be compatible with upstream 3.6.x at all anymore (even though it just barely is). However, Arch and Debian maintainers insist on getting rid of 3.4.1 OSG from their repositories, and OpenMW may potentially get cut off from a significant part of its audience if we don't get things together.

If there's a definite plan to break the backward compatibility, at least give us some grace period, please, and postpone this improvement to some time *after* the RC phase 3.6.5 is currently in the beginning of.

openscenegraph:
@Capostrophic, for the stable releases series my plan is never to break the API, but if bugs that users are hitting up against require API changes because the design/interface is broken then I'll consider these.  For all releases I repeated ask for feedback from the community.  

The OSG has hundreds of classes and plugins, over half a million lines of code, I have no knowledge of exactly what feature which end users utilitise, I can't call out ever possible revisions that some end user might have an issue with.  If there is a class that I know is widely used then I'll do a call on osg-users ML/forum, but there is simply too many minor tweaks to do this for everyone.  The support work on the OSG I do mostly for free, the effort I put on it I carve out of my work for paying clients and my spare time.  There is limit on just many pounds of flesh I personally donate month in month out.

I have to rely upon the community to pull some of the weight as well.  This is often in form of testing and bug fixes.  Sometimes we see regressions or changes that might push and pull between different user groups.  The only way we can catch all of this is for users to remain engaged in the project, checking the head of branches or master to make sure there own projects still work.

In the run up to release I time after time beg the user community to test.  If user groups don't respond to this and don't wave a flag when there is problem then there is *nothing* I can do for them.

The OSG is open source, you can build it yourself, if relying upon 3rd party libs such as from repositories is insufficient then statically build and link the OSG to your application.  If you want to rely upon libs from repositories then you as a project have to accept that you'll need to keep track of changes.  The include/osg/Version header contains macros to help your specialize code blocks to make it possible to work across different versions.

Finally, don't slurge stuff on PR's where it doesn't belong.

openscenegraph:
@mp3butcher I just did a diff between master and the 3.6 (with your PR merged) and the only difference I get is:

diff --git a/src/osg/OcclusionQueryNode.cpp b/src/osg/OcclusionQueryNode.cpp
index 02509788d..6be23a5f5 100644
--- a/src/osg/OcclusionQueryNode.cpp
+++ b/src/osg/OcclusionQueryNode.cpp
@@ -586,7 +586,7 @@ bool OcclusionQueryNode::getPassed( const Camera* camera, NodeVisitor& nv )
 
 void OcclusionQueryNode::traverseQuery( const Camera* camera, NodeVisitor& nv )
 {
-    if (!isQueryGeometryValid() || !_enabled)
+    if (!isQueryGeometryValid() || ! _enabled)

Which boils down to a stray space being in the 3.6 version, once I remove this space the diff goes away.  I presume the various commits in 3.6 made it into master and visa-versa but with some being done manually.

PR #904, opened 2020-01-21
fix: vendor ext prevent OQN to work on other HW

mp3butcher:
prevent Occlusion Qeury to work on other than NVIDIA
NB: on master too
As a general statement, we could replace all NV specific with their ARB.....most are standardized till a long time now


mp3butcher:
Noise sorry

openscenegraph:
As a general comment, the correct way to add support for other extensions to to just add the extra string to the get extensions. See other examples of it.

PR #906, opened 2020-01-25
[Discussion required] PLY plugin

mp3butcher:
I did mod on ply plugin in order to make Equalizer application specifics overridable
the idea is to downcast Reader in the code and customize VertexSemantics (perhaps via options too)
not backward compatible 
- color scaling removed, normal generation removed

what's new added:
- more generic support 
- Element Tree Is setted as UserData Tree (PLY(Node)->ELMENT(Node)->PROP(DrawElments or Arrays name after first PLY composite) so user can hack its own semantic out of it.

NB:
the pr is against 36 branch so not for merging purpose (or perhaps) but for community validation

complex header example that works (**_format keyword highlighted_**), test with your own and parse userdata

**ply**
**format binary_little_endian** 1.0
**comment** https://www.artec3d.com/3d-models/fox-skull
**element** vertex 349950
**property float** x
**property float** y
**property float** z
**element** face 699976
**property list uchar int** vertex_indices
**element** multi_texture_vertex 365208
**property uchar** tx
**property float** texture_u
**property float** texture_v
**element** multi_texture_face 699976
**property uchar** tx
**property uint** tn
**property list uchar int** texture_vertex_indices
**end_header**

Sample interpretation:
"multi_texture_face" is per "face" data (tx tn and texture_vertex_indices are per faces data)
"multi_texture_face" could be seen as redondant and could have been merged with "face" like that

**element** face 699976
**property list uchar int** vertex_indices
**property uchar** tx
**property uint** tn
**property list uchar int** texture_vertex_indices

However "texture_vertex_indices" don't refer to "vertex" but to "multi_texture_vertex" so separate 2 element have sense...

### DIFF
--- a/src/osgPlugins/ply/ReaderWriterPLY.cpp
+++ b/src/osgPlugins/ply/ReaderWriterPLY.cpp
@@ -60,6 +60,8 @@ public:
         _semantic.push_back({ "alpha", PLY_FLOAT,  2});
         _semantic.push_back({ "u", PLY_FLOAT, 3});
         _semantic.push_back({ "v", PLY_FLOAT,  3});
+        _semantic.push_back({ "texture_u", PLY_FLOAT, 3});
+        _semantic.push_back({ "texture_v", PLY_FLOAT,  3});
         _semantic.push_back({ "ambient_red", PLY_UCHAR, 4});
         _semantic.push_back({ "ambient_green", PLY_UCHAR, 4});
         _semantic.push_back({ "ambient_blue", PLY_UCHAR, 4});
--- a/src/osgPlugins/ply/ReaderWriterPLY.cpp
+++ b/src/osgPlugins/ply/ReaderWriterPLY.cpp
@@ -37,7 +37,6 @@ using namespace std;
 //! \brief This is the Reader for the ply file format
 //!
 //////////////////////////////////////////////////////////////////////////////
-///
 
 class ReaderWriterPLY : public osgDB::ReaderWriter
 {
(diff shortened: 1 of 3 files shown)


openscenegraph:
I have looked at the changes and find them curious, it seems like they are a stepping stone.  What are the plans next?

With the discussion on osg-users with Tom it strike me if we have issues with the how to interpret the ply format being application specific it makes me wonder if something like using a lua script to guide how to map the ply data to a scene graph might be appropriate.  The basic data in the ply file could be mapped to an osg::Object hierarchy that represents the data in the file, then have a lua script parse this and convert to a usable scene graph.  We could have different lua scripts that handle particular art path routes.

mp3butcher:
mapping isnot the only user parameter,  there's filtering and type conversion to
for the moment i only use vertexattribute index and exploit default compact aliasing to map as it was.
Will dig the lua idea

tompollok:
Maybe it would be helpful to define some headers that are/will be supported by osg by default and document them somewhere? Either in the osg group or as an issue or preferably in the code or readme file in the ply plugin?

Also here is the ply importer from MeshLab (http://www.meshlab.net/). They encapsulated it into vcglib.

https://github.com/cnr-isti-vclab/vcglib/blob/master/wrap/io_trimesh/import_ply.h

Maybe they do something differently that could be interesting.

It seems a more complex implementation that contains elements like vertices, faces, cameras and edges. I dont know if there is a scene ply exporter in osg, but that could be potentially nice for exporting scenes.

mp3butcher:
> It seems a more complex implementation that contains elements like vertices, faces, cameras and edges. I dont know if there is a scene ply exporter in osg, but that could be potentially nice for exporting scenes.

I think you don't understand: There no semantic in ply (in the header i copy paster, all not bold can't be given a meaning.
If you review my change in ReaderWriter, you'll see that i give meaning to not bold keyword as a parameter
ex 
        _semantic.push_back({ "x", PLY_FLOAT, 0 });

here i tell that x property will be map on VertexAttributeArray no 0

openscenegraph:
@mp3butcher I have created an osgply branch, could you generate a PR for that new branch rather than the 3.6 that it's currently PR'd relative.  This way various experiments can be merged and tested more easily by others.  If the work ends up suitable for merging I can merged with with either master or 3.6.

My own focus for now is 3.6.5 so just bug + build fixes.

openscenegraph:
Thanks @mp3butcher now merged as the new osgply branch

    https://github.com/openscenegraph/OpenSceneGraph/tree/osgply

@tompollok is you want to experiment then this new branch is the place to work with to see if this new approach can be made to work with your data.

PR #907, opened 2020-01-26
FBX plugin updates / PREFIX-NOTFOUND work-around

DeadParrot:
This updates the Autodesk FBX plugin support for newer FBX and MSVC versions and for the xml2 and zlib libraries needed on Windows. It also fixes up some CMake usage in FindFBX.cmake. Updates for the non-current versions are non-tested best guesses. These changes could also go into the master branch.

There is also a little patch for ModuleInstall.cmake to remove PREFIX-NOTFOUND prefixes that can break Windows builds. There could be build scenarios where leaving that as the prefix is needed that I am unaware of. The master ModuleInstall.cmake has many updates since the 3.6 branch but this patch could go in there as well.


openscenegraph:
Thanks Stuart, merged with 3.6. I also merged with master but there was a collision so I had to tweak it a bit.  Could you check to make sure I haven't introduced any issues:

https://github.com/openscenegraph/OpenSceneGraph/commit/8643c0001fb54ec7a852eae26a58abc7e69b5ece

DeadParrot:
The merge looks fine to me, Robert.

PR #911, opened 2020-01-30
Fix for older versions of fbxsdk without xml or zlib libraries,

LaurensVoerman:
Fix for older versions of fbxsdk without xml or zlib libraries,
fix cmake multiconfig generators (msvc) with irrelevant CMAKE_BUILD_TYPE.

Hi Robert,
I would like to suggest some changes to the fbx CMakeLists:
-older fbx sdk versions haven no xml or zlib libraries, current version fails to configure
-cmake multiconfig generators (msvc) needs to specify both optimized and debug libraries
-link to FBX_ZLIB_LIBRARY if it's found in all configurations (debug/release, static/dll)
Laurens.


openscenegraph:
Thanks @LaurensVoerman I've merged this change with master as well.

PR #919, opened 2020-03-02
TextureAttribute's _textureUnit not set on copy

LaurensVoerman:
while I cannot find any impact, I think the TextureAttribute's _textureUnit should be set in a copy operation.
While simply (deep) copying a stateset the TextureAttribute gets in the right list, while the _textureUnit  is still not set - so my simple test with read - clone - write failed to reveal anything.
The undefined value seems to be returned by "attribute->getTypeMemberPair()" used in State::applyAttribute, so I think rendering the cloned copy might cause trouble, but I did not test.
This only applies to master; 3.6 doesn't have a TextureAttribute class.
Regards, Laurens.

### DIFF
--- a/include/osg/TextureAttribute
+++ b/include/osg/TextureAttribute
@@ -28,7 +28,7 @@ class TextureAttribute : public StateAttribute
             _textureUnit(0) {}
 
         TextureAttribute(const TextureAttribute& ta, const CopyOp& copyop=CopyOp::SHALLOW_COPY):
-            StateAttribute(ta, copyop) {}
+            StateAttribute(ta, copyop), _textureUnit(ta._textureUnit) {}
 
         virtual bool isTextureAttribute() const { return true; }
 


openscenegraph:
The TextureAttribute::_textureUnit is not a public set property, but a means to track the unit when assigning to a StateSet.  It's not something that can be copied reliably as it's value is only valid once it's assigned to a StateSet and it's location stored.  Copying the _textureUnit would be mis-leading, but not initilizing it at all is also mis-leading.  I guess ideally one would have an un-assigned state for the _textureUnit.

openscenegraph:
I have an initializer for _textureUnit:

 https://github.com/openscenegraph/OpenSceneGraph/commit/3590e4c0574f33144acda03e162a80147fd3a64a

It's not perfect, I think we probably need an uninitialized state for this variable.  It might be that embedded the texture unit this way would need revisited.

PR #925, opened 2020-03-13
png and tiff plugins do not depend on C++ exceptions.

philip-lamb:


src/osgPlugins/rot/ReaderWriterROT.cpp
```diff
@@ -80,7 +80,7 @@ static bool getFilenameAndParams(const std::string& input, std::string& filename
  * This pseudo-loader makes it simple to change the orientation of a saved
  * model by specifying a correcting rotation as part of the filename.
  *
- * Usage: <modelfile.ext>.<rx>,<ry>,<rz>.globe
```
robertosfield: Why this change and why is part of this PR?

src/osgPlugins/trans/ReaderWriterTRANS.cpp
```diff
@@ -81,7 +81,7 @@ static bool getFilenameAndParams(const std::string& input, std::string& filename
  * This pseudo-loader make it simple to change the origin of a saved model
  * by specifying a correcting translation as part of the filename.
  *
- * Usage: <modelfile.ext>.<tx>,<ty>,<tz>.globe
```
robertosfield: Again, what's going on here?

robertosfield:
Could you explain the specific issue you've seen and how to reproduce it, cutting to the chase to "a solution" doesn't help me work out the most appropriate solution for the problem. Thanks.

src/osgPlugins/trans/ReaderWriterTRANS.cpp
philip-lamb: This is correcting what looks like a copy and paste error in the comment describing the usage of the TRANS plugin.

src/osgPlugins/rot/ReaderWriterROT.cpp
philip-lamb: This is correcting what looks like a copy and paste error in the comment describing the usage of the ROT plugin.

openscenegraph:
Thanks for the explanation this helped make sense of the changes.  I've had a look at the png plugin and it does have some throw/catch code but it's safely protected so it's indeed possible to build it without exceptions.  tiff plugin too doesn't have any throw/catch code so will be safe as well.  Comment changes look appropriate as well.

PR #928, opened 2020-03-23
Update LAS plugin to support version 1.4

scttgs0:


openscenegraph:
Thanks for the new plugin version.  I see that you've bundled lots of cmake changes unrelated the LAS plugin changes, could you generate a separate PR for each group of changes so I can review and merge each set separately, and any possible issues we may need to revisit with one batch needn't affect the others.

openscenegraph:
One of the questions that my initial code reviewer popped up was the adoption of find_package() instead of INCLUDE(FindPkgConfig OPTIONAL).  I prefer the find_package() usage, but I don't yet know if this will break the build for users with older versions of CMake.  

I have had a look at the cmake online docs but haven't found an answer to what the minimum cmake version is for it.  The OSG currently requires a minimum of 2.8.0.

PR #930, opened 2020-03-23
Improve PkgConfig usage

scttgs0:
Including FindPkgConfig directly into scripts is not a recommended practice.
Doing so will cause CMake to complain with a warning.
Replaced with find_package(PkgConfig QUIET)

### DIFF
--- a/CMakeModules/FindGTA.cmake
+++ b/CMakeModules/FindGTA.cmake
@@ -8,14 +8,10 @@
 # correspond to the ./configure --prefix=$GTA_DIR
 # used in building libgta.
 
-INCLUDE(FindPkgConfig OPTIONAL)
+find_package(PkgConfig QUIET)
 
 IF(PKG_CONFIG_FOUND)
-
-    INCLUDE(FindPkgConfig)
-
     PKG_CHECK_MODULES(GTA gta)
-
 ENDIF(PKG_CONFIG_FOUND)
 
 IF(NOT GTA_FOUND)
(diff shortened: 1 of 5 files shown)


robertosfield:
Do you know if CMake 2.8.0 supports find_package()?  If so then this PR should be safe, if not we'll need to make a decision on whether to up the CMake requirements of the OSG, or provide some alternative approach that can work with CMake 2.8.0.  

FYI, I want to avoid adding extra build constraints on users unless it's really required.

scttgs0:
It does.
[find_package](https://cmake.org/cmake/help/v2.8.0/cmake.html#command:find_package)

CMake 2.8.0 is very old.  I recommend upgrading to 3.0 as a minimum.  Considering that the purpose of CMake is to make building open source software easy, having such an old version of CMake isn't really helpful for the users.  Unfortunately, many just use the version within their distro package manager without realizing how easy it is to keep CMake current.  i.e. building CMake from source is very simple.  Downloading the latest package from their website is even easier.

robertosfield:
Thanks for looking it up, I did a quick search on find_package and didn't find any mention of when it was introduced.

As for forcing update to new version.  For OSG master I'm more open to it. For OSG-3.6 I'd rather avoid breaking anything and any new change can potentially break the build, but as people don't all respond to my calls for testing it can be many months or even years before you find out you broke the build for people.

W.r.t. using different tool versions, some groups have locked down developer environments are have strict rules on what tools they can use and what versions. It's a nightmare for them just to keep their systems functioning.  It's crazy for them but that's the constraints they live within.

PR #933, opened 2020-03-24
Add RenderLeafCallback to give consumers an oppertunity to manage glo…

EricStancliff:
…bal uniforms at draw time.

### DIFF
--- a/include/osgUtil/CullVisitor
+++ b/include/osgUtil/CullVisitor
@@ -326,6 +326,10 @@ class OSGUTIL_EXPORT CullVisitor : public osg::NodeVisitor, public osg::CullStac
         osg::RenderInfo& getRenderInfo() { return _renderInfo; }
         const osg::RenderInfo& getRenderInfo() const { return _renderInfo; }
 
+        /** Provide a callback to be invoked during the render stages.  Provides an oppertunity to update custom per-context uniforms. */
+        void setRenderLeafCallback(osg::ref_ptr<RenderLeafCallback> renderLeafCallback) { _renderLeafCallback = renderLeafCallback; }
+        osg::ref_ptr<RenderLeafCallback> getRenderLeafCallback() const { return _renderLeafCallback; }
+
     protected:
 
         virtual ~CullVisitor();
@@ -419,6 +423,8 @@ class OSGUTIL_EXPORT CullVisitor : public osg::NodeVisitor, public osg::CullStac
         DistanceMatrixDrawableMap                                  _farPlaneCandidateMap;
 
         osg::ref_ptr<Identifier> _identifier;
+
+        osg::ref_ptr<RenderLeafCallback> _renderLeafCallback;
 };
 
 inline void CullVisitor::addDrawable(osg::Drawable* drawable,osg::RefMatrix* matrix)
@@ -474,13 +480,13 @@ inline RenderLeaf* CullVisitor::createOrReuseRenderLeaf(osg::Drawable* drawable,
     if (_currentReuseRenderLeafIndex<_reuseRenderLeafList.size())
     {
         RenderLeaf* renderleaf = _reuseRenderLeafList[_currentReuseRenderLeafIndex++].get();
-        renderleaf->set(drawable,projection,matrix,depth,_traversalOrderNumber++);
+        renderleaf->set(drawable,projection,matrix,depth,_traversalOrderNumber++, _renderLeafCallback);
         return renderleaf;
     }
 
 
     // Otherwise need to create new renderleaf.
-    RenderLeaf* renderleaf = new RenderLeaf(drawable,projection,matrix,depth,_traversalOrderNumber++);
+    RenderLeaf* renderleaf = new RenderLeaf(drawable,projection,matrix,depth,_traversalOrderNumber++, _renderLeafCallback);
     _reuseRenderLeafList.push_back(renderleaf);
 
     ++_currentReuseRenderLeafIndex;
--- a/include/osgUtil/RenderLeaf
+++ b/include/osgUtil/RenderLeaf
@@ -27,6 +27,7 @@ namespace osgUtil {
 
 // Forward declare StateGraph
 class StateGraph;
+class RenderLeafCallback;
 
 /** Container class for all data required for rendering of drawables.
   */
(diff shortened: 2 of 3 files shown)


mp3butcher:
I don't follow your reasoning... please could you give us a usage scenario?

EricStancliff:
Sure.  I was implementing several shaders that required the ModelView matrix with double precision, as well as some lighting parameters.  Rather than in every drawable implementation, adding uniforms for the same thing to the stateset, I wanted to have a set of "global uniforms" that I could also promise consumers of my SDK would be available in any shaders they may write in my pipeline.  It's essentially extensibility for the "osg_*" set of uniforms.  Below is a reduced version of what I've implemented in my application:

    class GMUniforms : public osgUtil::RenderLeafCallback
    {
    public:
        enum UNIFORM_TYPE
        {
            PROJECTION = 0,
            MODELVIEW,
            MVP,
            LIGHTPOS,
            EYEPOS,
            LIGHTCOLOR,
            LIGHTENABLED
        };

        GMUniforms() {
            std::get<PROJECTION>(m_uniforms) = new osg::Uniform(osg::Uniform::DOUBLE_MAT4, "GM_ProjectionMatrixD");
            std::get<MODELVIEW>(m_uniforms) = new osg::Uniform(osg::Uniform::DOUBLE_MAT4, "GM_ModelViewMatrixD");
            std::get<MVP>(m_uniforms) = new osg::Uniform(osg::Uniform::DOUBLE_MAT4, "GM_ModelViewProjectionMatrixD");
            std::get<LIGHTPOS>(m_uniforms) = new osg::Uniform(osg::Uniform::FLOAT_VEC3, "GM_LightPosition");
            std::get<EYEPOS>(m_uniforms) = new osg::Uniform(osg::Uniform::FLOAT_VEC3, "GM_EyePosition");
            std::get<LIGHTCOLOR>(m_uniforms) = new osg::Uniform(osg::Uniform::FLOAT_VEC4, "GM_LightColor");
            std::get<LIGHTENABLED>(m_uniforms) = new osg::Uniform(osg::Uniform::BOOL, "GM_LightEnabled");
        }
        ~GMUniforms() = default;

        osg::ref_ptr<osg::Uniform> getUniform(UNIFORM_TYPE type) const
        {
            return m_uniforms.at(type);
        }

        virtual void run(const osgUtil::RenderLeaf& renderLeaf, osg::RenderInfo& renderInfo) override
        {
            if (!renderInfo.getState() || !renderInfo.getCurrentCamera() || !renderInfo.getState()->getLastAppliedProgramObject())
                return;

            getUniform(PROJECTION)->set(renderInfo.getState()->getProjectionMatrix());
            getUniform(MODELVIEW)->set(renderInfo.getState()->getModelViewMatrix());
            getUniform(MVP)->set(renderInfo.getState()->getModelViewMatrix() * renderInfo.getState()->getProjectionMatrix());
           
            bool lightEnabled = renderInfo.getState()->getLastAppliedMode(GL_LIGHTING);
            if (lightEnabled && renderInfo.getView() && renderInfo.getView()->getLight())
            {
                osg::Vec4 lightPos4 = renderInfo.getView()->getLight()->getPosition();
                osg::Vec3 lightPos = osg::Vec3(lightPos4.x(), lightPos4.y(), lightPos4.z());
                lightPos = lightPos * renderInfo.getCurrentCamera()->getInverseViewMatrix();
                getUniform(LIGHTPOS)->set(lightPos);
                getUniform(LIGHTCOLOR)->set(renderInfo.getView()->getLight()->getDiffuse());
                getUniform(EYEPOS)->set((osg::Vec3)renderInfo.getCurrentCamera()->getInverseViewMatrix().getTrans());
            }
            getUniform(LIGHTENABLED)->set(lightEnabled);

            for (auto&& uniform : m_uniforms)
                renderInfo.getState()->getLastAppliedProgramObject()->apply(*uniform);
        }

    private:
       std::array<osg::ref_ptr<osg::Uniform>, LIGHTENABLED + 1> m_uniforms;

    };

mp3butcher:
I don't understand why you don't use a drawcallback on your geometry to do that?

EricStancliff:
The same reasons the OSG source doesn't update the "osg_ModelViewProjectionMatrix" uniform in the drawable implementation - Less overhead, less code repeat, and cleaner per-context implementation.

EricStancliff:
I see you added "(or subclass them)" to your explanation, which was considered, but unfortunately I'm working with a very large codebase (>1mSLOC) and refactoring all existing geometries would be a huge task.  Also, in order to avoid newing up a list of uniforms for each geometry per each context (MVP is different in each view, we can have many), it would involve some nasty implementation of a mutexed static map of contextId to uniform list.  I really just wanted to add/modify the existing "osg_*" uniforms, and followed that as the example.

openscenegraph:
I am closing this PR as it's a *really* bad for performance.  It's damage performance for ALL users for the benefit of obscure usage case that is likely better solved in other means anyway.  You absolutely don't add extra work to inner loops like RenderLeaf unless it's absolutely essential.

If you want to discuss the challenges you are trying to tackle then please raise a topic for disucssion on the osg-users googlegroup.

EricStancliff:
How is it bad for performance for everyone?  If you don't set the callback, it does literally nothing.  And if you shouldn't be doing things there, then why are you updating the osg_* variables there?  Clearly trying to tackle the same challenges I was. 

Managing a default set of available uniforms isn't obscure in OpenGL, that's why they got rid of the gl_ModelViewProjectionMatrix variable, and why you provide the 4 defaults.  It's reasonable to want to extend these.

mp3butcher:
You clearly don't understand the history of OpenGL and so on the difficulty to maintain osg across the time.
OpenGL Api was derived from a monopolistic api belonging to SGI. That's the reason of the fixed pipeline maintenance and the original purpose of the gl_* uniforms.

openscenegraph:
On Tue, 24 Mar 2020 at 21:42, EricStancliff <notifications@github.com>
wrote:

> How is it bad for performance for everyone? If you don't set the callback,
> it does literally nothing.
>
You have much to learn about programming.

Every if statement and data member we add adds both branching, cache and
bandwidt load, ones on inner loops are critical.  You might not see it if
you see has 10 objects in it, but if it has 10,000 or 100,000 then the
performance cost will be felt.

And if you shouldn't be doing things there, then why are you updating the
> osg_* variables there? Clearly trying to tackle the same challenges I was.
>
The osg_ uniforms exist for GLSL because OpenGL has evolved from providing
gl_ builtins to not providing them.


> Managing a default set of available uniforms isn't obscure in OpenGL,
> that's why they got rid of the gl_ModelViewProjectionMatrix variable, and
> why you provide the 4 defaults. It's reasonable to want to extend these.
>
The osg allows you to provide your own uiniforms via osg::StateSet.  This
is the standard way to extend things in the OSG.

You earlier cited precision of modelview as the reason for wanting to
providing your own variants.  This suggest to me that you probably aren't
aware of other techniques for avoid precision issues.  Plenty of people use
the OSG for GIS systems, it's designed speciifically to handle precisions
issues associated with very large whole earth or larger systems.  Please
check the osg-users archives for the discussions how to tackle it.  If you
want ask you own question about how do you do what you want to.  Please
explain at a high level problem not the low level solution that you have
adopted so far.

EricStancliff:
I considered the added branching and decided there were already plenty in that call.  I'm also aware that making an average transform to put world points under will solve the precision issue.  I wanted to preserve the ability of my users to add shapes in world space, while leaving the door open for default implementation of lighting/material uniforms.

I think for now I'll add a method to my view implementation and tell my users to call that method in their draw callbacks to populate uniforms where needed.

Thanks for the insight.

PR #943, opened 2020-04-13
Add option to treat all TGA files as TGA 1.0

AnyOldName3:
Some files which are malformed TGA 2.0 files are technically valid TGA 1.0 files. For example, an image that just happened to have pixel data matching the TGA 2.0 footer or files created by buggy TGA 2.0 writers and were only ever fed to TGA 1.0 readers.

One such example from a Morrowind mod is attached:
[tx_comberry_01_p.zip](https://github.com/openscenegraph/OpenSceneGraph/files/4471548/tx_comberry_01_p.zip)

This file has a header saying it has eight attribute bits. A TGA 1.0 reader is free to interpret them how it wants (e.g. as alpha or stencil, or just ignoring them). A TGA 2.0 reader is supposed to check for a footer directing it to an extension area, which, if present, will have a field saying what those bits mean. The file's extension area says it doesn't have any such bits (which is different from saying they're present but unused).

This is against the TGA 2.0 specification published by Truevision, and, in fact, `TGADUMP.exe` from their TGAUTILS package will complain about *Inconsistent Alpha Data Specification*. It's a valid TGA 1.0 file, though (albeit one with garbage data at the end), so a conformant TGA 1.0 reader will handle it just fine.

In TGA 2.0 mode (i.e. the existing code), the extension area will be trusted above the header, and the alpha will be skipped. If the new option string is set, though, only the header will be read, and the alpha will be used.

OpenMW needs this behaviour as it needs to work with files made for the original Morrowind engine, which only had a TGA 1.0 reader.

I'm open to suggestions on how to change particular details of this PR as I've not interacted with option strings in OSG before.

### DIFF
--- a/src/osgPlugins/tga/ReaderWriterTGA.cpp
+++ b/src/osgPlugins/tga/ReaderWriterTGA.cpp
@@ -339,7 +339,8 @@ unsigned char *
 simage_tga_load(std::istream& fin,
 int *width_ret,
 int *height_ret,
-int *numComponents_ret)
+int *numComponents_ret,
+bool ignoreTGA2Fields)
 {
     unsigned char header[18];
     unsigned char footer[26];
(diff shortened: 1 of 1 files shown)


src/osgPlugins/tga/ReaderWriterTGA.cpp
```diff
@@ -867,7 +872,7 @@ class ReaderWriterTGA : public osgDB::ReaderWriter
 
             osgDB::ifstream istream(fileName.c_str(), std::ios::in | std::ios::binary);
             if(!istream) return ReadResult::FILE_NOT_HANDLED;
-            ReadResult rr = readTGAStream(istream);
+            ReadResult rr = readTGAStream(istream, options && options->getOptionString().find("IGNORE_TGA_2_FIELDS") != std::string::npos);
```
Capostrophic, line 875: Options are typically not DONE_LIKE_MACROS in OSG. Rather, they use camelCase or, rarely, snake_case.

Capostrophic:
The readerwriter constructor (ReaderWriterTGA() in this case) should declare the support of this option with supportsOption("option name", "option description");

src/osgPlugins/tga/ReaderWriterTGA.cpp
AnyOldName3: I copied the JPEG one https://github.com/openscenegraph/OpenSceneGraph/blob/OpenSceneGraph-3.6/src/osgPlugins/jpeg/ReaderWriterJPEG.cpp#L811

Are you sure that's the outlier?

AnyOldName3:
> The readerwriter constructor (ReaderWriterTGA() in this case) should declare the support of this option with supportsOption("option name", "option description");

Again, I copied the JPEG one, which doesn't do this: https://github.com/openscenegraph/OpenSceneGraph/blob/OpenSceneGraph-3.6/src/osgPlugins/jpeg/ReaderWriterJPEG.cpp#L823-L827

AnyOldName3:
I just checked the PNG reader, too, and it's done like this PR, which no explicit option declaration, and with upper snake case.

Capostrophic:
But there's a reasonable number of readerwriters that do [otherwise.](https://github.com/openscenegraph/OpenSceneGraph/search?q=supportsoption&unscoped_q=supportsoption)

openscenegraph:
On Sat, 25 Apr 2020 at 07:56, Alexei Dobrohotov <notifications@github.com>
wrote:

> But there's a reasonable number of readerwriters that do otherwise.
> <https://github.com/openscenegraph/OpenSceneGraph/search?q=supportsoption&unscoped_q=supportsoption>
>
There have been 570 contributors to the OSG project over the years, so
there has been a few differences preferences expressed. For the core OSG
classes I have tried to keep things coherent but it's more work to manage
the necessary back and fore to get things ship shape.  With getting on for
a 87 plugins that primarily a community effort I haven't policed style so
much, for options there isn't any on standard,  Not ideal but the project
is 20 years old, you do get a few things that aren't prefect.

Capostrophic:
Yes, I understand, though I guess I just prefer the style I've noticed happening the most in the source code of the project and requested here.

AnyOldName3:
Which style of capitalisation is preferred for new stuff and is registering the option preferred?

openscenegraph:
On Sat, 25 Apr 2020 at 16:04, Chris Djali <notifications@github.com> wrote:

> Which style of capitalisation is preferred for new stuff and is
> registering the option preferred?
>
The core OSG plugin usesThisTypeOfOption so is probably preferable.  I
won't loose sleep over plugins that diverge from this.  It's probably the
one to copy for new additions.

As for declaring the option in the ReaderWriter* constructor
ReaderWriter::supportsOption(..) certainly use this for all options, that
way one doesn't need to go through the source code, instead you can run:

   osgconv --formats

To list the features and options of all supported plugins or

  osgconv -format osg

To list the options of a specific plugin.

AnyOldName3:
These comments have been applied.

Also, I noticed that other plugins used a string stream to search for options one token at a time, but I went for `std::string::find` as it would be faster if all I cared about was the option's existence. Is there any reason to prefer the other approach?

openscenegraph:
On Sat, 25 Apr 2020 at 20:10, Chris Djali <notifications@github.com> wrote:

> These comments have been applied.
>
> Also, I noticed that other plugins used a string stream to search for
> options one token at a time, but I went for std::string::find as it would
> be faster if all I cared about was the option's existence. Is there any
> reason to prefer the other approach?
>
The most flexible approach is to the fact that all osg::Object have a meta
data object system, this means osgDB::Options can hold other objects.  This
functionality was added later than osgDB::Options option string approach,
and isn't an explicit feature of osgDB::Options so can easily be overlooked.

The VSG also has an vsg::Options equivalent to osgDB::Option and rather use
option strongs it just leaves it to the meta object system for passing in
custom values/objects.

AnyOldName3:
The `osgDB::Registry` sets option strings based on the command line, making them usable in things like the `osgviewer` application. I don't see any way of setting up object metadata on the options object outside a custom C++ application, so wouldn't the approach taken here be more useful, at least for now?

openscenegraph:
On Mon, 27 Apr 2020 at 16:03, Chris Djali <notifications@github.com> wrote:

> The osgDB::Registry sets option strings based on the command line, making
> them usable in things like the osgviewer application. I don't see any way
> of setting up object metadata on the options object outside a custom C++
> application, so wouldn't the approach taken here be more useful, at least
> for now?
>

If you want to be able to leverage the crude -O option string passthrough
that osgviewer provides then yes, using option string leaves that door
open.  However, this little application is just an OSG example.

I guess in theory one could add support for doing something like --option
"key=value" and then parse the key and value.  However, I can consider the
OSG primarily in maintenance mode now so coming up with fancy new
mechanisms isn't something I'm going to worry about chasing as my time is
better spent getting the VSG to 1.0 ASAP.

AnyOldName3:
So surely I shouldn't be introducing new behaviour that isn't used anywhere else?

openscenegraph:
On Mon, 27 Apr 2020 at 16:38, Chris Djali <notifications@github.com> wrote:

> So surely I shouldn't be introducing new behaviour that isn't used
> anywhere else?
>
My own personal focus on maintenance.  I'm not planning on major changes to
the code base, and contributions from the community that involve major
changes will be looked at hard - they'll have really justify the risks and
rewards involved.

Small tweaks to make life easier, bug and build fixes are all perfectly OK
during this phase of the project.

At this point if you can consider us keeping the ship steady, doing minor
repairs while we are at sea when they are required, but not heading into
port for a major refit and lots of effort and downtime.

For uses that want to push the bleeding edge along the OSG isn't the
project, the VSG is for that.

AnyOldName3:
I don't want to push the bleeding edge along. This PR just makes it possible to load some images as intended that couldn't be before by filling in the edge-case of files that are valid TGA 1.0 images, but invalid TGA 2.0 images.

I haven't found any other plugins that use the object metadata system to access their options, so I think doing that would be a bad idea. It might be less reliable (as there are fewer examples to copy and other plugins haven't tested this use case as thoroughly as option strings), it could surprise users of OSG (as it's different behaviour to everything else), and it's slightly less useful as there are features that set up option strings integrated into other parts of OSG. There's also no way of using `supportsOption` to explicitly state that it should be passed via the object metadata system.

This PR as-is (i.e. using the option string) works and doesn't do anything novel. All the other plugins with options use the option string, and I've been able to find some (such as the SVG reader/writer) that extract their options via `std::string::find`. That aspect should be risk-free. As for TGA reading in general, I've contributed the most to this plugin in the past few years, so can confidently say that it's no more trustworthy than it was before.

openscenegraph:
Hi Chris,

It's OK this type of change isn't a major change, it's something I'm happy
to see.  Choice on use of Options string is fine too.  I'm just provide
extra background info so you know all the functionality available etc.

Robert.

Issue #926, opened 2020-03-15
[macOS Catalina 10.15] Only the bottom left quadrant of the screen is used


PetrBo:
Hello,
the `osgviewer` draws the scene only in the bottom left quadrant of the screen on macOS 10.15 Catalina (see the image). I have tried to build OSG  from the latest source 3.7.0, as well as the stable 3.6.5 version – nothing helped.

It seems it is somehow related to the issue:
- https://github.com/ioquake/ioq3/issues/422
- https://github.com/redeclipse/base/issues/920

<img width="1440" alt="macos_catalina" src="https://user-images.githubusercontent.com/5457878/76705366-d8d80f80-66df-11ea-92e0-d884c0aff4aa.png">

robertosfield:
I don't have an macOS system or expertise with macOS so have to defer to Mac users to figure out things.  My guess is that Apple have changed the way they manage the scaling of the display, I believe there is already code in osgViewer's Mac support for this, but Apple have a habit of moving the goal posts so it might be worth having a look at the OpenSceneGraph/src/osgViewer/GraphicsWindowCocoa.mm source to see what t's doing now, and then have a to see if any of these API have changed or new API introduced by Apple in recent revs that deprecate the original API usage.

tkittel:
Just to confirm that we see the same issue in our OSG-based application, also starting with OSX 10.15 Catalina. Indeed, this "moving the goal posts" seems to be a major sport at Apple.

PetrBo:
Hi, so I solved the issue by following (thanks @robertosfield for pointing out the `GraphicsWindowCocoa.mm`):

- in `OpenSceneGraph/src/osgViewer/GraphicsWindowCocoa.mm` I have added the line
`[_view setWantsBestResolutionOpenGLSurface: NO];`

- under the block:

```
[_view setAutoresizingMask:  (NSViewWidthSizable | NSViewHeightSizable) ];
[_view setGraphicsWindowCocoa: this];
[_view setOpenGLContext:_context];
```
Probably it's not the "ultimate" solution, but as a quick fix good enough.

openscenegraph:
Good to hear you've found a workaround.  Could you post a PR for this fix
so others can try out the fix to see it works fine on different macOS
versions and other software.  Thanks,

tkittel:
[...]

PR #944, opened 2020-04-18
[macosx 10.15] Fix GC resize issue #926 from @PetrBo and @tkittel

remoe:
made by @PetrBo @tkittel

Not tested with older MacOSX


robertosfield:
Do you think it's safe to apply to the 3.6 branch, and likely to work with older MacOSX?

tkittel:
It looks safe to me, but I currently have no way to verify.

PR #950, opened 2020-05-10
Add rvalue constructors and assignment operators to ref_ptr.

Zyrin:


openscenegraph:
I don't believe having the OSG behave differently for different compilers is at all appropriate.  If you want to add functionality like this then you should do it via your own template helper functions.

Zyrin:
I submitted this pull request, because I wanted to share this optimization with the community. If it is the ifdefs yout are worried about I only put those in place because I could not find any information about what c++ version OSG requires or supports.

openscenegraph:
On Sun, 10 May 2020 at 12:40, Alexis Tsogias <notifications@github.com>
wrote:

> I submitted this pull request, because I wanted to share this optimization
> with the community. If it is the ifdefs yout are worried about I only put
> those in place because I could not find any information about what c++
> version OSG requires or supports.
>
The OSG still supports pre C++11 compilers by design, we are working to
maintain compatibility with older systems that some OSG users are tied to.
The OSG project focus these days is maintaining compatibility and
reliability.

For current scene graph work using latest compiler feature have a look at
the VulkanSceneGraph.

Zyrin:
Ok, thanks for the feedback :)

PR #954, opened 2020-05-20
ive support animation

locobrain:


robertosfield:
The .ive format is deprecated and only kept around for backwards compatibility.  For modern scene graph features you should use .osgb and avoid .ive.

openscenegraph:
I'm closing this PR as neither set of changes look appropriate.  The .ive format is deprecated and only support for backwards compatibility.  You should be using .osgb file format.  The change for a virtual GraphicsThread could easily break other codes if the threading isn't set up correctly, it's supposed to be an feature orthogonal to implementation of the graphics context.  

In both cases these cases it raises a red flag that you may be using the OSG in an unintended and inappropriate way.

PR #955, opened 2020-05-20
Fix GLExtensions static order of deletion bug

timoore:
The s_glExtensionSetList could be deleted while pointers to
GLExtension objects were still held e.g., by
VertexArrayStateManager, causing a segfault. This patch uses an
observer pointer to not access already-deleted data.

### DIFF
--- a/include/osg/GLExtensions
+++ b/include/osg/GLExtensions
@@ -33,6 +33,7 @@
 #include <osg/MatrixTemplate>
 #include <osg/Matrixd>
 #include <osg/Matrixf>
+#include <osg/observer_ptr>
 
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


robertosfield:
Utgh, that's not a pretty workaround.

Could you provide an example or a description of how to reproduce the problem.  I'd much rather find a cleaner solution.

timoore:
On Thu, May 21, 2020 at 3:20 PM Robert Osfield <notifications@github.com>
wrote:

> Utgh, that's not a pretty workaround.
>
> Could you provide an example or a description of how to reproduce the
> problem. I'd much rather find a cleaner solution.
>
> Run any osgEarth program compiled against an OSG built with
OPENGL_PROFILE=GLCORE. I can't reproduce it with osgviewer; I can send you
the valgrind trace if you like. The problem is absolutely classic: the
static arrays in GLExtensions.cpp are deleted before the per-context data
in ContextData.cpp, which holds a pointer to a VertexArrayStateManger,
which holds a pointer to a GLExtension... which tries to remove itself from
the deleted array s_glExtensionSetList.

You can try replacing the static arrays in GLExtensions.cpp with data that
lives in ContextData, but I bet you that any solution is going to involve
observer_ptr.

Tim

robertosfield:
Hi @timoore,  Finally got back to reviewing this change again.  I now think it's a reasonable solution, I'll merge and might do so small code tweaks.

robertosfield:
FYI, I've merged this fix with the OpenSceneGraph-3.6 branch.

PR #962, opened 2020-06-04
suport more type user value

locobrain:
suport more type user value by SetScalarValue and GetScalarValue


openscenegraph:
In C++ you can't just go around static casing an object into another type and it expect it to be safe.  This modification is most definitely not safe, if you wanted an example of what no to do then this would be a good one to use.

If there is something not supported by the present meta object system please post your usage case to the osg-users googlegroup and we have walk you through what is required.

PR #968, opened 2020-07-01
Get names of bones and skeleton to osg nodes

unelsson:
Dae-plugin doesn't seem to transfer dae-node names to osg-nodes. This PR browses through dae-object, and sets the names for the skeleton and bones.


psi29a:
Thanks @unelsson and @robertosfield

PR #973, opened 2020-07-14
fixed installation of PDB files on Windows with Ninja generator.

arnauddgbes:
Also added installation of PDB files for OpenThreads


robertosfield:
Checking master I see more extensive support for Ninja, have you tried master?  Are these changes backported?

Any changes to 3.6 I'm reluctant to make as it endangers breaking the build next time we make an maintenance release.

arnauddgbes:
This fix is ported almost identically from my pull request for the master that was merged on Oct 4, 2019.
I tested again this fix in this branch using ninja and MSVC generator, without issue.

robertosfield:
HI @arnauddgbes, apologies for the slow turnaround.  I'm just doing another review.  

Why do you define a  ${INSTALL_BINDIR} to bin and then reuse this variable but never ever change its value.  Seems like a it extra extra complexity for no value.

PR #974, opened 2020-07-17
Enabled windows clang builds.

MeyerFabian:
Hey, I would like to enable windows clang builds (https://github.com/llvm/llvm-project) of openscenegraph. 

**I am gonna shortly list up what I had to do:**
1. Replaced instances of ``WIN32`` with ``_WIN32``. (https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c) AFAIK Visual Studio defines both, clang only defines ``_WIN32``. Alternatively we could do sth. like 
``` cpp
#if defined(__clang__) && defined(_WIN32) // or defined(_MSC_VER)
    #define WIN32 
```
2. Clang complaines about assigning negative values to an ``unsigned int``: https://github.com/openscenegraph/OpenSceneGraph/blob/d5225ea534d44cc3dce2f26f4255b90c49244469/src/osgViewer/GraphicsWindowWin32.cpp#L849

3. Clang defines ``_MSC_VER`` to be MSVC compatible. Thus I handled some type warnings where clang does not need MSVC stuff. Could just leave these changes out.
```cpp 
#if defined(__BORLANDC__) || (defined(_MSC_VER) && !defined(__clang__))
```

**Tests**
1. Visual Studio 2019 worked before this pull request and after. (Tested some examples)
2. clang-10 works after this pull request. (Tested some examples)

Very open to other ideas that would fix the problems mentioned.


robertosfield:
Changes look sound.  Should be safe to apply to the 3.6 branch as well.  Have you tried it?

MeyerFabian:
No. I will do so. Should I then reopen this to target 3.6?

PR #975, opened 2020-07-19
Update IntersectionVisitor.cpp

ifcquery:
Prevent crash here: 
Drawable might be 0, because getDrawable might return 0: 
getDrawable( unsigned int i ) { return _children[i].valid() ? _children[i]->asDrawable() : 0; }

### DIFF
--- a/src/osgUtil/IntersectionVisitor.cpp
+++ b/src/osgUtil/IntersectionVisitor.cpp
@@ -236,7 +236,13 @@ void IntersectionVisitor::apply(osg::Geode& geode)
 
     for(unsigned int i=0; i<geode.getNumDrawables(); ++i)
     {
-        intersect( geode.getDrawable(i) );
+        // drawable might be 0, because getDrawable might return 0: 
+		// getDrawable( unsigned int i ) { return _children[i].valid() ? _children[i]->asDrawable() : 0; }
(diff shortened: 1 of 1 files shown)


robertosfield:
A nullptr Drawable in a Geode suggests a bug elsewhere for some reason and is an invalid scene graph, this might fix the crash for you but it doesn't fix the real problem - how this invalid scene graph came about.

Added checks for potentially invalid scene graphs everywhere is not a good use of CPU resource and code complexity, it's far better to fix the problem at source i.e. avoid creating invalid scene graphs in the first place.

ifcquery:
Thank you for your great work fist!
And I agree, checks during draw traversal are not a good idea.
But there could be a check in Geode::addChild, allowing only Drawables, not something that leads to a crash sometime later.
I had a case where a Geode had another Geode as child, which is allowed since Geode is derived from Group and that lead to the crash.
Is it possible to un-derive Geode from Group? It is a bit dangerous, allowing for inconsistent scenegraphs

robertosfield:
I have applied a more flexible and reliable fix:

   https://github.com/openscenegraph/OpenSceneGraph/commit/324750fded8cc28f158c5cf6aa00481fd795c2ed

PR #995, opened 2020-11-12
Fix for FileUtils.cpp not compiling under windows

tomhog:
Appears to just be a typo


robertosfield:
Thanks. I've merged with 3.6 branch as well.

PR #999, opened 2020-11-17
ProxyNode Failed to load child node

longhuan2018:
When the LoadingExternalReferenceMode value of ProxyNode is not LOAD_IMMEDIATELY, if children's filename is a relative path, then when it calls getDatabaseRequestHandler()->requestNodeFile to load, unless osgDB::Registry::instance()->setDataFilePathList is set, it will not be able to The relative path is converted to the correct absolute path, the loading will fail.

### DIFF
--- a/src/osgWrappers/serializers/osg/ProxyNode.cpp
+++ b/src/osgWrappers/serializers/osg/ProxyNode.cpp
@@ -109,16 +109,20 @@ struct ProxyNodeFinishedObjectReadCallback : public osgDB::FinishedObjectReadCal
     {
         osg::ProxyNode& proxyNode = static_cast<osg::ProxyNode&>(obj);
 
+        if (proxyNode.getDatabasePath().empty())
+        {
+            osgDB::FilePathList& fpl = ((osgDB::ReaderWriter::Options*)is.getOptions())->getDatabasePathList();
+            if (!fpl.empty())
+                proxyNode.setDatabaseOptions((osgDB::ReaderWriter::Options*)is.getOptions());
+        }
+
         if (proxyNode.getLoadingExternalReferenceMode() == osg::ProxyNode::LOAD_IMMEDIATELY)
         {
             for(unsigned int i=0; i<proxyNode.getNumFileNames(); i++)
             {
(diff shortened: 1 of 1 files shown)


src/osgWrappers/serializers/osg/ProxyNode.cpp
```diff
@@ -109,16 +109,20 @@ struct ProxyNodeFinishedObjectReadCallback : public osgDB::FinishedObjectReadCal
     {
         osg::ProxyNode& proxyNode = static_cast<osg::ProxyNode&>(obj);
 
+        if (proxyNode.getDatabasePath().empty())
+        {
+            osgDB::FilePathList& fpl = ((osgDB::ReaderWriter::Options*)is.getOptions())->getDatabasePathList();
```
openscenegraph, line 114: I's inappropriate to just case away constness with a C cast.  A InputStream is const for a reason - loaders can be run multi-threaded.  osgDB::Options supports a cloneOptions() method that should be used if you wish to use the Options elswhere such as assigning it to another object.

src/osgWrappers/serializers/osg/ProxyNode.cpp
```diff
@@ -109,16 +109,20 @@ struct ProxyNodeFinishedObjectReadCallback : public osgDB::FinishedObjectReadCal
...
+                proxyNode.setDatabaseOptions((osgDB::ReaderWriter::Options*)is.getOptions());
+        }
+
         if (proxyNode.getLoadingExternalReferenceMode() == osg::ProxyNode::LOAD_IMMEDIATELY)
         {
             for(unsigned int i=0; i<proxyNode.getNumFileNames(); i++)
             {
                 if(i >= proxyNode.getNumChildren() && !proxyNode.getFileName(i).empty())
                 {
-                    osgDB::FilePathList& fpl = ((osgDB::ReaderWriter::Options*)is.getOptions())->getDatabasePathList();
```
openscenegraph, line 118: Why remove these lines?

openscenegraph:
I'm closing this PR as it's hacky and inappropriate.  I've added explanation to the problem lines.

Issue #992, opened 2020-11-04
osgText fails to use FBOs when it should


jpabst0:
I have observed the issue using OpenSceneGraph 3.6.5.
I have set up scene where I render a Text object after a Geometry object.
The Text object is set up to use FBOs while the Geometry object is set up to not use FBOs.
VAOs are not used.

Now the Text object fails to use it's FBO when dispatching the _coords / _texcoords array.
The corresponding log messages are:

VertexArrayDispatch::enable_and_dispatch(108)
TexCoordArrayDispatch::enable_and_dispatch(108) unit=0

If the Geometry object is changed to also use FBOs, suddenly the Text object uses its FBO properly:

VertexArrayDispatch::enable_and_dispatch(184, vbo=000001B8457C07C0)
TexCoordArrayDispatch::enable_and_dispatch(184, vbo=000001B8457C07C0) unit=0

After some debugging I found out what's wrong.
In the State object, the global VertexArrayState object is set current.
It isn't changed throughout all drawing as VAOs are not used.
When the Geometry object is drawn, it disables VBO dispatching in the global VertexArrayState object.
Here is the corresponding code:
 
State& state = *renderInfo.getState();
bool usingVertexBufferObjects = state.useVertexBufferObject(_supportsVertexBufferObjects && _useVertexBufferObjects);
osg::VertexArrayState* vas = state.getCurrentVertexArrayState();
vas->setVertexBufferObjectSupported(usingVertexBufferObjects);

The Text object also dispatches it's arrays using the global VertexArrayState object.
It doesn't make a call to vas->setVertexBufferObjectSupported however.
So the boolean flag that tells whether to use FBOs is left disabled from the Geometry object drawn before.

A possible fix might be adding the line
 
vas->setVertexBufferObjectSupported(usingVertexBufferObjects);

also to both, Text and Text3d, classes. I wonder however, if there arn't better (less viral) solutions.

openscenegraph:
I have no clue why you are using the word "FBO" in this context.

FBO is my world means Frame Buffer Object. It's really not something that
Text knows about or uses, or should use other than to render to if you are
doing some high level render to texture effect.

jpabst0:
I'm sorry. I mixed up VBO and FBO in the text. Of course I meant VBO. I fixed title and description.

robertosfield:
Could you try the OpenSeneGrah-3.6 branch head, it has a few fixes over 3.6.5.  This I one made back in January might be relevant:

[...]

PR #1000, opened 2020-11-17
Fix VBO dispatching

jpabst0:
Here are a few points to summarize my assumptions and understanding of the code.

* When VAOs are used, VBOs are used (if supported) regardless off the usingVertexBufferObjects setting.

This is status quo and seems reasonable. However, the usingVertexBufferObjects stetting still determines if EBOs are used.

So, if one sets up an Drawable

drawable->setUseVertexBufferObjects(false);
drawable->setUseVertexArrayObject(true);

one would get VAOs with VBOs, but not EBOs.

I don't know what is really intended and if such setup is practically used. (The compileGLObjects functions don't seem to handle it either.) I think it would make sense to also use EBOs, as it's a more efficient use of VAOs. To achieve this, one can simply replace in drawImplementations the line

bool usingVertexBufferObjects = state.useVertexBufferObject(_supportsVertexBufferObjects && _useVertexBufferObjects);

with this one:

bool usingVertexBufferObjects = vas->getVertexBufferObjectSupported();

Anyway, one should be aware that the variable usingVertexBufferObjects is somewhat misleading because it doesn't tell whether VBO are really used.

* When not using VAOs, the usingVertexBufferObjects setting determines if VBOs and EBOs are used.
 
 I added these lines to Drawable::draw:

  vas->setVertexBufferObjectSupported(usingVertexBufferObjects);
  vas->setRequiresSetArrays(true);

 The first fixes the missing setup so that the VBOs get properly dispatched, the latter helps setting up the vas object transparently for subclass drawImplementations.

It simplifies the condition

!usingVertexArrayObjects || vas->getRequiresSetArrays()

to  

vas->getRequiresSetArrays()

* I couldn't figure out why the VBO's are unbound at the end of each drawImplementation. Shouldn't the VertexArrayState and it's lazy update mechanism handle this?  I guess it is for compatibility with custom Drawables.

Actually I discovered a reason to unbind the VBO when VAOs are used. In OpenGL, the bound VBO is not part of the VAO state while in OSG it is modeled so. Not unbinding the VBO can cause it to not get bound in the next frame, if it is falsely assumed to not have changed. I will provide a detailed description and an example program in the issue report.

As simplification and workaround I replaced the lines

if  (usingVertexBufferObjects && !usingVertexArrayObject)
{
        // unbind the VBO's if any are used.
        vas->unbindVertexBufferObject();
        vas->unbindElementBufferObject();
 }

with these ones

   // unbind the VBO's if any are used.
    vas->unbindVertexBufferObject();
    if (!state.useVertexArrayObject(_useVertexArrayObject)) vas->unbindElementBufferObject();

Note that in case of using VAOs, the unbindVertexBufferObject function is a no-op, except for the case that one of the VBOs really got updated. This is even the case when the data variance is set to DYNAMIC.  So it shouldn't hurt performance or undermine the purpose of VAOs.

When not using VAOs, both buffer objects are unbound like it has been the case before.

### DIFF
--- a/include/osg/Drawable
+++ b/include/osg/Drawable
@@ -575,16 +575,20 @@ inline void Drawable::draw(RenderInfo& renderInfo) const
         return;
     }
 
+    bool usingVertexBufferObjects = state.useVertexBufferObject(_supportsVertexBufferObjects && _useVertexBufferObjects);
+
     // TODO, add check against whether VAO is active and supported
-    if (state.getCurrentVertexArrayState())
+    osg::VertexArrayState* vas = state.getCurrentVertexArrayState();
+    if (vas)
     {
-        //OSG_NOTICE<<"state.getCurrentVertexArrayState()->getVertexArrayObject()="<< state.getCurrentVertexArrayState()->getVertexArrayObject()<<std::endl;
-        state.bindVertexArrayObject(state.getCurrentVertexArrayState());
+        //OSG_NOTICE<<"vas->getVertexArrayObject()="<< vas->getVertexArrayObject()<<std::endl;
+        vas->setVertexBufferObjectSupported(usingVertexBufferObjects);
+        vas->setRequiresSetArrays(true);
+        state.bindVertexArrayObject(vas);
     }
 
-
 #ifdef OSG_GL_DISPLAYLISTS_AVAILABLE
-    if (!state.useVertexBufferObject(_supportsVertexBufferObjects && _useVertexBufferObjects) && _useDisplayList)
+    if (!usingVertexBufferObjects && _useDisplayList)
     {
         // get the contextID (user defined ID of 0 upwards) for the
         // current OpenGL context.
--- a/src/osg/Drawable.cpp
+++ b/src/osg/Drawable.cpp
@@ -687,16 +687,20 @@ void Drawable::draw(RenderInfo& renderInfo) const
         return;
     }
 
+    bool usingVertexBufferObjects = state.useVertexBufferObject(_supportsVertexBufferObjects && _useVertexBufferObjects);
+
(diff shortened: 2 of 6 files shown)


openscenegraph:
I have done a first pass review and build/tested the example mentioned in #992.  The changes are significant and non obvious enough that I will need to really deeply examine the issue - it'd be very easy to "fix" one usage case but then break others in subtle and difficult to track down ways.

For now I'll move on to review other PR's that more straightforward to review as this one could take a while to properly review and test and consolidate on what should be done.

valid-ptr:
I've applied this PR in november 2020 and there was no difference in my apps.
**NOTE**: I use VAO everywhere where possible.

robertosfield:
@valid-ptr What you say there was no difference, what do you mean?  No fixes to problems you see, or no new problems appeared?  What about performance?

robertosfield:
@jpabst0 This PR really touches on the design/implementation of the VertexArrayState which has some complex interactions - it's all so complicated because OpenGL/OSG has evolve from fixed function, no vertex arrays, to supporting vertex arrays, to supporting vertex buffer objects, to supporting vertex array objects.  Trying to keep all the different usage cases working has been a bit of nightmare.

To properly dive into this topic will likely take me a day or two, something I can't afford to do right away - I have to juggle my time between the VSG and OSG open source work, with bits of paid client work that pays for it, so far this month I'm two weeks of unpaid open source work vs 1 day paid work, so can't go indefinitely doing support.

valid-ptr:
> What you say there was no difference, what do you mean? No fixes to problems you see, or no new problems appeared? What about performance?

1) no new problems appeared
2) same performance

PR #1001, opened 2020-11-19
State's popAllStateSets fixed: should not reset _lastAppliedProgramOb…

valid-ptr:
…ject to zero. As result: osg::State synced with OpenGL state => no unnecessary (redundant) glUseProgram calls

### DIFF
--- a/src/osg/State.cpp
+++ b/src/osg/State.cpp
@@ -541,8 +541,6 @@ void State::popAllStateSets()
 
     applyProjectionMatrix(0);
     applyModelViewMatrix(0);
-
-    _lastAppliedProgramObject = 0;
 }
 
 void State::popStateSet()


openscenegraph:
You worried about a possible redundent OpenGL call being done once per frame?

I'm going to close this PR as there are potential problems in removing something without wider testing for little real word gain.

valid-ptr:
`_lastAppliedProgramObject = 0` without `glUseProgram(0) `can be just wrong.
osg::State **MUST be synced** with OpenGL state.
This cause potential problems, not vice versa.

This PR solves also some problems with enabled ShaderComposer as I remember...

Again... You decide here

openscenegraph:
On Tue, 19 Jan 2021 at 12:44, Konstantin S. Matveyev <
notifications@github.com> wrote:

> _lastAppliedProgramObject = 0 without glUseProgram(0) can be just wrong.
>
State::popStateSets() is essentially a reset of State.  The State::apply()
is what normally would apply the state variables.

The State has an attribute map's that popStateSet() and normally would
dirty the associated attribute and when the next apply happens should apply
the correct type.

I'm a bit rusty on the interaction with ProgramObject, been a number of
years since I dug over the code.  If there are cases that it's not working
then we need to pinpoint these.


> osg::State *MUST be synced* with OpenGL state.
> This cause potential problems, not vice versa.
>
> This PR solves also some problems with enabled ShaderComposer as I
> remember...
>
> Again... You decide here
>

The OSG is in maintenance mode, as much as possible I'm trying to keep it
working.  Each of the 3.6.x series I've down dozens and dozens of calls for
testing and not made releases till no problems have been reported.  Any
change to the code has the potential for breaking things, a fix for one
problem could easily cause a bug elsewhere, every change introduces a
change of regression.  Being conservative about what changes are merged is
the default approach I will continue to take.

So... if we if there is problem that it fixes then I'm open, but we really
need to make sure there is no regression.

PR #1002, opened 2020-11-19
osg::Viewport's constructor with no arguments should create an invali…

valid-ptr:
…d Viewport: this will reduce OpenGL calls

I've found many unnecessary
`glViewport(x = 0, y = 0, width = 800, height = 600)`
calls while apitracing.

NOTE:
This comes from "StateAttribute mechanism" on the osg::State.
`global_default_attribute = new osg::Viewport()` in two words

PS
I don't see the reason why the default Viewport must be 800x600

### DIFF
--- a/src/osg/Viewport.cpp
+++ b/src/osg/Viewport.cpp
@@ -18,8 +18,8 @@ Viewport::Viewport()
 {
     _x = 0;
     _y = 0;
-    _width = 800;
-    _height = 600;
+    _width = 0;
+    _height = 0;
 }
 
 
@@ -29,6 +29,9 @@ Viewport::~Viewport()
 
 void Viewport::apply(State&) const
 {
+    if (!valid())
+        return;
+
     glViewport( static_cast<GLint>(_x),static_cast<GLint>(_y),
                 static_cast<GLsizei>(_width),static_cast<GLsizei>(_height) );
 }


robertosfield:
Could you qualify "many"?  Are we talking once per frame?  Thousands of times per frame?  

800 x 600 is simply a default that should work on most displays, it just a fallback with half sensible aspect ratio.  The intention is for applications to always set their own value.

My concern with applying this PR is that we could break code that inadvertently relies upon a valid default Viewport.  The are tens of thousands of OSG applications out there.  We have to be really carefully about just changing stuff as we can't just get thousands of developers to retest their applications.

robertosfield:
Have you actually benchmarked a real application before and after this change?

If you really care about picoseconds then you are using the wrong API, OpenGL/OSG are always going to be significantly CPU limited, not because of operations done a couple times a frame like Viewport settings but because of the cost of traversing the scene graph and dispatch textures, uniforms, geometry to the GPU.

If you really care about performance then your focus should be on Vulkan/VSG.  It's not just a couple of % faster here or there,  100% to 1000% faster is the range you'd be looking at.  If you want to unleash the full performance of your hardware it's simply not possible with OpenGL/OSG.

valid-ptr:
Your decision! I just vote here for the clear OpenGL state. Maybe osg::State should be changed (prevent unnecessary calls), but it’s another question.

PR #1003, opened 2020-11-20
{s,g}etDivisor methods added to osg::Array. Use it instead of osg::Ve…

valid-ptr:
…rtexAttribDivisor which does not work with VAO

PS
Should VertexAttribDivisor by marked as **DEPRECATED**?

### DIFF
--- a/include/osg/Array
+++ b/include/osg/Array
@@ -140,6 +140,7 @@ class OSG_EXPORT Array : public BufferData
             _dataSize(dataSize),
             _dataType(dataType),
             _binding(binding),
+            _divisor(0),
             _normalize(false),
             _preserveDataType(false) {}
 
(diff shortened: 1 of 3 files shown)


robertosfield:
Changing the API in a stable branch should only be done for pressing matters like bugs.

This PR would introduce a bug in the case where one geometry sets the VertexAttribDivisor but then another doesn't have a _divisor great than zero so would inherit the previously applied valid value.  

If we were to be this functionality from osg::VertexAttribDivisor into osg::Array then we'd have to apply the value for all osg::Geometry.  This is forcing an API calls from every geometry array which is a big step backwards performance wise - most applications will never use this state so for everyone to have to pay a penalty for the small % of users that would use it is far from ideal.

I'm closing this PR I it breaks the API for a non bug fix, is buggy, and a higher level doesn't have clear merit as a replacement for the existing API.

valid-ptr:
**osg::VertexAttribDivisor + VAO is a bug!**
I think, osg::VertexAttribDivisor must be marked as deprecated. At least, **Divisor** must not be "StateAttribute".
Divisor must be presented in the **Array**, **Geometry** or **VAS** maybe.
I think, **Array** is a good place :)

PR #1005, opened 2020-11-26
ASTC Compression Support Added; GLExtensions improvements

valid-ptr:


src/osg/Texture.cpp
```diff
@@ -285,61 +315,102 @@ void Texture::TextureObject::setAllocated(GLint     numMipmapLevels,
...
+        case GL_COMPRESSED_RG11_EAC:                       numBitsPerTexel = 8.0;  break;
+        case GL_COMPRESSED_SIGNED_RG11_EAC:                numBitsPerTexel = 8.0;  break;
+
+        // ASTC
+        case GL_COMPRESSED_RGBA_ASTC_4x4_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_4x4_KHR:      numBitsPerTexel = 8.0;  break;
+        case GL_COMPRESSED_RGBA_ASTC_5x4_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x4_KHR:      numBitsPerTexel = 6.4;  break;
+        case GL_COMPRESSED_RGBA_ASTC_5x5_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x5_KHR:      numBitsPerTexel = 5.12; break;
```
robertosfield, line 391: This really doesn't feel like a good solution.  It's block compression, each block is always fits to a defined alignment of the block, it's not a ratio.

A better way would be to specify the block size and the dimensions of the xy block, then dived the widh, heigh by the number of pixels per block, round to the block size then compute the final size.

valid-ptr: I've tested this solution, It works for different astc images in KTX container.


Maybe just to add such comments:
```

case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x4_KHR:      numBitsPerTexel = 6.4;  break;   // 128.0/(5*4) = 6.4
case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x5_KHR:      numBitsPerTexel = 5.12; break;   // 128.0/(5*5) = 5.12

```
?

There is a table (**Block size -> Bits per pixel**) in specs:
https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_texture_compression_astc_hdr.txt
https://www.khronos.org/opengl/wiki/ASTC_Texture_Compression

robertosfield: That doesn't fix my concerns, it's the wrong way to compute sizes required.  The x and y dimensions both have to be independently clamped to the block size.

src/osg/Texture.cpp
```diff
@@ -285,61 +315,102 @@ void Texture::TextureObject::setAllocated(GLint     numMipmapLevels,
...
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x8_KHR:     numBitsPerTexel = 1.6;  break;
+        case GL_COMPRESSED_RGBA_ASTC_10x10_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x10_KHR:    numBitsPerTexel = 1.28; break;
+        case GL_COMPRESSED_RGBA_ASTC_12x10_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_12x10_KHR:    numBitsPerTexel = 1.07; break;
+        case GL_COMPRESSED_RGBA_ASTC_12x12_KHR:
+        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_12x12_KHR:    numBitsPerTexel = 0.89; break;
     }
 
     _size = (unsigned int)(ceil(double(_width * _height * _depth * numBitsPerTexel)/8.0));
```
valid-ptr, line 416: Do you mean, _size calculation is wrong here for ASTC format?

robertosfield: I just explain what is wrong above.  The maths used doesn't work for block formats.  The correct way would be to compute the number of blocks in width and height separately, then multiple up to compute the final size.

valid-ptr: I understand you.
What about this implementation:

```
        // ASTC
        case GL_COMPRESSED_RGBA_ASTC_4x4_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_4x4_KHR:
        case GL_COMPRESSED_RGBA_ASTC_5x4_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x4_KHR:
        case GL_COMPRESSED_RGBA_ASTC_5x5_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_5x5_KHR:
        case GL_COMPRESSED_RGBA_ASTC_6x5_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_6x5_KHR:
        case GL_COMPRESSED_RGBA_ASTC_6x6_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_6x6_KHR:
        case GL_COMPRESSED_RGBA_ASTC_8x5_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_8x5_KHR:
        case GL_COMPRESSED_RGBA_ASTC_8x6_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_8x6_KHR:
        case GL_COMPRESSED_RGBA_ASTC_8x8_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_8x8_KHR:
        case GL_COMPRESSED_RGBA_ASTC_10x5_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x5_KHR:
        case GL_COMPRESSED_RGBA_ASTC_10x6_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x6_KHR:
        case GL_COMPRESSED_RGBA_ASTC_10x8_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x8_KHR:
        case GL_COMPRESSED_RGBA_ASTC_10x10_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_10x10_KHR:
        case GL_COMPRESSED_RGBA_ASTC_12x10_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_12x10_KHR:
        case GL_COMPRESSED_RGBA_ASTC_12x12_KHR:
        case GL_COMPRESSED_SRGB8_ALPHA8_ASTC_12x12_KHR:
        {
            _size = 0;
            for (int i = 0; i < std::max(_numMipmapLevels, 1); ++i)
            {
                GLint blockSize;
                GLint size;
                getCompressedSize(_internalFormat, _width/(1 << i), _height/(1 << i), _depth, blockSize, size);
                _size += size;
            }
            return;
        }
```

robertosfield: Looking at Texture::getCompressedSize(..) it looks broken when handling the ASTC formats as well. The code looks like it could all do with a refactor to properly handling block formats of different block dimensions and total block memory size.

robertosfield: FYI, with the VSG I implemented the compressed blocks as types much like vec3, vec4 etc, and have the image data stored in a Array,Array2D/Array3D.  The width, height are actually done in number of blocks, the final width passed to Vulkan for the image is the computed from the block size multiplied the block dimensions.  This avoids any problems with potentially data sizes being off.  There is none of this complex switch/if/else handling of block formats.

Alas trying to retrofit something like this to the OSG would required lots of changes to how osg::Image is managed so is not a practical solution.

It might gives us pause though - perhaps the OSG scheme is just broken.  I worked in adhoc way for the STC orginal formats but simply doesn't scale to handling all these new formats.  Trying to get it to scale is possibly the wrong approach, it perhaps should just be generalized in a better way to handle the different block formats.

What this might be I can't say.  I need to get back to VSG work now so will wrap up.

Issue #994, opened 2020-11-07
daePlugin convert istringstream occasionally fails


unelsson:
In project OpenMW, we are using virtual file system for storing collada files. When loading from cache, istringstream method `convert` fails, giving error messages:
`Error: Extra content at the end of the document`
`Error: Failed to load XML document from memory`

My guess is that collada-dom is expecting the end of file or end of line character or something like that. I'm not sure if this bug is related to using vfs, or if it should be corrected in OSG repository.

A very "classic" fix works here through: Adding "+1" to this line fixed random loading from cache errors:
https://github.com/openscenegraph/OpenSceneGraph/blob/9dad0b4148249ca1379530365611abc183131326/src/osgPlugins/dae/daeReader.cpp#L297

`std::streampos length = static_cast<int>(fin.tellg()) + 1;`

openscenegraph:
THis looks more like a "classic hack" rather than a "classic fix".  It
would be worth looking to see if it's an issue with the daeString not being
null terminated.  If so the fix should be to create a buffer one longer and
then explicitly set the last character 0.

unelsson:
The issue is indeed related to lack of null-termination. A decent fix is adding `buffer.emplace_back('\0');` here which does exactly what you suggest. Should I make a PR?

I am not experienced in handling `std::istream`, but to my understanding, it is terminated by `\0`, and the function `bool daeReader::convert( std::istream& fin )` uses C string handling methods that always create `std::vector<char> buffer` without the null in the end. That sounds like it's guaranteed to fail?

unelsson:
I made a PR that adds `buffer.emplace_back('\0');` for convenience.
https://github.com/openscenegraph/OpenSceneGraph/pull/1012

PR #1012, opened 2020-12-13
Terminate C string

unelsson:
Fix to issue https://github.com/openscenegraph/OpenSceneGraph/issues/994

Adds a null to the end of the C string. This has been tested to work project OpenMW.

### DIFF
--- a/src/osgPlugins/dae/daeReader.cpp
+++ b/src/osgPlugins/dae/daeReader.cpp
@@ -300,6 +300,7 @@ bool daeReader::convert( std::istream& fin )
     // use a vector as buffer and read from stream
     std::vector<char> buffer(length);
     fin.read(&buffer[0], length);
+    buffer.emplace_back('\0');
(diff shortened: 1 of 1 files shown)


psi29a:
Thanks @unelsson

src/osgPlugins/dae/daeReader.cpp
```diff
@@ -300,6 +300,7 @@ bool daeReader::convert( std::istream& fin )
     // use a vector as buffer and read from stream
     std::vector<char> buffer(length);
```
glebm, line 301: Initialize to `length + 1` to avoid reallocation?
Then there is no need to append '\0' either.

glebm: I've sent #1024 with this and other fixes

robertosfield:
I'm closing this PR as #1024 look more better.

Issue #1013, opened 2020-12-21
Texture2DArray fails to regenerate after a releaseGLObjects/dirtyTextureObject (possible solution)


gwaldron:
If you use a `Texture2DArray`, and call `releaseGLObjects()` on it (which calls `dirtyTextureObject`), the texture fails to recompile properly from that point forward.

I believe the problem is here:
https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Texture2DArray.cpp#L403

When it allocates a new texture object, it checks the component images' modification flags before subloading. This prevents the subload from running unless the user manually calls `image->dirty()` on each component image. (Note, `Texture2D` and `Texture3D` do not check the modified count -- they skip this test in the same situation.)

Let me know if you agree and I can submit a PR. Thanks. (master branch)

robertosfield:
A quick look at the code and I think this branch is only ever called when a
textureObject is null so it wouldn't make sense to check the dirty count as
the the data needs subheading no matter what, so I think removing the check
is the correct thing to do.

PR #1014, opened 2020-12-24
Fix for #1013 - Texture2DArray fails to regenerate after a releaseGLO…

gwaldron:
…bjects/dirtyTextureObject - remove the modified count condition on component images when subloading when there is no texture object


robertosfield:
Thanks fix merged with 3.6 branch as well.

PR #1015, opened 2020-12-29
Wrong argument name rename: uniformMap -> defineMap

valid-ptr:


robertosfield:
Thanks, fix merged with 3.6 and master.

PR #1019, opened 2021-01-10
Fix

madsbuvi:
When rendering stereo, OSG runs the drawPreRenderStages() of both eyes before continuing to the draw() of either. This causes the second eye's prerender RTT operations to overwrite the first eye's. This breaks any RTT operations that depend on the view/projection of the corresponding eye, such as the water shader in OpenMW.

I have not fixed the interlace/checkerboard case as i do not understand the stencil code well enough to suggest a re-ordering.

I also fixed a CMake error i encountered on the way.


psi29a:
Thanks @madsbuvi for helping to upstream this.

robertosfield:
This would suggest that the two rendering stages are modifying the same state, which they in general shouldn't be.  It could be that something in the scene graph that is not handling use in a stereo configuration, the proposed change just looks to be a workaround for that.

The CMake should also be submitted as a separate PR.  For that one it would be worth explaining the circumstances for this issue as the OSG has been used by lots of folk without anyone raising an issue on it.

I'm closing this PR as it conflates two topics and I'm not clear that it's not an application level issue, rather than core scene graph issue.

AnyOldName3:
What's happening here is that an RTT needs redoing for each eye. It's better for memory usage etc. to reuse the same texture twice a frame, i.e. doing the left RTT pass, then the left main pass, then the right RTT pass, then the right main pass as there's no interdependency between the two. However, with OSG's current setup, that's not an option, and the RTT pass needs to be made stereo-aware and have two copies of everything.

openscenegraph:
On Mon, 18 Jan 2021 at 16:05, Chris Djali <notifications@github.com> wrote:

> What's happening here is that an RTT needs redoing for each eye. It's
> better for memory usage etc. to reuse the same texture twice a frame, i.e.
> doing the left RTT pass, then the left main pass, then the right RTT pass,
> then the right main pass as there's no interdependency between the two.
> However, with OSG's current setup, that's not an option, and the RTT pass
> needs to be made stereo-aware and have two copies of everything.
>
There is fallback path in the OSG that allows RTT to be done by writing the
main framebuffer and then copying the result to a texture.  This fallback
scheme only works if you do all the RTT work up front prior to rendering
anything to the final framebuffer.  The proposed change would break this
without any means to work around it at the top level.

If you want to avoid duplicating textures then it would probably be best to
look that OpenGL MultiViewOVR extension.  There is preliminary support for
this the OSG MultiView branch, have a look at the osgmultiviewOVR example.
This provided a scheme for doing stereo in a single cull and draw pass.

Robert.

madsbuvi:
> On Mon, 18 Jan 2021 at 16:05, Chris Djali ***@***.***> wrote: What's happening here is that an RTT needs redoing for each eye. It's better for memory usage etc. to reuse the same texture twice a frame, i.e. doing the left RTT pass, then the left main pass, then the right RTT pass, then the right main pass as there's no interdependency between the two. However, with OSG's current setup, that's not an option, and the RTT pass needs to be made stereo-aware and have two copies of everything.
> There is fallback path in the OSG that allows RTT to be done by writing the main framebuffer and then copying the result to a texture. This fallback scheme only works if you do all the RTT work up front prior to rendering anything to the final framebuffer. The proposed change would break this without any means to work around it at the top level.
Would it be possible to switch based on whether this fallback is active or not?

openscenegraph:
On Mon, 18 Jan 2021 at 16:18, Mads Sandvei <notifications@github.com> wrote:

> On Mon, 18 Jan 2021 at 16:05, Chris Djali *@*.***> wrote: What's
> happening here is that an RTT needs redoing for each eye. It's better for
> memory usage etc. to reuse the same texture twice a frame, i.e. doing the
> left RTT pass, then the left main pass, then the right RTT pass, then the
> right main pass as there's no interdependency between the two. However,
> with OSG's current setup, that's not an option, and the RTT pass needs to
> be made stereo-aware and have two copies of everything.
> There is fallback path in the OSG that allows RTT to be done by writing
> the main framebuffer and then copying the result to a texture. This
> fallback scheme only works if you do all the RTT work up front prior to
> rendering anything to the final framebuffer. The proposed change would
> break this without any means to work around it at the top level.
>
> Would it be possible to switch based on whether this fallback is active or
> not?
>
That would lead to different behaviors for end user applications.  The
whole point of fallback is that you try to get as close to the original
behavior.

AnyOldName3:
To save others some googling, the extension spec is here: https://www.khronos.org/registry/OpenGL/extensions/OVR/OVR_multiview.txt

madsbuvi:
We can't use that extension yet as it requires OpenGL 3 and no FFP, which is not where OpenMW is yet.
Would you accept a PR if i wrote this as a configurable setting in osg::DisplaySetting (or suggest me where i should add it).
That way users would default to the old behaviour while i could switch to the behaviour i need.

openscenegraph:
On Mon, 18 Jan 2021 at 20:25, Mads Sandvei <notifications@github.com> wrote:

> We can't use that extension yet as it requires OpenGL 3 and no FFP, which
> is not where OpenMW is yet.
> Would you accept a PR if i wrote this as a configurable setting in
> osg::DisplaySetting (or suggest me where i should add it).
> That way users would default to the old behaviour while i could switch to
> the behaviour i need.
>
Adding an extra layer of complexity to the interface and implementation as
a workaround for specific applications is not a good engineering solution.
It's a layering hack upon hack.

If the default behavior isn't what you want then you could look at
subclassing from SceneView.

madsbuvi:
All RTTs that depend on the view are broken with the current stereo implementation. It's not application specific. and will require needlessly complicated hacks on the application side.

It does add complexity to your interface but isn't the alternative effectively to ignore a defect in OSG?

openscenegraph:
On Mon, 18 Jan 2021 at 21:21, Mads Sandvei <notifications@github.com> wrote:

> All RTTs that depend on the view are broken with the current stereo
> implementation. It's not application specific. and will require needlessly
> complicated hacks on the application side.
>
> It does add complexity to your interface but isn't the alternative
> effectively to ignore a defect in OSG?
>
I don't do this support for free to put up with BS.

psi29a:
@madsbuvi can you explain what you mean by "All RTTs ... are broken"? I assume that someone, somewhere, makes use of this in OSG.

@robertosfield I think it is frustration about trying to find something that works efficiently that is bubbling up. It is unclear to @madsbuvi and others how the current situation works. 

Do you, or anyone, have links to other applications that uses OSG in this way?

openscenegraph:
On Mon, 18 Jan 2021 at 21:51, Bret Curtis <notifications@github.com> wrote:

> Do you, or anyone, have links to other applications that uses OSG in this
> way?


Lets be clear.  The way you are trying to do things is a hack that works in
a narrow usage case, it's a solution for a general case which is why it
fails when doing stereo.  It will also fail when doing multiple views or
managing multiple graphics contexts.

The correct, and long been the standard, way to handle RTT on a scene graph
that has multiple views or stereo is to buffer.  The OSG itself has a
number of examples of this being done. osgSim::OverlayNode and osgShadow
both buffer using the CullVisitor pointer to differentiate between the
different sets of state and RTT setup.   To start with have a look at
OverlayNode's per View OverlayData:
It's a general solution.


https://github.com/openscenegraph/OpenSceneGraph/blob/df901fba7acff353feab5f766e2d7a914d804799/include/osgSim/OverlayNode#L165

https://github.com/openscenegraph/OpenSceneGraph/blob/df901fba7acff353feab5f766e2d7a914d804799/src/osgSim/OverlayNode.cpp#L948

When I say long been the standard way of handling this issue, I wrote
OverlayNode back in 2005.  Works multi-threaded, handling multiple views,
handles multiple graphics contexts, handles stereo or more exotic VR setups.

And if you're worried about the inefficiency of not reusing textures, have
about graphics card memory since 2005... multi-buffering wasn't a problem
back in 2005, while today when we have multi-gigabyte graphics cards.

So please just fix your application and stop wasting my time.

Issue #1017, opened 2021-01-04
Duplicate symbol FileExists when linking OSG as a static library


ankurverma85:
osgPlugins/zip/unzip.cpp includes a function "FileExists" in the global namespace which is a very common function name.
When linking this in static-lib mode on linux, it conflicts with gdal which has the same function name in the global namespace and is a dependency.

The conflict also happens for inflate_copyright variable

I believe the convention here is to mark all local functions or variables as "static" so they do not collide with other functions of similiar name.

More details on how vcpkg incorporates a fix are here:
https://github.com/microsoft/vcpkg/pull/15374/files#

JackBoosY:
Hi there, any official reply?

openscenegraph:
On Mon, 11 Jan 2021 at 02:44, Jack·Boos·Yu <notifications@github.com> wrote:

> Hi there, any official reply?
>
If there are PR's to the OSG itself then I can review and merge changes.

ankurverma85:
@robertosfield, Can you please review the PR ?

openscenegraph:
On Fri, 15 Jan 2021 at 09:43, ankurverma85 <notifications@github.com> wrote:

> @robertosfield <https://github.com/robertosfield>, Can you please review
> the PR ?
>
I will do a review of all the PR's when I get a clear block of time.

PR #1023, opened 2021-01-13
Fix Duplicate symbols when linking OSG as a static library

ankurverma85:
Fixes bug #1017


robertosfield:
Thanks merged with master and 3.6 branch

PR #1024, opened 2021-01-14
daeReader::convert( std::istream& ) fixes

### LINKED ISSUE #994: daePlugin convert istringstream occasionally fails

glebm:
1. Pass a \0-terminated C-string to `_dae->openFromMemory` (fixes #994)
2. Use a heap array instead of a vector. This avoids unnecessary filling of the buffer with zeroes (which vector always does).
3. Correctly get the input stream size by subtracting the position of the beginning of the stream.
4. Handle stream error.

Please cherry-pick this into 3.6 when merging

### DIFF
--- a/src/osgPlugins/dae/daeReader.cpp
+++ b/src/osgPlugins/dae/daeReader.cpp
@@ -12,6 +12,9 @@
  */
 
 #include "daeReader.h"
+
+#include <cstddef>
+
 #include <dae.h>
 #include <dae/domAny.h>
 #include <dom/domCOLLADA.h>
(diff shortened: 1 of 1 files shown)


unelsson:
Great job! I tested this with osgviewer and OpenMW engine using various dae-models. Hoping this to get merged, it's absolutely crucial to enable loading models from memory.

glebm:
Thanks for testing!

psi29a:
Cool! Thanks.

psi29a:
@glebm can you retarget this for 3.6 instead of master, or create a 2nd one specifically for 3.6 ?

glebm:
Can just ask the maintainer to cherry-pick it for 3.6 as well at merge time

Confirmed that it cherry picks without conflicts

psi29a:
@robertosfield any objections?

src/osgPlugins/dae/daeReader.cpp
```diff
@@ -12,6 +12,9 @@
  */
 
 #include "daeReader.h"
+
+#include <cstddef>
```
robertosfield, line 16: Why add this header?

I can't see anything in the new code that would require it.  I don't know whether it's inclusion would cause problems with very old compilers so unless it's really needed I wouldn't add it.

glebm: It's for `std::size_t`, I think this header is very old, certainly in C++98

robertosfield:
On a second review of the changes I've decided that having two explicitly delete [] is not a good solution as code changes could easily miss this or if an exception happens it'll leak. 

I have applied the following commit retaining the use of std::vector<>.

   https://github.com/openscenegraph/OpenSceneGraph/commit/295bfffc17c8f25ab65ec8151f8c2ef8c97388b6

I have merged this commit with the 3.6 branch and master. Could you test?

glebm:
Yeah, in C++20 we'd just use:

```c++
auto buffer = std::make_unique_for_overwrite<char[]>(length + 1);
```

Or in C++11:

```c++
auto buffer = std::unique_ptr<T>(new char[length + 1]);
```

Your solution looks good, even if it may be a bit slower.

src/osgPlugins/dae/daeReader.cpp
psi29a: Which is preferable then guessing over uL and uLL depending on platform and architecture.

PR #1027, opened 2021-01-14
Fix osgViewer `USE_GRAPHICSWINDOW` without a windowing system

glebm:
There is no way for the user to know whether OSG was compiled with a
windowing system.

This means calling USE_GRAPHICSWINDOW() in client libraries was unsafe
in the case of the "None" windowing system, as it would fail to link.

Turns the call into a no-op in that case.

Please also cherry-pick this fix into 3.6 when merging.

/cc @psi29a


robertosfield:
Thanks merged with master and 3.6 branch.

PR #1031, opened 2021-01-19
VertexAttribDivisor functionality in Array

valid-ptr:
There was no possibilities to use VertexAttribDivisor with VertexArrayObject(ed) geometry.
This PR gives such possibility.
You can still use osg::VertexAttribDivisor for not VAOed geometries.
Array->setDivisor can be used for any configuration.

I've compiled OSG with/without `OSG_GL_VERTEX_ARRAY_FUNCS_AVAILABLE`
I've tested it mostly for OpenGL Core profile (with/without VAO + with/widthout VBO)

### DIFF
--- a/include/osg/Array
+++ b/include/osg/Array
@@ -140,6 +140,7 @@ class OSG_EXPORT Array : public BufferData
             _dataSize(dataSize),
             _dataType(dataType),
             _binding(binding),
+            _divisor(0),
             _normalize(false),
             _preserveDataType(false) {}
 
@@ -149,6 +150,7 @@ class OSG_EXPORT Array : public BufferData
             _dataSize(array._dataSize),
             _dataType(array._dataType),
             _binding(array._binding),
+            _divisor(array._divisor),
             _normalize(array._normalize),
             _preserveDataType(array._preserveDataType) {}
 
@@ -191,6 +193,11 @@ class OSG_EXPORT Array : public BufferData
         /** Get how this array should be passed to OpenGL.*/
         Binding getBinding() const { return _binding; }
 
+        /** Specify the rate at which generic vertex attributes advance during instanced rendering.*/
+        void setDivisor(GLuint divisor) { _divisor = divisor; }
+
+        /** Get the rate at which generic vertex attributes advance during instanced rendering.*/
+        GLuint getDivisor() const { return _divisor; }
 
         /** Specify whether the array data should be normalized by OpenGL.*/
         void setNormalize(bool normalize) { _normalize = normalize; }
@@ -226,6 +233,7 @@ class OSG_EXPORT Array : public BufferData
         GLint   _dataSize;
         GLenum  _dataType;
         Binding _binding;
+        GLuint  _divisor;
         bool    _normalize;
         bool    _preserveDataType;
 };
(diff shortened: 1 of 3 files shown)


robertosfield:
It's a CPU overhead horror story that ALL users would have to pay for even
though the vast majority of users won't need it.  It's an inner loop
issue.  It's crazy to go worrying about trying to remove a reducent
glViewport call done once or twice a frame when you add a whole bunch of if
statements and GL calls per vertex array.  It's I'll give you a cent, but
then go take a  thousand dollars.

This is for sure it's the way to introduce it into the OSG.  We can't go
adding more CPU overhead just to expose all possible OpenGL features.  The
OSG/OpenGL are already slowed down massively by CPU overhead.

It might be better to just remove vertex divisor support completely.

valid-ptr:
> It's crazy to go worrying about trying to remove a reducent
glViewport call done once or twice a frame when you add a whole bunch of if
statements and GL calls per vertex array.

Robert, I am disappointed with your answer at least ... It's different stories!

> It's an inner loop issue.

If you are talking about "draw" (in stats) then it will be just the same for _all_ users I think:
1) initial value of **divisor** == 0 everywhere, so the  `if (currentDivisor == newDivisor)` statement will be `true` in most cases.
2) `if (currentDivisor == newDivisor)` statement is not expensive compared to any "gl*" (host -> gpu) command in most cases :)
3) For any VAO(ed) geometry `callVertexAttribDivisor` function must be processed only once per array while VAO compilation.

> It might be better to just remove vertex divisor support completely.

Don't do this :^]
Do you have other ideas of implementation (VAO compat)?

About **glViewport(0, 0, 800, 600)** in the **gltrace** - it's not a question of the performance.
1) It's a story about clear OpenGL trace first of all. When you see the trace with 800x600 viewport you can be disappointed (like i was).
2) Imagine, you have 10 osg::Camera(s) with the same viewport, you will have 20 `glViewport` commands without #1002 PR; with  #1002 PR - just one `glViewport` command. So performance can be a reason too for some applications. What about bus bandwidth!?

Good night))

openscenegraph:
There is a reason why the VSG is 10 x faster the OSG in traversals.  It's
because I minimize memory bandwidth, avoids cache misses and branching.,
something the OSG with all it's callbacks and memory bloat does badly at
This PR distills all the poor design/implementation weaknesses of the OSG
is one neat package.  It bad for performance even when the divisor is 0. It
bloats memory and adds branching where there was none before.

As for a solution for this seroos regression for performance.  I don't
know.  If you have VAO enabled then perhaps you can avoid some of the call
overhead.  Perhaps if VAO is not used then you just disable the divisor
completely and don't attempt it.

There really is only so much potential you can squeeze out of the
OSG/OpenGL.  My time if far better used getting VSG to 1.0 so users can
move on rather trying to flog an old horse.

valid-ptr:
I've tested the scene with **100000** texturedQuadGeometries with/without this PR on
` x86_64 Intel(R) Core(TM) i7 CPU  860  @ 2.80GHz GenuineIntel GNU/Linux`  + `NVIDIA Geforce 1050 TI on PCI Express 2`
with disabled VAO, enabled VBO

There is no difference in performace at all.
Same FPS (about 3.8), same draw (139ms), all the same

> As for a solution for this seroos regression for performance.  I don't
know.  If you have VAO enabled then perhaps you can avoid some of the call
overhead.  Perhaps if VAO is not used then you just disable the divisor
completely and don't attempt it.

Bad news.
If will be one more statement somewhere or it will work only in GL3 CORE profile where VAO is required, no GLES etc.

robertosfield:
I'm closing this PR as 3.6 is stable branch, experimental ABI changing functionality doesn't below in this branch.

PR #1033, opened 2021-01-19
3.6: Fix graph_array compilation on modern C++ compilers

glebm:
`std::mem_fun_ref` has been completely removed in C++17 and is not
available in certain compilers, such as Android NDK clang, even when
targeting C++98.

### DIFF
--- a/src/osgUtil/tristripper/include/detail/graph_array.h
+++ b/src/osgUtil/tristripper/include/detail/graph_array.h
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2004 Tanguy Fautré.
+// Copyright (C) 2004 Tanguy FautrÃ©.
 // For conditions of distribution and use,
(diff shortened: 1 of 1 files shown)


robertosfield:
Duplicating a code block for different compiler versions just is just not the way to solve problems like this, it's bad programming that can easily to compile and runtime errors creeping bit by bit into the software.

One code path for all compilers is the gold standard we should working towards.  If new compilers don't compile old code then ammend that old code so it will work on both compiler sets.

robertosfield:
This is what I've applied, the key is we retain a single code path that all uses will compile, run and test the same code path:

    https://github.com/openscenegraph/OpenSceneGraph/commit/8a0114a46a4bad9041297950fe3bfbb2aea6e1da

glebm:
Great, thanks!

PR #1035, opened 2021-01-21
Convert osg::Geode to osg::Group in dae-plugin

unelsson:
This is more of a suggestion/idea than direct pull request, and may require further work. I hope for feedback, and suggest reviewing the code before merging, and considering whether this kind of change is helpful.

What it does: Converts osg::Geode, used by the Collada plugin, to osg::Group. I've understood that Geode is deprecated, so this should be a step forward. It does fix a few Collada-related problems with project OpenMW, particulary the ability to pick objects with mouse click (collision related?), and I haven't at least yet encountered any regressions.

However, I have limited knowledge of how the scenegraph should actually be handled, and if this is a decent way to do it.

### DIFF
--- a/src/osgPlugins/dae/daeRGeometry.cpp
+++ b/src/osgPlugins/dae/daeRGeometry.cpp
@@ -31,58 +31,50 @@
 
 using namespace osgDAE;
 
-osg::Geode* daeReader::getOrCreateGeometry(domGeometry *pDomGeometry, domBind_material* pDomBindMaterial, const osg::Geode** ppOriginalGeode)
+osg::Group* daeReader::getOrCreateGeometry(domGeometry *pDomGeometry, domBind_material* pDomBindMaterial, const osg::Group** ppOriginalGeometryGroup)
 {
     // Check cache if geometry already exists
-    osg::Geode* pOsgGeode;
+    osg::Group* pOsgGeometryGroup;
 
-    domGeometryGeodeMap::iterator iter = _geometryMap.find( pDomGeometry );
-    if ( iter != _geometryMap.end() )
+    domGeometryGroupMap::iterator iter = _geometryGroupMap.find( pDomGeometry );
+    if ( iter != _geometryGroupMap.end() )
     {
-        pOsgGeode = iter->second.get();
+        pOsgGeometryGroup = iter->second.get();
     }
     else
     {
-        pOsgGeode = processGeometry( pDomGeometry );
-        _geometryMap.insert( std::make_pair( pDomGeometry, pOsgGeode ) );
+        pOsgGeometryGroup = processGeometryGroup( pDomGeometry );
+        _geometryGroupMap.insert( std::make_pair( pDomGeometry, pOsgGeometryGroup ) );
     }
 
-    if (ppOriginalGeode)
+    if (ppOriginalGeometryGroup)
     {
-        *ppOriginalGeode = pOsgGeode;
+        *ppOriginalGeometryGroup = pOsgGeometryGroup;
     }
 
-    if (!pOsgGeode)
-        return NULL;
-
-    // Create a copy of the cached Geode with a copy of the drawables,
+    // Create a copy of the cached Group with a copy of the drawables,
     // because we may be using a different material or texture unit bindings.
-    osg::Geode *pCopiedOsgGeode = static_cast<osg::Geode*>(pOsgGeode->clone(osg::CopyOp::DEEP_COPY_DRAWABLES));
-    if ( pCopiedOsgGeode == NULL )
+    osg::Group *pCopiedOsgGeometryGroup = static_cast<osg::Group*>(pOsgGeometryGroup->clone(osg::CopyOp::DEEP_COPY_DRAWABLES));
+    if ( pCopiedOsgGeometryGroup == NULL )
     {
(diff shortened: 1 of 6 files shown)


src/osgPlugins/dae/daeRGeometry.cpp
```diff
@@ -31,58 +31,50 @@
...
         OSG_WARN << "Failed to load geometry " << pDomGeometry->getName() << std::endl;
         return NULL;
     }
 
-    // Compute optimized geometry by expanding all indexed arrays so we are no longer rendering with the slow path
-    for(unsigned int i=0;i < pCopiedOsgGeode->getNumDrawables();++i)
+    for(unsigned int i=0;i < pOsgGeometryGroup->getNumChildren();++i)
     {
-        osg::Geometry* geom = pCopiedOsgGeode->getDrawable(i)->asGeometry();
-        if (geom)
```
unelsson, line 71: Is the removal of this null check safe here?

src/osgPlugins/dae/daeRGeometry.cpp
```diff
@@ -160,18 +152,15 @@ osg::Node* daeReader::processMorph(domMorph* pDomMorph, domBind_material* pDomBi
...
-    osg::Geometry* pOsgGeometry = dynamic_cast<osg::Geometry*>(pOsgGeode->getDrawable(0));
-    if (!pOsgGeometry)
-        return NULL;
+    // Expects a single geometry inside the group, should change this
+    osg::Geometry* pOsgGeometry = dynamic_cast<osg::Geometry*>(pOsgGeometryGroup->getChild(0));
 
     osgAnimation::MorphGeometry* pOsgMorphGeometry = new osgAnimation::MorphGeometry(*pOsgGeometry);
-    pOsgGeode->removeDrawables(0);
-    pOsgGeode->addDrawable(pOsgMorphGeometry);
+    pOsgGeometry = pOsgMorphGeometry;
```
unelsson, line 163: Is this method safe? Is there a chance of memory leak? This creates osgAnimation::MorphGeometry(*pOsgGeometry), then changes the pOsgGeometry pointer to point to the new object. I haven't checked what osgAnimation::MorphGeometry constructor actually does.

src/osgPlugins/dae/daeRGeometry.cpp
```diff
@@ -456,10 +440,10 @@ osg::Geode *daeReader::processGeometry(domGeometry *pDomGeometry)
 
 
 template< typename T >
-void daeReader::processSinglePPrimitive(osg::Geode* geode,
+void daeReader::processSinglePPrimitive(osg::Group* geometryGroup,
     const domMesh* pDomMesh, const T* group, SourceMap& sources, GLenum mode)
 {
-    osg::ref_ptr<osg::Geometry> geometry = new osg::Geometry();
```
unelsson, line 462: This is the only case where osg::ref_ptr<osg::Geometry> is used, other templates use just pointers. Was there a specific reason for this?

src/osgPlugins/dae/daeRSceneObjects.cpp
```diff
@@ -18,7 +18,6 @@
 
 #include <osg/Switch>
 #include <osg/LightSource>
-#include <osg/Geode>
```
unelsson, line 21: I didn't find this used anywhere.

psi29a:
I should also mention, that some of these collada issues came up when doing imports/exports with Godot's better collada blender plugin. Seeing that the dae osgPlugin stumble made us look at the plugin more carefully.

src/osgPlugins/dae/daeRGeometry.cpp
```diff
@@ -470,22 +454,22 @@ void daeReader::processSinglePPrimitive(osg::Geode* geode,
     domP_Array domPArray;
     domPArray.append(group->getP());
     std::vector<std::vector<GLuint> > indexLists;
-    resolveMeshArrays(domPArray, group->getInput_array(), pDomMesh, geometry.get(), sources, indexLists);
+    resolveMeshArrays(domPArray, group->getInput_array(), pDomMesh, geometry, sources, indexLists);
     if (!indexLists.front().empty())
```
glebm, line 458: `geometry` leaks if this is false

Use `ref_ptr` like before but `geometry.release()` instead of `geometry.get()`?

unelsson: "If this is false" - what do you mean? resolveMeshArrays is of type `void`, and above, `osg::Geometry* geometry = new osg::Geometry;` shouldn't return a nullptr. Maybe I'm not noticing something here, but please explain. One point of view is that if someone changes the code, perhaps ref_ptr can be safer?

glebm: if the code doesn't go into the if branch, the ownership of `geometry` does not get transferred into `geometryGroup`, so it never gets deleted.

As I understand, the previous version was buggy in another way - `geometry` was deleted immediately after the function finished.

Or perhaps `geometryGroup->addChild` does not transfer ownership, in which case the previous code was correct.

psi29a:
@robertosfield how do you feel about this?

psi29a:
@openscenegraph We've have had this being used widespread and the patch is already being used in OpenMW's PPA, Ubuntu and Debian latest releases on top the official 3.6.5 release; can this be included officially upstream?

Issue #1007, opened 2020-12-04
Rendering errors with collada using Qt's QOpenGLWidget and QGLWidget


unelsson:
I also made a bug report to Qt5 - https://bugreports.qt.io/browse/QTBUG-89140 and uploaded a few screenshots there to show the problem.

When Qt5/QOpenGLWidget is integrated with OpenSceneGraph, some models load in a wrong fashion. Screenshots are attached. OpenSceneGraph alone renders the same models correctly. This may be a Qt5 bug, but I'm reporting it here, because it may be related to OSG, perhaps OSG's dae-plugin setting some variable in a wrong fashion that is not compatible with Qt4 and Qt5.

This was tested using a modified version of osgQt ( https://github.com/openscenegraph/osgQt ), which uses QOpenGLWidget along with osgViewer::CompositeViewer. The same bug however occurs with Qt4 and Qt5 using QGLWidget as well (which is clearly visible in OpenMW-CS, free software using OSG, osgQt and Qt4/Qt5).

The problem exists, when loading dae/collada models. When tested with fbx-model and osgt-model, the rendering is correct (although the model itself has a minor problem with non-applied mirror modifier).

An example code along with modifier osgQt for CompositeViewer is at https://gitlab.com/unelsson/osgqt_tests/

This line (and the ones next to it) are for testing models in different formats: https://gitlab.com/unelsson/osgqt_tests/-/blob/master/osgViewerQt_CompositeViewer.cpp#L157

Sidenote: A related issue at osgQt is https://github.com/openscenegraph/osgQt/issues/37 . Note thought that the bug is confirmed to be present also in the old osgQt::GraphicsWindowQt, using CompositeViewer and QGLWidget with Qt4. This is confirmed in project OpenMW-CS.

unelsson:
This OpenMW-issue shows screenshots in OpenMW/OpenSceneGraph and in OpenMW-CS/OSG via OsgQt. All three collada models are corrupted with osgQt.
https://gitlab.com/OpenMW/openmw/-/issues/5713

unelsson:
While investigating this, I converted dae-plugin's osg::Geode to osg::Group https://github.com/unelsson/OpenSceneGraph/tree/daeplugingeodetogroup . It's based on OpenMW's fork, so it won't merge directly, but in case this is needed, I can make another PR later.

The original issue still exists. It's seems to be specific to dae-plugin with Qt.

robertosfield:
[...]

PR #1039, opened 2021-01-23
Use c-locale with std::istream

unelsson:
Related to https://github.com/openscenegraph/OpenSceneGraph/issues/1007

Tested to work with Qt5.


unelsson:
I haven't tested the fout for writing, and unfortunately can't test that, but should I add that to this PR too?

robertosfield:

With a brief look I couldn't see any writeNode to an ostream, the write itself looks to be with the COLLADA dom.  If there needs to be a fix there then I'll leave that to others who actually require the functionality if any do,

PR #1044, opened 2021-02-01
nodePath contains bad pointer, since pagedlod might be deleted during the intersection

zengjinchong:
The node path contians pagedlod pointer, if the pagedlod pointer is released during the intersection. That might happen if the read   callback is set to the intersection visitor. For example, databaseCacheReadCallback is set to the intersection visitor. The databaseCacheReadCallback will release the pagedlod when the cache is full.

### DIFF
--- a/include/osgUtil/LineSegmentIntersector
+++ b/include/osgUtil/LineSegmentIntersector
@@ -15,6 +15,7 @@
 #define OSGUTIL_LINESEGMENTINTERSECTOR 1
 
 #include <osgUtil/IntersectionVisitor>
+#include <osg/ObserverNodePath>
 
 namespace osgUtil
 {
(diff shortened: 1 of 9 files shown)


robertosfield:
An observer pointer isn't a cheap item to add, it's not something that should be done a traversals that are meant to be efficient. This change would force that overhead for everyone doing intersection traversals just to catch your specific "odd" usage.

I don't know why you are doing what you are doing, nor whether its at all appropriate or the most efficient way to do it.  What I can say is very likely there will be a better way of doing what you want to do robustly without modifying the core OSG so I'm going to close PR it's cost/benefit with the approach is awful - first you have to explore other means of doing your task robustly without hacking at the OSG itself.

PR #1045, opened 2021-02-02
las color might be between 0 and 255, so the calculation of the color will be 0

zengjinchong:
las color might be between 0 and 255, so the calculation of the color will be 0. 
I fixed the bug by adding the judgement that If the color is less than 255, then the calculation will not be done.
Hope you can adopt my suggestion

### DIFF
--- a/src/osgPlugins/las/ReaderWriterLAS.cpp
+++ b/src/osgPlugins/las/ReaderWriterLAS.cpp
@@ -137,9 +137,18 @@ class ReaderWriterLAS : public osgDB::ReaderWriter
 
                 // Extract color components from LAS point
                 liblas::Color c = p.GetColor();
-                uint32_t r = c.GetRed() >> 8;
-                uint32_t g = c.GetGreen() >> 8;
-                uint32_t b = c.GetBlue() >> 8;
+
+				uint32_t r = c.GetRed();
+				uint32_t g = c.GetGreen();
+				uint32_t b = c.GetBlue();
+
+				if (r > 255 || g > 255 || b > 255)
+				{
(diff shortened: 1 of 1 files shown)


robertosfield:
This introduces a serious bug so I'm closing this PR.  The process of the colour data will be completely spurious, if it's a red is value of 255 then it'll have a value of 255, but if it's 256 it'll have a value of 1.  I can't imagine this is what you intended, or what anyone else would appreciate.

My best guess given a buggy PR and odd explanation that gives no real context to the issue you are seeing, is that you are see a precision issue with the OSG LAS plugin mapping the colour range to 8 bit.  Would a 16bit colour be more appropriate?  Or just use a Vec4Array rather than a Vec4ubArray.

zengjinchong:
the problem i want to solve is that  las color might have two range of representing color. One is 0-255 and another is 0-65536. In current case, we only use 0-65536. If the color range is 0-255, then the color will be zero. the point cloud is black.

openscenegraph:
On Tue, 2 Feb 2021 at 09:02, zengjinchong <notifications@github.com> wrote:

> the problem i want to solve is that las color might have two range of
> representing color. One is 0-255 and another is 0-65536. In current case,
> we only use 0-65536. If the color range is 0-255, then the color will be
> zero. the point cloud is black.
>
This is a data source issue.  Check if LAS itself provides any guidance on
what the colour range interpretation should be.  If there is no official
guidance on the colour range and it becomes an ad-hoc different users data
assumes different ranges then on osgDB::Option setting would need to be
used.  None of this changes how unsuitable this PR is.

zengjinchong:
Ok, i see

PR #1049, opened 2021-02-20
Do not delete GlobalReferencedMutex on exit

### LINKED ISSUE #1048: heap-use-after-free on exit in OpenThreads/ScopedLock:54
heap-use-after-free happens on exit causing a crash.
It seems to be caused by the global mutex being destroyed before some other things.

Here is the salient part of ASAN report:

```
==81666==ERROR: AddressSanitizer: heap-use-after-free on address 0x603000018fd0 at pc 0x5580cd51d3f1 bp 0x7ffea172dc00 sp 0x7ffea172dbf0
READ of size 8 at 0x603000018fd0 thread T0
    #0 0x5580cd51d3f0 in OpenThreads::ScopedPointerLock<OpenThreads::Mutex>::ScopedPointerLock(OpenThreads::Mutex*) extern/fetched/osg/include/OpenThreads/ScopedLock:54
    #1 0x5580cd51d3f0 in osg::StateAttribute::removeParent(osg::StateSet*) extern/fetched/osg/src/osg/StateAttribute.cpp:38
    #2 0x5580cd6fb344 in osg::StateSet::clear() extern/fetched/osg/src/osg/StateSet.cpp:734
    #3 0x5580cd700eb6 in osg::StateSet::~StateSet() extern/fetched/osg/src/osg/StateSet.cpp:285
    #4 0x5580cd703145 in osg::StateSet::~StateSet() extern/fetched/osg/src/osg/StateSet.cpp:286
    #5 0x5580d2c53f2e in osg::Referenced::unref() const extern/fetched/osg/include/osg/Referenced:201
    #6 0x5580d2c53f2e in osg::Referenced::unref() const extern/fetched/osg/include/osg/Referenced:178
    #7 0x5580d2c53f2e in osg::ref_ptr<osg::StateSet>::~ref_ptr() extern/fetched/osg/include
    #8 0x5580d2c53f2e in void std::_Destroy<osg::ref_ptr<osg::StateSet> >(osg::ref_ptr<osg::StateSet>*) /usr/include/c++/10/bits/stl_construct.h:140
    #9 0x5580d2c53f2e in void std::_Destroy_aux<false>::__destroy<osg::ref_ptr<osg::StateSet>*>(osg::ref_ptr<osg::StateSet>*, osg::ref_ptr<osg::StateSet>*) /usr/include/c++/10/bits/stl_construct.h:152
    #10 0x5580d2c53f2e in void std::_Destroy<osg::ref_ptr<osg::StateSet>*>(osg::ref_ptr<osg::StateSet>*, osg::ref_ptr<osg::StateSet>*) /usr/include/c++/10/bits/stl_construct.h:185
    #11 0x5580d2c53f2e in void std::_Destroy<osg::ref_ptr<osg::StateSet>*, osg::ref_ptr<osg::StateSet> >(osg::ref_ptr<osg::StateSet>*, osg::ref_ptr<osg::StateSet>*, std::allocator<osg::ref_ptr<osg::StateSet> >&) /usr/include/c++/10/bits/alloc_traits.h:738
    #12 0x5580d2c53f2e in std::vector<osg::ref_ptr<osg::StateSet>, std::allocator<osg::ref_ptr<osg::StateSet> > >::~vector() /usr/include/c++/10/bits/stl_vector.h:680
    #13 0x5580d2c53f2e in osgText::Font::~Font() extern/fetched/osg/src/osgText/Font.cpp:295
    #14 0x5580d2e1c318 in osgText::DefaultFont::~DefaultFont() extern/fetched/osg/src/osgText/DefaultFont.cpp:35
    #15 0x5580d2e1c318 in osgText::DefaultFont::~DefaultFont() extern/fetched/osg/src/osgText/DefaultFont.cpp:37

0x603000018fd0 is located 0 bytes inside of 24-byte region [0x603000018fd0,0x603000018fe8)
freed by thread T0 here:
    #0 0x7f04b51dff7f in operator delete(void*, unsigned long) (/lib/x86_64-linux-gnu/libasan.so.6+0xb2f7f)
    #1 0x5580cd372627 in osg::ResetPointer<OpenThreads::Mutex>::~ResetPointer() extern/fetched/osg/src/osg/Referenced.cpp:47
    #2 0x7f04b1fd2a56  (/lib/x86_64-linux-gnu/libc.so.6+0x44a56)

previously allocated by thread T0 here:
    #0 0x7f04b51def17 in operator new(unsigned long) (/lib/x86_64-linux-gnu/libasan.so.6+0xb1f17)
    #1 0x5580cd370e71 in osg::Referenced::getGlobalReferencedMutex() extern/fetched/osg/src/osg/Referenced.cpp:86
    #2 0x5580c6bb8c58 in osg::InitGlobalMutexes::InitGlobalMutexes() extern/fetched/osg/src/osg/Referenced.cpp:95
    #3 0x5580c6bb8c58 in __static_initialization_and_destruction_0 extern/fetched/osg/src/osg/Referenced.cpp:98
    #4 0x5580c6bb8c58 in _GLOBAL__sub_I__ZN3osg10Referenced24getGlobalReferencedMutexEv extern/fetched/osg/src/osg/Referenced.cpp:318
    #5 0x5580d39029fc in __libc_csu_init (/home/gleb/repos/glebm/openmw/build/openmw+0x25b039fc)

SUMMARY: AddressSanitizer: heap-use-after-free extern/fetched/osg/include/OpenThreads/ScopedLock:54 in OpenThreads::ScopedPointerLock<OpenThreads::Mutex>::ScopedPointerLock(OpenThreads::Mutex*)
```

Full ASAN report: https://gist.githubusercontent.com/glebm/1c255220ebe6e073ee2d0a763f465592/raw/ba85dd16b200f3dfebc1fe5d44944a27c84a05e2/osg-asan.txt

glebm:
Static (de)initialization order caused this mutex to be deleted before
the objects that depended on it, causing a crash on exit.

With this change, OSG no longer deletes the mutex on exit. This is OK
because it is deleted by the OS on exit anyway.

Fixes #1048

---

Please also cherry-pick into 3.6 (applies cleanly)

### DIFF
--- a/src/osg/Referenced.cpp
+++ b/src/osg/Referenced.cpp
@@ -79,12 +79,11 @@ struct ResetPointer
 };
 
 typedef ResetPointer<DeleteHandler> DeleteHandlerPointer;
-typedef ResetPointer<OpenThreads::Mutex> GlobalMutexPointer;
 
 OpenThreads::Mutex* Referenced::getGlobalReferencedMutex()
 {
-    static GlobalMutexPointer s_ReferencedGlobalMutext = new OpenThreads::Mutex;
-    return s_ReferencedGlobalMutext.get();
+    static OpenThreads::Mutex* s_ReferencedGlobalMutext = new OpenThreads::Mutex;
+    return s_ReferencedGlobalMutext;
 }
 
 // helper class for forcing the global mutex to be constructed when the library is loaded.


glebm:
Note that e.g. the Google style guide only allows static objects with trivial destructors for this reason.

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

robertosfield:
Just leaking stuff because the OS will clean it up anyway.... geeze why don't we just remove all destructors?  The OS will clean up anyway right...

That's not a good solution.

glebm:
> geeze why don't we just remove all destructors

This is exactly what the Google style guide suggests (and how we actually write code at Google)

Why do you want to manually destroy static objects?

robertosfield:
Static objects can have resources associated with them, letting destructors be called to clean them up is a good practice.  If we don't well end up with various debugging tools shouting at users for leaks.

Try moving the static declaration outside the GlobalReferencedMutex() method.

glebm:
From http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1247r0.html#intro:

> The destruction of variables with static or thread local storage duration can introduce bugs and surprising behavior that is hard to anticipate, and isn’t fully supported across all implementations of C++.

> Try moving the static declaration outside the GlobalReferencedMutex() method:

While this may fix the issue in a particular compiler/linker/codebase combination, it will just cause the initialization order to be undefined, as translation unit order in the binary is implementation-defined.

openscenegraph:
On Sat, 20 Feb 2021 at 12:52, Gleb Mazovetskiy <notifications@github.com>
wrote:

> From
> http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1247r0.html#intro
> :
>
> "The destruction of variables with static or thread local storage duration
> can introduce bugs and surprising behavior that is hard to anticipate, and
> isn’t fully supported across all implementations of C++."
>
> Try moving the static declaration outside the GlobalReferencedMutex()
> method:
>
> While this may fix the issue in a particular compiler/linker/codebase
> combination, it will just cause the initialization order to be undefined,
> as translation unit order in the binary is implementation-defined.
>
The method has existed for a very long time, and used successfully across
many platforms across thousands of applications.  Yes static objects are a
problem, in hindsight I'd avoid them far more, but this very old code,
changing as few things as possible is the sensible thing to do.

Your particular application usage approach has an issue with code that has
worked fine for decades, then fine, try what I suggest, if it works great.

For stability of the OSG is crucial that we try to minimize changes,
especially on stable branches.

glebm:
I think I've presented sufficient arguments for why statics should not be destroyed?

This only affects behaviour at exit, by fixing a crash.
The crash at exit is very annoying when using tools such as sanitizers, perf, etc.

In terms of stability, surely, Google's and Apple's code runs on even more platforms than OSG

openscenegraph:
On Sat, 20 Feb 2021 at 13:35, Gleb Mazovetskiy <notifications@github.com>
wrote:

> I think I've presented sufficient arguments for why statics should not be
> destroyed?
>
> This only affects behaviour at exit, by fixing a crash.
> The crash at exit is very annoying when using tools such as sanitizers,
> perf, etc.
>
> In terms of stability, surely, Google's and Apple's code runs on even more
> platforms than OSG
>
Wow.  Have you ever looked at the platforms the OSG supports?

The OSG existed before Android, macOS and iOS.  It's been compiling on many
platforms that neither Google or Apple have ever built for.

I've spent decades trying to keep the OSG running on all the various
platforms and all the various usage cases, compiler.  It's that experience
is why I push back on dodgy looking submissions.  When I don't issue, come
back and bite the community.

glebm:
> Wow.  Have you ever looked at the platforms the OSG supports?

I have not, sounds like I made a wrong assumption there.

I'm surprised this PR is causing any controversy. The language proposal I linked above even considers disabling static destructors by default in future C++ versions (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1247r0.html#intro).

openscenegraph:
On Sat, 20 Feb 2021 at 15:15, Gleb Mazovetskiy <notifications@github.com>
wrote:

> Wow. Have you ever looked at the platforms the OSG supports?
>
> I have not, sounds like I made a wrong assumption there.
>
> I'm surprised this PR is causing any controversy. The language proposal I
> linked above even considers disabling static destructors by default in
> future C++ versions (
> http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1247r0.html#intro
> ).
>
The OSG is a super mature project. It's over two decades old.  Just letting
stuff leak is against two decades of experience with the project, I'm not
about to go sanctioning a policy of hacking code to quieten santiziers by
leaking stuff.

There has to be a better solution than the one you have come up with.  It
might be good enough for you, but not for me.

If you can't accept then please go find another project lead to harass,
you're wasting my time.

glebm:
I absolutely did not want you to feel harassed. This open-source project should bring you joy, and I'm sorry that's how my replies came across. :peace_symbol: :heart:

PR #1055, opened 2021-03-16
Fix C++17 MSVC compilation error

glebm:
With C++17, Windows headers must not be included after `using namespace std;`.

Windows headers define a `byte` type internally and `using namespace std`
causes it to conflict with `std::byte`:

    error C2872: 'byte': ambiguous symbol

MSVC thread: https://developercommunity.visualstudio.com/t/error-c2872-byte-ambiguous-symbol/93889

### DIFF
--- a/src/osg/DisplaySettings.cpp
+++ b/src/osg/DisplaySettings.cpp
@@ -22,15 +22,15 @@
 #include <algorithm>
 #include <string.h>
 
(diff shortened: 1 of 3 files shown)


Smitter:
The problem happens when this:

using namespace osg;
using namespace std;

occures before this:

#include<windows.h>

PR #1056, opened 2021-03-16
Check GL_NV_framebuffer_multisample_coverage

### LINKED ISSUE #1028: glRenderbufferStorageMultisampleCoverageNV used on Mesa, despite not having NV_framebuffer_multisample_coverage
`glRenderbufferStorageMultisampleCoverageNV` is provided by the [`NV_framebuffer_multisample_coverage`](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_framebuffer_multisample_coverage.txt) extension for implementing what Nvidia call [Coverage Sampled Antialiasing](https://developer.download.nvidia.com/SDK/9.5/Samples/DEMOS/Direct3D9/src/CSAATutorial/docs/CSAATutorial.pdf). It's used in OSG by [`osg::RenderBuffer::getObjectId`](https://github.com/openscenegraph/OpenSceneGraph/tree/40bddd8d9532620dc704adce0d2fb0c854d62d0f/src/osg/FrameBufferObject.cpp#L138-L145) following a check to `osg::GLExtensions::isRenderbufferMultisampleCoverageSupported()` in favour of the non-vendor-specific `glRenderbufferStorageMultisample`.

`osg::GLExtensions::isRenderbufferMultisampleCoverageSupported()` just checks whether or not the function pointer is null, which fails on Mesa as it exports the function, despite not implementing the relevant extension, and OSG doesn't check before setting the function pointer [here](https://github.com/openscenegraph/OpenSceneGraph/tree/40bddd8d9532620dc704adce0d2fb0c854d62d0f/src/osg/GLExtensions.cpp#L1097). It seems that on Mesa, the call is a no-op, as I get incomplete FBO errors when OSG uses that path, but not if I replace the call with the non-coverage variant, and I can't find an actual implementation in the Mesa source, but can for the regular variant.

I expect that the most sensible solution would be to check for the presence of the actual extension, either before setting the function pointer or in `isRenderbufferMultisampleCoverageSupported`.

glebm:
`isRenderbufferMultisampleCoverageSupported()` should not return true if the extension is not supported.

Fixes #1028

### DIFF
--- a/examples/osgfpdepth/osgfpdepth.cpp
+++ b/examples/osgfpdepth/osgfpdepth.cpp
@@ -138,9 +138,7 @@ void getPossibleConfigs(GraphicsContext* gc, BufferConfigList& colorConfigs,
         return;
     if (ext->isMultisampleSupported)
         glGetIntegerv(GL_MAX_SAMPLES_EXT, &maxSamples);
-    // isMultisampleCoverageSupported
-    if (isGLExtensionSupported(contextID,
-                               "GL_NV_framebuffer_multisample_coverage"))
+    if (ext->isRenderbufferMultisampleCoverageSupported())
     {
         glGetIntegerv(GL_MAX_MULTISAMPLE_COVERAGE_MODES_NV,
                       &coverageSampleConfigs);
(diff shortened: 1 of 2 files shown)


robertosfield:
Thanks merged with master and 3.6 branch.

PR #1058, opened 2021-03-22
Fix: Removed drawing from compileGLObjects method of text.

deimon:
At the moment, in compileGLObjects method of text, not only the preparation of glBuffers takes place, but also the rendering itself. This leads to text rendering in those moments when it should not be.


robertosfield:
I've just done a first pass review.  That's a lot of additional code, all that additional code will need to debugged and maintained.  So while I can see it might fix a particular usage case, I think we should look for simpler ways of addressing the issue.

Could you provide a code example where the problem you are trying to address appears?

PR #1059, opened 2021-03-24
FreeType plugin: 'otf' extension added

valid-ptr:


robertosfield:

PR #1062, opened 2021-04-13
Fix bounds for empty text nodes

calumr:
_lineCount can be 0 if the TextBase has no text, so we get underflow if the alignment is one of the *_BOTTOM_BASE_LINE. Empty text nodes seem to come up in some of the .dxf files our users are importing.

I've worked around the bug in our code for now by deleting the empty text nodes on load.

### DIFF
--- a/src/osgText/TextBase.cpp
+++ b/src/osgText/TextBase.cpp
@@ -484,6 +484,14 @@ void TextBase::computePositions()
 
 void TextBase::computePositionsImplementation()
 {
+    _normal = osg::Vec3(0.0f,0.0f,1.0f);
+
+    if (_text.empty())
+    {
+        _offset = Vec3();
+        return;
+    }
+
     switch(_alignment)
     {
         case LEFT_TOP:      _offset.set(_textBB.xMin(),_textBB.yMax(),_textBB.zMin()); break;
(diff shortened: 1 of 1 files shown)


robertosfield:
Could provide details of how to reproduce the problem?

I think it would be better to have a dedicated code at the start of the method, along the lines of:

     if (empty) { set fallback values; return; }

calumr:
```
    osg::ref_ptr<osg::Node> root = osgDB::readRefNodeFile("c:\\dev\\emptyTextNodes.dxf");

    osg::ComputeBoundsVisitor cbv;
    root->accept(cbv);
    if (cbv.getBoundingBox().xMax() > 10000)
    {
        osg::notify(osg::NOTICE) << "too big: "<< cbv.getBoundingBox().xMax() << std::endl;
    }
```

Run that on the attached .dxf, it should print "too big: 7.50612e+09".

[emptyTextNodes.zip](https://github.com/openscenegraph/OpenSceneGraph/files/6303955/emptyTextNodes.zip)

calumr:
I've updated the change to check `if (_text.empty())` as you suggested.

openscenegraph:
Hi Calum,

Thanks for the model.  I can't look at it right away, but once I've
finished my current task I'll have a look.

W.r.t to the changed commit.  I don't know whether the Vec3{} usage will
work with older compilers, () would be safer.  Also using an if () else ()
block results in less readable code, for this type of usage I typically
doing a return in the if (null_condition) { hande_nul_condition; return; }
as it minimizes the changes required to unrelated code.

Robert.

calumr:
All done. I had to move the `_normal` calculation up above the `if`, I hope this is OK. There's no rush on this, thanks for taking a look.

robertosfield:
Looks good!

robertosfield:
Also merged with 3.6 and MeshShaders branches.

PR #1065, opened 2021-04-28
Add move constructor and move assignment operator to ref_ptr

elsid:
Use conditional compilation to make it work only with C++11 support.

This gives a performance boost for inserting and erasing from a middle of a `std::vector<osg::ref_ptr>` because no reference counting is done while elements are shifted. Insert example is `Group::insertChild` function [here](https://github.com/openscenegraph/OpenSceneGraph/blob/d64ad2e654001a4323a7fb39fa3d937ae71f1646/src/osg/Group.cpp#L98) and erase example is `Group::removeChildren` function [here](https://github.com/openscenegraph/OpenSceneGraph/blob/d64ad2e654001a4323a7fb39fa3d937ae71f1646/src/osg/Group.cpp#L199).

In the specific OpenMW location loading case total cost of `Group::insertChild` is reduced by 11 times and `Group::removeChildren` by 13 times. Overall this reduces specific location loading by half. For more details see profiling [before](https://gitlab.com/OpenMW/openmw/-/issues/5992#note_563104417) the change and [after](https://gitlab.com/OpenMW/openmw/-/issues/5992#note_563130087) the change. It might be questionable if we are actually doing the right thing in OpenMW by inserting and erasing so many elements. Still this change should make performance improvement even for a single insert or erase.

### DIFF
--- a/include/osg/ref_ptr
+++ b/include/osg/ref_ptr
@@ -22,6 +22,10 @@
 #include <string>
 #endif
 
+#if __cplusplus >= 201103L
+#include <utility>
+#endif
+
 namespace osg {
 
 template<typename T> class observer_ptr;
@@ -36,6 +40,9 @@ class ref_ptr
         ref_ptr() : _ptr(0) {}
         ref_ptr(T* ptr) : _ptr(ptr) { if (_ptr) _ptr->ref(); }
         ref_ptr(const ref_ptr& rp) : _ptr(rp._ptr) { if (_ptr) _ptr->ref(); }
+#if __cplusplus >= 201103L
+        ref_ptr(ref_ptr&& rp) noexcept : _ptr(rp._ptr) { rp._ptr = 0; }
+#endif
         template<class Other> ref_ptr(const ref_ptr<Other>& rp) : _ptr(rp._ptr) { if (_ptr) _ptr->ref(); }
         ref_ptr(observer_ptr<T>& optr) : _ptr(0) { optr.lock(*this); }
         ~ref_ptr() { if (_ptr) _ptr->unref();  _ptr = 0; }
@@ -51,6 +58,17 @@ class ref_ptr
             assign(rp);
             return *this;
         }
+
+#if __cplusplus >= 201103L
+        template<class Other> ref_ptr& operator = (ref_ptr<Other>&& rp)
+        {
+            if (_ptr == rp._ptr) return *this;
+            if (_ptr != nullptr) _ptr->unref();
+            _ptr = rp._ptr;
+            rp._ptr = nullptr;
+            return *this;
+        }
+#endif
 
         inline ref_ptr& operator = (T* ptr)
         {


include/osg/ref_ptr
```diff
@@ -22,6 +22,10 @@
 #include <string>
 #endif
 
+#if __cplusplus >= 201103L
+#include <utility>
```
glebm, line 26: This header shouldn't be needed, `std::move` in C++11 isn't in `utility`

include/osg/ref_ptr
```diff
@@ -141,6 +156,20 @@ class ref_ptr
             if (tmp_ptr) tmp_ptr->unref();
         }
 
+#if __cplusplus >= 201103L
+        template<class Other> void moveAssign(ref_ptr<Other>&& rp)
```
glebm, line 160: This is only used by the move assignment operator, perhaps there is no need for this separate function?

include/osg/ref_ptr
```diff
@@ -52,6 +59,14 @@ class ref_ptr
             return *this;
         }
 
+#if __cplusplus >= 201103L
+        template<class Other> ref_ptr& operator = (ref_ptr<Other>&& rp)
```
glebm, line 63: Move assignment should also be marked `noexcept`

include/osg/ref_ptr
elsid: If it's not [utility](https://en.cppreference.com/w/cpp/utility/move) then which one? It's a part of the standard library so some header should contain it.

include/osg/ref_ptr
elsid: Maybe, but I wasn't sure what's the reasoning behind copy-assignment with the similar approach. Even there are 3 different definitions for it and two of them have common implementation, there is no actual difference.

include/osg/ref_ptr
elsid: Why?

include/osg/ref_ptr
AnyOldName3: All the sources I see say it's in <utility>. If it's in a different header on your machine, that's probably not something safe to rely on elsewhere.

glebm: There is no header you need to include for this `std::move`, it is always available.
The one in utility is a different function from moving all elements of one container to another.

include/osg/ref_ptr
glebm: https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#c66-make-move-operations-noexcept

This CppCon 2019 video is good for learning more 
https://www.youtube.com/watch?v=St0MNEU5b0o

include/osg/ref_ptr
glebm: Nevermind, I confused this with `std::move` from `<algorithm>`

include/osg/ref_ptr
glebm: This is always a good practice. Specifically for vectors this is a good practice because vector move-assignment is only noexcept if the elements are nothrow-move-assignable:

https://github.com/llvm-mirror/libcxx/blob/78d6a7767ed57b50122a161b91f59f19c9bd0d19/include/vector#L2929-L2932

include/osg/ref_ptr
elsid: Do you have an example of such compiler? `std::move` is a function not operator. And compiler [can't](https://godbolt.org/z/67x3fsPzs) get it out of nowhere including `std` namespace. For example libstdc++ that provides standard library for gcc defines it [explicitly](https://github.com/gcc-mirror/gcc/blob/406f58e1e38e92e4b881f3666b596843da308783/libstdc%2B%2B-v3/include/bits/move.h#L101-L105). libc++ does it different but when running only the compiler preprocessor it's there:
```c++
template <>
struct is_destructible<void>
    : public std::__1::false_type {};
# 2564 "/home/elsid/dev/llvm-project/build/clang/release/install/include/c++/v1/type_traits" 3
template <class _Tp>
inline __attribute__ ((__visibility__("hidden"))) __attribute__ ((__exclude_from_explicit_instantiation__)) constexpr
typename remove_reference<_Tp>::type&&
move(_Tp&& __t) noexcept
{
    typedef __attribute__((nodebug)) typename remove_reference<_Tp>::type _Up;
    return static_cast<_Up&&>(__t);
}
```

Without including utility gcc and clang give error:
```
error: use of undeclared identifier 'std'
    A b = std::move(b);
          ^
```

There is another [move](https://en.cppreference.com/w/cpp/algorithm/move) that is similar to `std::copy` except it moves elements instead of copying but it's defined in the algrorithm header.

include/osg/ref_ptr
elsid: The problem here is that `unref` potentially can throw exceptions. It can add element to a vector for example. I've changed the code to guarantee exception safety for ref_ptr state. If I add noexcept process will terminate when exception is thrown. Whether to allow such behaviour is up to this library maintainer.

include/osg/ref_ptr
elsid: Done

include/osg/ref_ptr
glebm: I see, up to @robertosfield then

robertosfield: @elsid is correct, an assignment to a ref_ptr<> that already has reference to an object could unref that object, and that could either directly throw an exception, or for the destruction of the object could throw an exception.

To be able to make the assignment operator noexcept you'd also have to make unref() a noexcept, and then the destructors that it might call.

However, if the ref_ptr<> is null prior to the move assignment then it'd be safe to make it a noexcept.  You can do this for the move copy constructors, but not the assignment.

include/osg/ref_ptr
robertosfield: I've merged this submission and am testing, but I can't see any use of std::move in the ref_ptr header, so why does it need to include a header for a function that it doesn't use?

Seems like there is no need for the include at all.  If code elsewhere relies upon std::move() then it should be the one to include<utility> not include/osg/ref_ptr.

robertosfield:
@elsid  While I've merged this PR, I now think it was premature.  The general implementation looks OK, and I've tested it and it does produce a nice performance improvement in a insert many children to the front of a group test (the worst case scenario).  

The problem is that the #include <utility> does look to have an direct purpose.  The ref_ptr<> code changes don't use std::move, so why you need to have an include for it?  My best guess is that some code you have uses std::move() but you aren't including utility in it, instead relying upon include/osg/ref_ptr including it for you.  This just isn't appropriate so I'm poised to remove the include <utility>, if you can't give a solid reason for it's inclusion I'll remove it this afternoon.

AnyOldName3:
> 
> 
> @elsid While I've merged this PR, I now think it was premature. The general implementation looks OK, and I've tested it and it does produce a nice performance improvement in a insert many children to the front of a group test (the worst case scenario).
> 
> The problem is that the #include does look to have an direct purpose. The ref_ptr<> code changes don't use std::move, so why you need to have an include for it? My best guess is that some code you have uses std::move() but you aren't including utility in it, instead relying upon include/osg/ref_ptr including it for you. This just isn't appropriate so I'm poised to remove the include , if you can't give a solid reason for it's inclusion I'll remove it this afternoon.

It was used earlier, but isn't any more since the force-push a couple of days ago. It's safe to remove.

elsid:
There was code:
```c++
template<class Other> ref_ptr& operator = (ref_ptr<Other>&& rp)
{
    moveAssign(std::move(rp));
    return *this;
}
```
I've removed moveAssign function but forgot to remove the include. It can be removed now. You may see the full version of that code [here](https://github.com/OpenMW/osg/commit/76e061739610bc9a3420a59e7c9395e742ce2f97).

robertosfield:
Thanks for the clarification.  I've remove the include and checked it into master.  I have also merged these changes with 3.6, MultiView and MeshShaders  branches.

robertosfield:
I have also been inspired to add move constructor and assignment operation to the vsg::ref_ptr<> and created a test program to test the insert child to position 0 as a worst case scenario for the none move variants:

    file:///home/robert/Contributors/CopyTest/main.cpp

In testing the OSG performance improves by around 15x for this test, the VSG improves by 20x.  

However, this worst case scenario, users really shouldn't be doing a repeated insert of thousands of children into a vector/group, so it'll likely over be noticeable on algorithms/scene graph that may well be best done in other ways.

AnyOldName3:
Not strictly relevant to this PR, but I thought I'd point out that the markdown for that snippet's wrong, and will make it a pain for anyone copying and pasting to try the benchmark themselves. As explained [here](https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#syntax-highlighting), the correct syntax is three backticks (`` ``` ``), then a Linguist language identifier, in this case `cpp`, then a line break, then the code, then a line break, then three more backticks. If the code's got more than three backticks in a row, add extra backticks to the opening and closing backticks. In this case, you'd want something like:
````md
```cpp
int main(int argc, char** argv)
{
    return 0;
}
```
````
which produces:
```cpp
int main(int argc, char** argv)
{
    return 0;
}
```

robertosfield:
Thanks, learn something new every day.  I've updated the entry above.

PR #1066, opened 2021-05-05
Fix bounds until the text is actually rendered.

deimon:
The bound can be used in LOD, if the text is placed in LOD. In this case, the wrong boundary may cause the text placed in LOD never is rendered.

### DIFF
--- a/src/osgText/TextBase.cpp
+++ b/src/osgText/TextBase.cpp
@@ -424,16 +424,18 @@ osg::BoundingBox TextBase::computeBoundingBox() const
 
     if (_textBBWithMargin.valid())
     {
-        bbox.expandBy(_textBBWithMargin.corner(0)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(1)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(2)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(3)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(4)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(5)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(6)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(7)*_matrix);
+        if (!_matrix.isIdentity())
+        {
+            bbox.expandBy(_textBBWithMargin.corner(0)*_matrix);
+            bbox.expandBy(_textBBWithMargin.corner(1)*_matrix);
+            bbox.expandBy(_textBBWithMargin.corner(2)*_matrix);
+            bbox.expandBy(_textBBWithMargin.corner(3)*_matrix);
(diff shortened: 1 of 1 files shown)


src/osgText/TextBase.cpp
```diff
@@ -424,16 +424,18 @@ osg::BoundingBox TextBase::computeBoundingBox() const
...
     {
-        bbox.expandBy(_textBBWithMargin.corner(0)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(1)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(2)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(3)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(4)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(5)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(6)*_matrix);
-        bbox.expandBy(_textBBWithMargin.corner(7)*_matrix);
+        if (!_matrix.isIdentity())
```
robertosfield, line 427: _matrix being identity can be a valid setting for it so this PR will introduce a bug.

robertosfield:
This PR looks dubious to me - there seems to be a lot of changes and I get the feeling there may be unwanted side effects.  

Do any of the OSG examples reproduce the problem you are seeing, or can you provide an example that reproduces the problem so the appropriate fix can be looked at.

deimon:
Really, my description was incomplete.
An example demonstrating the problem:
```
#include <osgText/Text>
#include <osgViewer/Viewer>
#include <osgViewer/ViewerEventHandlers>

void main(void)
{
  osgViewer::Viewer viewer;
  viewer.setUpViewOnSingleScreen();
  viewer.addEventHandler(new osgViewer::StatsHandler);

  osg::Group* root = new osg::Group;

  osgText::Text* text0 = new osgText::Text;
  text0->setAxisAlignment(osgText::Text::SCREEN);
  text0->setCharacterSizeMode(osgText::Text::SCREEN_COORDS);
  text0->setText("0");
  root->addChild(text0);

  osgText::Text* text1 = new osgText::Text;
  text1->setAxisAlignment(osgText::Text::SCREEN);
  text1->setCharacterSizeMode(osgText::Text::SCREEN_COORDS);
  text1->setText("1");

  osg::LOD* lod = new osg::LOD;
  root->addChild(lod);
  lod->addChild(text1, 0, 10.0f);

  viewer.setSceneData(root);
  viewer.realize();
  viewer.run();
}

```
Here are the important points: text1 has a constant screen size and is placed in a lod.
If the maximum range set about 10, then when the camera is closer to the text (where we expect it), it will not appear.
It will work when we set the maximum value of the range to about 16.

Why exactly 16: In our case, the size of the bounding box for text "1" is about 21 by 32. Taking into account all alignments, the center is obtained approximately at the point (11, 11). Therefore, the radius is approximately 16.
Therefore, when we move the camera closer to the point (0, 0) (where we expect the text to appear, given the constant screen size of the text), the distance to the center remains about 16.

PR #1069, opened 2021-06-10
Fix OpenThreads install on MSVC without a prefix

ananace:
All this change does is make the install config for OpenThreads identical to the one in ModuleInstall.cmake


robertosfield:
Also merged with 3.6, MeshShaders and MltiView branches

PR #1073, opened 2021-07-16
Fix for Optimizer handling of Switch nodes

cj-mcdonald:
If Geometry nodes are directly attached to a Switch node (without using Geodes) then the Optimizer MergeGeometryVisitor incorrectly merges them all together.  Simple fix to not merge directly attached Geometry nodes, but still traverse other child types.


cj-mcdonald:
Sorry, don't know what I'm doing with github, I'll try again

PR #1094, opened 2021-10-07
removes invalid glReadBuffer calls

bosvensson1:
While investigating reports of Open GL errors in Open MW, we found osg makes such invalid calls during the setup of framebuffer objects that do not use color attachments. The Open GL specification indicates GL_NONE is not an acceptable value for glReadBuffer.

### DIFF
--- a/src/osgUtil/RenderStage.cpp
+++ b/src/osgUtil/RenderStage.cpp
@@ -519,9 +519,7 @@ void RenderStage::runCameraSetUp(osg::RenderInfo& renderInfo)
             {
             #if !defined(OSG_GLES1_AVAILABLE) && !defined(OSG_GLES2_AVAILABLE) && !defined(OSG_GLES3_AVAILABLE)
                 setDrawBuffer( GL_NONE, true );
-                setReadBuffer( GL_NONE, true );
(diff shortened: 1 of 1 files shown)


openscenegraph:
> While investigating reports of Open GL errors in Open MW, we found osg
> makes such invalid calls during the setup of framebuffer objects that do
> not use color attachments. The Open GL specification indicates GL_NONE is
> not an acceptable value for glReadBuffer.
>
Hi Bo, just had a quick look at your PR and docs on glReadBuffer
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glReadBuffer.xhtml
and concur that it looks like GL_NONE is not avoid, but... the code
alongside the part you changed explicitly suggests that this is needed.
This particular bit of code will have been froma submissions, and could
well be over a decade old.  Just removing code to quieten the warning you
are seeing might have some unforeseen consequences, but perhaps the
original code was wrong to include the reset of the ReadBuffers mode to
GL_NONE.

// If no color attachment make sure to set glDrawBuffer/glReadBuffer to none
// otherwise glCheckFramebufferStatus will fail
// It has to be done after call to glBindFramebuffer (fbo->apply)
// and before call to glCheckFramebufferStatus

PR #1116, opened 2022-01-27
OBJ reader preserves relative paths for image filenames

matthew-reid:
This change ensures that relative image paths are preserved when the OBJ is written out. This makes the saved OBJ portable. Without this change, images will only load correctly if they are stored in model.getDatabasePath().


src/osgPlugins/obj/ReaderWriterOBJ.cpp
```diff
@@ -243,6 +243,9 @@ static void load_material_texture(    obj::Model &model,
         {
             // first try with database path of parent.
             image = osgDB::readRefImageFile(model.getDatabasePath()+'/'+filename, options);
+
+			// Preserve relative path in case filename is written out later
+			image->setFileName(filename);
```
CarSim-bporter, line 248: You may want to change the tab characters in your indents to spaces to be consistent with the existing code.

matthew-reid: Fixed

PR #1126, opened 2022-04-06
Make sure GUIActionAdapter has a single vtable

pvalsecc:
With clang 10, a simple `dynamic_cast<osgViewer::Viewer*>(&aa)` with aa of
type `osgGA::GUIActionAdapter` but actually referencing a `osgViewer::Viewer`
was returning `nullptr`.

This is due to the fact that the compiler didn't know where to put the vtable
implementation. By the way, clang has a nice warning for that:
Class has no out-of-line virtual method definitions; its vtable will be
emitted in every translation unit [-Wweak-vtables]

Since the Viewer was instanciated in the code of the lib, I guess it used a
different vtable than the dynamic_cast in my binary, making dynamic_cast think
it was a different class and thus returning a nullptr.

I've made the GUIActionAdapter class visible, since it's used as base class
of other visible classes.

### DIFF
--- a/src/osgGA/GUIActionAdapter.cpp
+++ b/src/osgGA/GUIActionAdapter.cpp
@@ -1,0 +1,18 @@
+/* -*-c++-*- OpenSceneGraph - Copyright (C) 1998-2006 Robert Osfield
+ *
+ * This library is open source and may be redistributed and/or modified under
+ * the terms of the OpenSceneGraph Public License (OSGPL) version 0.0 or
+ * (at your option) any later version.  The full license is in LICENSE file
+ * included with this distribution, and on the openscenegraph.org website.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


src/osgGA/GUIActionAdapter.cpp
```diff
@@ -0,0 +1,18 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * OpenSceneGraph Public License for more details.
+*/
+
+#include <osgGA/GUIActionAdapter>
+
+using namespace osgGA;
+
+GUIActionAdapter::~GUIActionAdapter() = default;
```
robertosfield, line 18: You'll need to manually putting an a { } as older compilers won't handle = default.

PR #1136, opened 2022-05-01
Fix in Text3D: two line breaks were perceived as one

valid-ptr:


valid-ptr:
It is just the same implementation as in 2D Text.

PR #1139, opened 2022-05-16
Remove installing pdb files with ninja in realase mode

ahoarau:
Release build do not generate pdb files and cause the installation to fail.

Remove those 2 lines makes a successful install.


ahoarau:
@robertosfield Can we merge this ? Thank !

PR #1157, opened 2022-07-19
Fix RigTransformSoftware copy

calumr:
* Force init for new copy as _uniqVertexGroupList is not copied

The problem can occur if you deep copy a scene with a RigTransformSoftware in it after it has already been displayed. We were copying a pedestrian model and trying to use a different animation on it, but the second animation fails to apply as the copied RigTransformSoftware never gets inited.


mp3butcher:
I don't remember so well the code, so can't tell if i what i say is very safe but:
have you tried to copy _uniqVertexGroupList instead of forcing initdata to true?
It may avoid the expensive call to preparedata...

PR #1172, opened 2022-11-18
Add function osgDB::executableFilePath() to find the executable path platform independent

rhabacker:
The implementation has been taken from the VulkanSceneGraph project.


robertosfield:
I think it would be worth adding a comment before the function about it being taken from the VSG project.

PR #1188, opened 2022-12-30
Fix: Changes to make the non-const get method thread-safe.

deimon:
We had a crash in this line of code:
`osg::ref_ptr<osg::Referenced>& ptr = _managerMap[id];`
There was an attempt to add an element in _managerMap with an index that already existed.
As far as we understand OSG, we have not done anything illegal. In our case, during the execution of the destructor Program::PerContextProgram::~PerContextProgram(). We believe that calling this destructor simultaneously from different threads is a valid case.
In addition, in the comment to the const get-method, it is written that it can be called outwith a the graphics context thread. This means that it is quite legal to call this method from another thread and at the same time a non-const get method from the graphics context thread. Which can lead to crash.
We propose the following changes to make both get methods thread-safe.
We use COW(Copy-on-write) to minimize read time. It also makes sense to encapsulate _managerMap and COW.

### DIFF
--- a/include/osg/ContextData
+++ b/include/osg/ContextData
@@ -33,15 +33,25 @@ class OSG_EXPORT ContextData : public GraphicsObjectManager
         osg::GraphicsContext* getCompileContext() { return _compileContext.get(); }
 
         /** Get a specific GL extensions object or GraphicsObjectManager, initialize if not already present.
-          * Note, must only be called from a the graphics context thread associated with this osg::State. */
+          * Note, safe to call outwith a the graphics context thread associated with this osg::State. */
         template<typename T>
         T* get()
         {
             const std::type_info* id(&typeid(T));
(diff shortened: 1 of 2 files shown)


robertosfield:
While you've found a way to provoke a crash calling a destructor from multiple threads is not legal ever, with the OSG or elsewhere.

The problem is elsewhere in your system, trying to "fix" parts of the OSG to workaround this is not appropriate so I'm closing this PR.  As a general note, the changes make the code more complicated and far more difficult to follow what on earth is going on and for what reason. When this happens with any code you have to take a step back and ask whether this is simpler and more efficient solution,

deimon:
> calling a destructor from multiple threads is not legal ever, with the OSG or elsewhere

I meant parallel execution of destructors of different objects, but of the same type.

PR #1302, opened 2024-02-15
AC: Triangulate quads and polygons

amalon:
I'm a bit unclear whether there's any interest in accepting fixes, especially for use of newer GL features, and what branch it makes sense to apply them to, so I've just left this one on OpenSceneGraph-3.6.5 for now. It stems from me playing around with instanced geometry shaders for single-pass multiviewport rendering in osgXR / FlightGear (where work is ongoing to switch to GL core profile), since mesa on my AMD GPU doesn't support the OVR_multiview2 extension.

The AC plugin creates geometry containing quads and polygons, neither of which can be handled with geometry shaders, resulting in GL_INVALID_OPERATION errors from glDrawElements when geometry shaders are in use.

This is fixed by generating a single DrawElements of triangles from both _triangles and _quads (triangulating the quads as if each is a triangle fan), and treating convex polygons (in _polygons) as triangle fans.

Concave polygons (in _toTessellatePolygons) are already converted into triangle modes by a Tessellator object.

### DIFF
--- a/src/osgPlugins/ac/ac3d.cpp
+++ b/src/osgPlugins/ac/ac3d.cpp
@@ -1050,8 +1050,8 @@ class SurfaceBin : public PrimitiveBin {
             Tessellator.retessellatePolygons(*geometry);
         }
 
-        // handle triangles
-        if (!_triangles.empty())
+        // handle triangles & quads
+        if (!_triangles.empty() || !_quads.empty())
         {
             osg::ref_ptr<osg::DrawElementsUInt> drawElements = new osg::DrawElementsUInt(osg::PrimitiveSet::TRIANGLES);
             for (unsigned i = 0; i < _triangles.size(); ++i)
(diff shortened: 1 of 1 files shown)


robertosfield:
I'm so busy with the VSG project and associated client work right now I don't have time to review and merge submissions and make releases for the OSG.  Once I have some quiet time I will look at making a new OSG-3.6.6 release from the OSG-3.6 branch, and an OSG-3.8 release after that.

Projects looking at porting to OpenGL Core Profile would probably be better off just moving the VulkanSceneGraph, it support for shader based scene graphs is far better than the OSG.  OpenGL is a dead end, not touched by Khronos since 2017.  Vulkan is the future, it's much faster and more capable than OpenGL.  The VSG is much faster than the OSG even before adding in the Vulkan advantage.

amalon:
Okay, thanks. No problem.

Yeh I get that Vulkan is the future. The core profile is I believe needed for the FG pretty HDR pipeline to work on certain GPUs/drivers, and is a much smaller transition than to vulkan/VSG.

robertosfield:
OpenGL Core Profile was a mistake that Khronos made, screwing over legacy applications and forcing them to break old code and rewrite to something that wasn't really fit for purpose.  It was meant to get OpenGL closer to what Vulkan is now but fell far short of actually support modern hardware in the way that it actually works.

Far better would have been to maintain OpenGL for legacy applications without breaking them and just get folks to adopt a modern Vulkan if they wanted modern hardware support.

Porting to Vulkan/VulkanSceneGraph is a bigger task but the end result is far better and has a future.  OpenGL Core Profile is just as much as dead end as the rest of OpenGL.

PR #1360, opened 2025-02-23
Fix linking with libtiff when libtiff is compiled with CMake

DaGenix:
The problem is that we are trying to link against TIFF_LIBRARY. However, according to the documentation for FindTIFF.cmake, we should be using TIFF_LIBRARIES.
(https://github.com/Kitware/CMake/blob/master/Modules/FindTIFF.cmake).

Why is it this way? It was set this way in commit
0328691f76e70f2e7b8cfe9d46b97808e086d28d where the message was:

> From Sherman Wilcox, "there's a bug in the cmake file for the tiff plugin. See
> attached. The problem was that the output files were not properly
> setting the debug/release libs due to this cmake bug. What occurred was
> the release lib was set in all configurations."

Sadly, that commit was form 2010 and I cannot find the attachment since it appears to predate the use of GitHub. I can hope, however, that whatever the CMake issue was, it was fixed in the 15 years since that commit.

The reason to make this change is because while using TIFF_LIBRARY works if libtiff is compiled with autotools, it fails if libtiff is compiled with CMake.

The reason is subtle. When, libtiff is compiled with autotools, it will be located by CMake using FindTIFF.cmake
(https://github.com/Kitware/CMake/blob/master/Modules/FindTIFF.cmake). That module will first attempt to find libtiff using CMake configuration files, but, when that fails, falls back to calling find_library() to locate libtiff. Aftwards, it calls select_library_configurations() which set the TIFF_LIBRARY variable. And so, things work fine.

However, if libtiff is compiled with CMake, FindTIFF.cmake will be successful in finding the CMake config files and return without every calling select_library_configurations(). As such, TIFF_LIBRARY is not set.

When building using GCC and the standard linker on Linux, this doesn't produce an error - but it does result in a shared object with unresolved symbols. When using Clang on Darwin, however, the result is that linking fails due to the unresolved symbols. The net result is that neither works - but it appears on Linux to have succeeded while on Darwin there is a hard failure that breaks the entire build.


carlo-bramini:
CYGWIN build is also affected by this bug and simply replacing `TIFF_LIBRARY` with `TIFF_LIBRARIES` fixed the issue on my side too.

